PR #5, opened 2016-06-14
fix build issues

foochris:
[#119285059]


jamescw:
@foochris this is really bad practice that we have created two build files for the frontend, this is not really acceptable and will cause even more confusion, I am only merging this on the basis of getting prod fixed ASAP. I trust you and @robertavram will reach consistency on the way the frontend is built and released.

PR #18, opened 2016-06-22
fix issue with IE11+ caused by XMLHttp response type bug (JSON object vs. string)

foochris:
[fixed #121662753]


robertavram:
This has been merged.

PR #20, opened 2016-06-24
Fix small issues

achamseddine:
@jamescw 

Result display name
Approve trip by supervisor
Do not send trip notification to PA if no budget owner

### DIFF
--- a/EquiTrack/trips/forms.py
+++ b/EquiTrack/trips/forms.py
@@ -188,8 +188,9 @@ class RequireOneLocationFormSet(BaseInlineFormSet):
         new_locations = sorted([f.instance for f in self.forms if f.cleaned_data])
         if self.instance.approved_by_supervisor:
             old_locations = [i for i in self.instance.triplocation_set.all()]
-            if new_locations != old_locations:
-                raise ValidationError('You cannot modify the location after the trip has been approved')
+            for location in new_locations:
+                if location not in old_locations:
+                    raise ValidationError('You cannot modify the location after the trip has been approved')
 
         form_count = len(new_locations)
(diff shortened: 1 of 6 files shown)


EquiTrack/trips/forms.py
```diff
@@ -188,8 +188,9 @@ def clean(self):
         new_locations = sorted([f.instance for f in self.forms if f.cleaned_data])
         if self.instance.approved_by_supervisor:
             old_locations = [i for i in self.instance.triplocation_set.all()]
-            if new_locations != old_locations:
-                raise ValidationError('You cannot modify the location after the trip has been approved')
+            for location in new_locations:
+                if location not in old_locations:
```
robertavram: this will allow you to remove locations after the trip was approved.

EquiTrack/trips/models.py
```diff
@@ -402,7 +402,7 @@ def send_trip_request(cls, sender, instance, created, **kwargs):
                     instance.travel_assistant.email
                 )
 
-            if instance.ta_required and instance.programme_assistant and not instance.ta_drafted:
+            if instance.ta_required and instance.programme_assistant and instance.budget_owner and not instance.ta_drafted:
                 emails.TripTAEmail(instance).send(
```
robertavram: Why do we need to add budget_owner here?

EquiTrack/trips/forms.py
achamseddine: You alright, I will add a length validation

EquiTrack/trips/models.py
achamseddine: Supply section requested that no needs to send email notification to "Programme assistant" if there's no "Budget owner" selected

robertavram:
Closing and pulling in develop

PR #25, opened 2016-06-27
Http https link

achamseddine:
@jamescw http links are no longer accessible by users


robertavram:
Please pull this agains develop

PR #27, opened 2016-06-28
display spinner when waiting for map location search results

foochris:


foochris:
@robertavram The code has been simplified.

PR #28, opened 2016-06-29
TPM Visit API

achamseddine:
@robertavram


jamescw:
@robertavram why is this invalid?

jamescw:
Fixes issue #51

EquiTrack/partners/serializers.py
```diff
@@ -248,6 +232,23 @@ class Meta:
...
 
+class InterventionSerializer(serializers.ModelSerializer):
+
+    pca_id = serializers.CharField(source='id', read_only=True)
+    pca_title = serializers.CharField(source='title')
+    pca_number = serializers.CharField(source='reference_number')
+    partner_name = serializers.CharField(source='partner.name')
+    partner_id = serializers.CharField(source='partner.id')
+    pcasector_set = PCASectorSerializer(many=True, read_only=True)
+    locations = GWLocationSerializer(many=True, read_only=True)
```
robertavram: we're consuming this a lot without locations for the partner portal, trying to keeps things lite since we're having serious compounding db queries (in the thousands)... where is it consumed that needs locations?

EquiTrack/tpm/views.py
```diff
@@ -0,0 +1,58 @@
...
+
+from rest_framework import viewsets, mixins, permissions
+from rest_framework.response import Response
+from rest_framework import status
+
+from .models import TPMVisit
+from .serializers import TPMVisitSerializer
+
+
+class TPMVisitViewSet(mixins.RetrieveModelMixin,
```
robertavram: is this api only to migrate the data? or is it consumed by anything?

achamseddine: for the moment we need it to migrate the TPM data

EquiTrack/partners/serializers.py
achamseddine: I need load PCA locations to make the relation between TPM Visit and PCA during the migration

EquiTrack/tpm/views.py
jamescw: we need to both read and write TPM visits @robertavram, I think we are the only country using this feature currently and it is used alot by our staff to arrange monitoring visits, without the ability to provision the old visits and fix issues we new visits we are blocked

EquiTrack/partners/serializers.py
jamescw: This should only be the locations for a single intervention, if you are worried about performance then we can implement a lazy load on the nested collection via the depth mechanism but it will just mean we make two calls instead on one

robertavram: It makes sense to add locations to this api, just not fully serialized. (id only)

If it makes sense we can implement a good locations api for the intervention locations... where you could potentially get bulk locations serialized from a list of location Ids...
let me know if that would be helpful...

robertavram:
there have been a lot of model changes since this pr... if for Lebanon and other countries we need to migrate old data.. please give us the data and we'll do it... access to this api will be superuser only anyway. if anyone has any better thoughts let's have a bilateral talk

PR #29, opened 2016-06-29
Intervention Budget total cash

achamseddine:
Currently multi year budgets are displayed incorrectly as the only the last budget line is displayed and the total across years is therefore incorrect.


robertavram:
@achamseddine , please tag what features/bugs you're delivering in all these PRs,
 by using:  [(Finishes|Fixes|Delivers) #TRACKER_STORY_ID]

achamseddine:
@robertavram [Fixes #123412861]

robertavram:
Waiting for BO to confirm that amended budgets are not an issue if they
get aggregated here.

robertavram:
according to FRG, the best practices for adding an amended budget is to add the total value of the budget for the amendment.  IF we aggregate all budgets including all amendments we'll end up with an inaccurate budget.  The way to tackle this is to only take the last budget (amendment or original).

achamseddine:
In case we have interventions with start date on 2015 and end date on 2016, that mean we have to split budget into 2 line 2015 and 2016 (percentage for each year) so with this action we can have the accurate numbers in the HACT dashboard for each year otherwise we get 0. I agree that we have to ignore the amendment

robertavram:
For HACT we need the planned amount for for current year not for the whole intervention...
we can only set / year anyway...

closing this...

FIX for hact dashboard is coming today

jamescw:
@robertavram HACT is clear no doubt but there is straghtforward bug here that is not being fixed, an example:
image

There is a two budget lines in this PD, one for 2015 and one for 2016. HACT dashboard will already take the 2016 line as it relates to this year, but the total value of the PD is incorrect in the summary and export screens:

image

It should equal both years, in this case: 116574

jamescw:
[Fixes #123200751]

robertavram:
I understand what you mean...
The issue with this fix is that it aggregates amendments as well... which still ends up with an inaccurate amount. Please update so that the budget always reflects the last amendment.

There is a small subtlety here that, when an amendment is generated the whole budget gets affected and not only part of the budget. (For example if an original budget is  2015 and 2016, and in the system there is a budget amendment but only for the year 2016 the code should still aggregate 2015 original with 2016 amended).
PLease let me know if this makes sense... Otherwise we'll have to check with the Business owner to confirm that when there is an amendment to the budget and that budget spans multiple years, all years will be reflected in the amendment if if the budget for some of those years has not changed.

@achamseddine please include tests for this.

ntrncic:
I asked nazeef, hact dashboard should only reflect the current year.

cndesrosiers:
Hi all -- from what I gather from our engineering team, this falls outside of the business requirements.  I've put a note in my calendar for the beginning of September to revisit this bug ticket when everyone is back in the office.

jamescw:
@robertavram there is an easy way to deal with this. When making an amendment only the additional cash should be added for the affected year, so if i have a two year budget and amend in 2016 then i create a new line with the additional amount, tagged to the relevant amendment log and the relevant year. **Therefore when summing the rows across all budgets lines, the budget will always be correct as the rows are distinct. The HACT dashboard only selects rows for the current year so its calculations will also be correct.**

robertavram:
sounds good... will discuss with Nazeef again..
the only difference between what we're saying is: you are saying that budget amendments reflect only the difference in amounts and the guidance that I've gotten so far says that budget amendments need to reflect the total sum of the entire budget to avoid scenarios in which when a budget gets cut for the year, the addition is negative.

robertavram:
Fixed in #160 closing...

basically any amendment should represent the entire value for that year not just the delta... (BO guidance) therefore budget needs to be additive only for latest amendments and totaled across years

PR #34, opened 2016-07-01
set agreement start end dates based on signed dates and result structure

ntrncic:
PCAs start date needs to be linked with last sign date from partner or UNICEF and end date of PCA link to country programme structure end date

[Finishes:#105549138]


EquiTrack/partners/forms.py
```diff
@@ -341,6 +344,26 @@ def clean(self):
                     _(u'SSFA can not be more than a year')
                 )
 
+        #  set start date to one of the signed dates
+        if start is None and agreement_type == Agreement.PCA:
```
robertavram: if there is a start date then the partner signed it, we're not updating the date here.
Another case is: signed_by_partner_date sets the start date, then unicef signs it... the start date needs to update.

My preference here would be to do validation only..  not automatic assignment of the start date. I will check with BO.

Also if you could include tests for these.. it would be great!

EquiTrack/partners/forms.py
```diff
@@ -341,6 +344,26 @@ def clean(self):
...
+                if signed_by_partner_date > signed_by_unicef_date:
+                    self.cleaned_data[u'start'] = signed_by_partner_date
+                else:
+                    self.cleaned_data[u'start'] = signed_by_unicef_date
+
+            if signed_by_partner_date and not signed_by_unicef_date:
+                self.cleaned_data[u'start'] = signed_by_partner_date
+
+            if signed_by_unicef_date and not signed_by_partner_date:
+                self.cleaned_data[u'start'] = signed_by_unicef_date
```
robertavram: take out the previous two conditions and add tests.. :)

EquiTrack/partners/forms.py
```diff
@@ -341,6 +344,26 @@ def clean(self):
...
+
+        # set end date to result structure end date
+        if end is None:
+            result_structure = ResultStructure.objects.order_by('to_date').last()
```
robertavram: self.cleaned_data[u'end'] = ResultStructure.current()

PR #41, opened 2016-07-13
update Android / iOS builds for 117711369 & 109391948

foochris:


foochris:
@robertavram Instead of deploying a new version of eTools, it may be easier to simply copy the files to the static site.

robertavram:
sounds good! will merge in a bit

foochris:
@robertavram let me know when the files have been copied so I can test and if successful, mark the stories as Delivered.

PR #62, opened 2016-07-29
feat(commands) create add_users command

fafcrumb:
This command creates dummy users for testing.
The user specifies how many to create with the `num_users` argument.
Users are created with:
username: testuser_'#'
email: testuser_'#'@unicef.org
password: un1c3f
country: uat


EquiTrack/users/management/commands/add_users.py
```diff
@@ -0,0 +1,37 @@
+import sys
+from django.core.management.base import BaseCommand, CommandError
+from users.models import UserProfile, Country, Office, Section
+from django.contrib.auth.models import User
+
+
+class Command(BaseCommand):
+    help = 'Creates new test users with '             \
```
ntrncic: take this help stuff out please. it's a public repo we don't want people knowing this

PR #63, opened 2016-08-02
carto db import fix

ntrncic:
Fix for importing cartoDB tables and update for map layer images


robertavram:
@ntrncic I merged the develop branch upstream. colsing this since develop already had your update

PR #69, opened 2016-08-03
added ability to hide partner records

ntrncic:
Ability to hide Partner Organization records in the admin


cndesrosiers:
Can you reference where this request came from?

ntrncic:
From Kidus. It's on the white board

PR #79, opened 2016-08-08
Etrips download fix

robertavram:


robertavram:
not needed at this time.

PR #80, opened 2016-08-10
Feature results api [WIP]

ilcsik:
@robertavram can you please review? 

You were right regarding M2M fields, in case of existing related objects, there's no need for custom  create/update method, a simple` PrimaryKeyRelatedField() `with `many=True` will do the trick when updating using existing IDs.

I'm going to add some tests so it's WIP.

### DIFF
--- a/EquiTrack/EquiTrack/factories.py
+++ b/EquiTrack/EquiTrack/factories.py
@@ -205,7 +205,7 @@ class ResultFactory(factory.django.DjangoModelFactory):
     sections = [factory.SubFactory(SectionFactory)]
     labels = ["label1", "label2"]
     geotag = [factory.SubFactory(LocationFactory)]
-    status = "status"
+    status = "Target Met"
     prioritized = False
     metadata = {"status_description": "some status"}
 
--- a/EquiTrack/EquiTrack/factories.py
+++ b/EquiTrack/EquiTrack/factories.py
@@ -171,8 +171,8 @@ class MilestoneFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = report_models.Milestone
 
-    description = factory.Faker("text")
-    assumptions = factory.Faker("text")
+    description = factory.Sequence(lambda n: 'Description {}'.format(n))
+    assumptions = factory.Sequence(lambda n: 'Assumptions {}'.format(n))
 
 
 class ResultTypeFactory(factory.django.DjangoModelFactory):
@@ -200,7 +200,7 @@ class ResultFactory(factory.django.DjangoModelFactory):
     name = factory.Sequence(lambda n: 'Result {}'.format(n))
     from_date = date(date.today().year, 1, 1)
     to_date = date(date.today().year, 12, 31)
-    assumptions = factory.Faker("text")
+    assumptions = factory.Sequence(lambda n: 'Assumptions {}'.format(n))
     users = [user_models.UserProfile.objects.all().first().id]
     milestone = factory.SubFactory(MilestoneFactory)
     sections = [fuzzy.FuzzyText().fuzz() for _ in xrange(3)]
--- a/EquiTrack/EquiTrack/factories.py
+++ b/EquiTrack/EquiTrack/factories.py
@@ -203,7 +203,6 @@ class ResultFactory(factory.django.DjangoModelFactory):
     users = []
     milestone = factory.SubFactory(MilestoneFactory)
     sections = [factory.SubFactory(SectionFactory)]
-    labels = ["label1", "label2"]
     geotag = [factory.SubFactory(LocationFactory)]
     status = "Target Met"
     prioritized = False
--- a/EquiTrack/EquiTrack/factories.py
+++ b/EquiTrack/EquiTrack/factories.py
@@ -203,7 +203,7 @@ class ResultFactory(factory.django.DjangoModelFactory):
     assumptions = factory.Sequence(lambda n: 'Assumptions {}'.format(n))
     users = []
     milestone = factory.SubFactory(MilestoneFactory)
-    sections = [fuzzy.FuzzyText().fuzz() for _ in xrange(3)]
+    sections = [factory.SubFactory(SectionFactory)]
     labels = ["label1", "label2"]
     geotag = [factory.SubFactory(LocationFactory)]
     status = "status"
@@ -218,6 +218,15 @@ class ResultFactory(factory.django.DjangoModelFactory):
         if extracted:
             for geotag in extracted:
                 self.geotag.add(geotag)
+
+    @factory.post_generation
+    def sections(self, create, extracted, **kwargs):
+        # Handle M2M relationships
+        if not create:
+            return
+        if extracted:
+            for sections in extracted:
+                self.sections.add(sections)
 
 
 # class FundingCommitmentFactory(factory.django.DjangoModelFactory):
--- a/EquiTrack/EquiTrack/factories.py
+++ b/EquiTrack/EquiTrack/factories.py
@@ -5,8 +5,10 @@ __author__ = 'jcranwellward'
 
 from datetime import datetime, timedelta, date
 from django.db.models.signals import post_save
+from django.contrib.gis.geos import GEOSGeometry
 
 import factory
+import factory.fuzzy as fuzzy
 
 from users import models as user_models
 from trips import models as trip_models
@@ -102,6 +104,22 @@ class TripFactory(factory.django.DjangoModelFactory):
     to_date = from_date + timedelta(days=1)
 
 
+class GatewayTypeFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = location_models.GatewayType
+
+    name = factory.Sequence(lambda n: 'GatewayType {}'.format(n))
+
+
+class LocationFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = location_models.Location
+
+    name = factory.Sequence(lambda n: 'Location {}'.format(n))
+    gateway = factory.SubFactory(GatewayTypeFactory)
+    point = GEOSGeometry("POINT(20 20)")
+
+
 class LinkedLocationFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = location_models.LinkedLocation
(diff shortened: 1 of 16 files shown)


EquiTrack/reports/models.py
```diff
@@ -94,6 +115,24 @@ class Result(MPTTModel):
     sic_name = models.CharField(max_length=255, null=True, blank=True)
     activity_focus_code = models.CharField(max_length=8, null=True, blank=True)
     activity_focus_name = models.CharField(max_length=255, null=True, blank=True)
+    sections = pgfields.ArrayField(models.CharField(
```
robertavram: let's us id's here.. because if the name of the section gets updated querying breaks

EquiTrack/reports/serializers.py
```diff
@@ -72,9 +75,22 @@ class Meta:
         model = Indicator
 
 
+class MilestoneSerializer(serializers.ModelSerializer):
+
+    id = serializers.CharField(read_only=True)
```
robertavram: i thought ids are read-only by default... I don;t think we need this here.. can you confirm?

EquiTrack/reports/tests/test_views.py
```diff
@@ -35,6 +38,7 @@ def test_api_results_list(self):
         response = self.forced_auth_req('get', '/api/reports/results/', user=self.unicef_staff)
 
         self.assertEquals(response.status_code, status.HTTP_200_OK)
+        print(response)
```
robertavram: let's remove the print statement :)

EquiTrack/reports/views.py
```diff
@@ -52,6 +61,14 @@ class SectorViewSet(mixins.RetrieveModelMixin,
 #     serializer_class = GoalCreateSerializer
 
 
+class MilestoneViewSet(viewsets.ModelViewSet):
+    """
+    CRUD api for Milestones
+    """
+    queryset = Milestone.objects.all()
+    serializer_class = MilestoneSerializer
```
robertavram: let's add permission classes onl;y for staff usrers

robertavram:
there seems to be something up with the updated libraries... do you have visibility on why the CI is failing?

👍  looks great so far!
thanks!

ilcsik:
Yes it was the updated factory boy lib that failed due to some importerror. There's some moving/migrating going on regarding a dependency (Faker). 

https://github.com/joke2k/faker/issues/284
https://github.com/joke2k/faker/issues/331

I couldnt get this work in an hour for some reason, so decided to move back to the earlier version.

EquiTrack/reports/models.py
ilcsik: Should it be a choicefield with static list like labels and status?
Or a new model called Section, and an M2M field with CRUD API?

EquiTrack/reports/serializers.py
ilcsik: Actually this was copy-pasted from the existing models, they seem to have it consistently. 

I think the only purpose is to expose the ID on a GET, and I used to use Meta fields for this so I refactored the Milestone model. If there's no other reason we can refactor other models on the way as well.

EquiTrack/reports/views.py
ilcsik: I added IsAdminUser as a permission class - shall I add to all related like Results view as a mixin? Is there anything that would break in this case?

EquiTrack/reports/models.py
robertavram: Section model exists in users.models

EquiTrack/reports/views.py
robertavram: IsAdminUser  permission required should be enough on the viewset

EquiTrack/reports/models.py
ilcsik: Ok, done.

EquiTrack/reports/serializers.py
ilcsik: Done in https://github.com/unicef/etools/pull/80/commits/1eeccc70fd19912fc8faf099e677e3157afa4dda

EquiTrack/reports/tests/test_views.py
ilcsik: Fixed in https://github.com/unicef/etools/pull/80/commits/22cf5778e722dd6798c30a62b2a8d5640a1754c4

EquiTrack/reports/views.py
ilcsik: Added to ResultViewSet as well, done in https://github.com/unicef/etools/pull/80/commits/032e3d49f7142999db8c3ce66753cca2137054b0

EquiTrack/EquiTrack/factories.py
```diff
@@ -5,8 +5,10 @@
 
 from datetime import datetime, timedelta, date
 from django.db.models.signals import post_save
+from django.contrib.gis.geos import GEOSGeometry
 
 import factory
+import factory.fuzzy as fuzzy
```
robertavram: are you using fuzzy in this file?

EquiTrack/EquiTrack/factories.py
```diff
@@ -149,6 +167,21 @@ class Meta:
...
+
+    description = factory.Sequence(lambda n: 'Description {}'.format(n))
+    assumptions = factory.Sequence(lambda n: 'Assumptions {}'.format(n))
+
+
+class ResultTypeFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = report_models.ResultType
+
+    name = factory.Sequence(lambda n: 'ResultType {}'.format(n))
```
robertavram: resultType .. there are only 3 result_types possible at the moment and they get populated initially.
"Outcome", "Output", "Activity"
If you could put this in the DB pre-test it would be awesome :)

EquiTrack/reports/models.py
```diff
@@ -1,8 +1,13 @@
 __author__ = 'jcranwellward'
 
 from django.db import models
+import django.contrib.postgres.fields as pgfields
+from jsonfield import JSONField
+import django.contrib.gis.db.models as gismodels
```
robertavram: I believe you;re not using this anymore in this file.. (gismodels)

EquiTrack/reports/models.py
```diff
@@ -83,6 +94,16 @@ class Result(MPTTModel):
         related_name='children',
         db_index=True
     )
+    assumptions = models.TextField(null=True, blank=True)
+    # Holds UserProfile ids. For tagging this solution is sufficient,
+    # but it lacks integrity checking. Switch to M2M field if needed.
+    users = pgfields.ArrayField(models.IntegerField(), default=list)
+    milestone = models.OneToOneField(
```
robertavram: I believe we can have more than one milestone / Activity or Output. This could thoretically be a foreign key for Result from the Milestone Model

EquiTrack/reports/models.py
```diff
@@ -94,6 +115,23 @@ class Result(MPTTModel):
     sic_name = models.CharField(max_length=255, null=True, blank=True)
     activity_focus_code = models.CharField(max_length=8, null=True, blank=True)
     activity_focus_name = models.CharField(max_length=255, null=True, blank=True)
+    sections = models.ManyToManyField(Section)
+    LABELS = (
```
robertavram: We've talked internally and we'll look into better ways of labeling things for now.. (tagging Results) if you know of a library that handles this well .. we're open to sugestions.. let's remove labels for now... we'll get back to it soon

EquiTrack/reports/models.py
```diff
@@ -94,6 +115,23 @@ class Result(MPTTModel):
...
+        ("label1","label1"),
+        ("label2","label2"),
+        ("label3","label3"),
+    )
+    labels = pgfields.ArrayField(models.CharField(
+        max_length=255, null=True, blank=True, choices=LABELS), default=list)
+    STATUS = (
+        ("on track","on track"),
+        ("constraint","constraint"),
+        ("status3","status3"),
```
robertavram: statuses for now should be:
"On Track", "Constrained", "No Progress", "Target Met"

EquiTrack/requirements/base.txt
```diff
@@ -52,6 +52,7 @@ django-easy-pdf
 xhtml2pdf==0.0.6
 html5lib==0.9999999 #must be added after xhtml2pdf
 reportlab
+sqlparse==0.1.19  # for django debug toolbar to not break admin
```
robertavram: let's move this into a local.txt requirements file if it;s only used for the debug toolbar

EquiTrack/EquiTrack/factories.py
ilcsik: Removed in https://github.com/unicef/etools/pull/80/commits/ccb3a32b042fbd231a27edc2eb95cf734f8d2ba1

EquiTrack/reports/models.py
ilcsik: Removed in https://github.com/unicef/etools/pull/80/commits/bbc8427a1e81e96d0d59cfe9dc7d8d8fbaade594

EquiTrack/requirements/base.txt
ilcsik: That's handled in the related PR https://github.com/unicef/etools/pull/81

EquiTrack/reports/models.py
ilcsik: Added in https://github.com/unicef/etools/pull/80/commits/076510a5fca8b349c010d0d80286b3690dd44b11

EquiTrack/reports/models.py
ilcsik: Ok, removed in https://github.com/unicef/etools/pull/80/commits/bca3403b62c6715d0f2d037b7a43cc036ef2571a

EquiTrack/EquiTrack/factories.py
ilcsik: There's already a fixture for this and it gets loaded during test as well.

Utilizing those during tests is done in https://github.com/unicef/etools/pull/80/commits/80cce3ccd1154a08fab82df10779367a05c2d147

EquiTrack/reports/models.py
ilcsik: Ok, refactored in https://github.com/unicef/etools/pull/80/commits/613bf9f604628ffbcd845d636566d8542a2a469f

EquiTrack/EquiTrack/factories.py
robertavram: I see what you did, it makes sense, however .. we should never create a Result other than "Activity, Output, Outcome".. so in the ResultTypeFactory .. try to get those from the db, or create if they don;t exist...  
get ResultType random from [Output, Outcome, Activity] ->  if DoesNotExist create.

ilcsik: Yes, currently I don't use this Factory but doing what you suggested during test setup, using the fixture ids:

https://github.com/unicef/etools/pull/80/commits/80cce3ccd1154a08fab82df10779367a05c2d147

I have yet to figure out how to do this properly with factory boy's Faker wrapper - at first sight there seem to be inconsistencies regarding functionality in the built-in Faker and the most recent standalone Faker package, but I'll dig into this.

robertavram:
This looks good... can you make a PR against develop with this and update from develop + fix conflicts 👍

ilcsik:
Done in https://github.com/unicef/etools/pull/109

PR #94, opened 2016-08-15
Fix Agreement start end dates

ntrncic:


EquiTrack/partners/forms.py
```diff
@@ -340,6 +340,19 @@ def clean(self):
                     )}
                 )
 
+            # set end date to result structure end date
```
robertavram: basically you just moved it in the "agreement is PCA" condition.. right?

ntrncic: yup, exactly!

PR #109, opened 2016-08-24
Feature results api

ilcsik:


robertavram:
Hey Zoli,

Notice that we've done some model changes as well... so there are some conflicting migrations...
could you make sure the migration numbers follow after the latest migrations that exist in develop

ilcsik:
I resolved the conflicts with the `--merge` option in https://github.com/unicef/etools/pull/109/commits/18b8bebfde3d8b00839df9f05d36dbffaf362fbf

There was an other conflict under `partners` which is not related to this branch or my modifications, I hope it's ok to resolve it here and it won't clash anything.

robertavram:
made some migration tweaks..
it should be fine on your end if in your local db you rollback partners to 64,

and then rollback reports to 18... 
then run migrations again so that reports 19 runs again

PR #123, opened 2016-08-29
Add responsible field to Result model

ilcsik:
PL15

### DIFF
--- a/EquiTrack/reports/migrations/0021_merge.py
+++ b/EquiTrack/reports/migrations/0021_merge.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('reports', '0020_result_responsible'),
+        ('reports', '0020_indicator_assumptions'),
+    ]
+
+    operations = [
+    ]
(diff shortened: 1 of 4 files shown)


EquiTrack/reports/models.py
```diff
@@ -118,6 +118,7 @@ class Result(MPTTModel):
     geotag = models.ManyToManyField(Location)
     prioritized = models.BooleanField(default=False)
     metadata = JSONField(null=True, blank=True)
+    responsible = models.ForeignKey(UserProfile, null=True, blank=True)
```
ntrncic: Change foreign key to User model instead of UserProfile. Make sure the migrations sequence is correct.

ilcsik: Done in https://github.com/unicef/etools/pull/123/commits/a0b95f9000456199aa68b870d5a3bec425393e31

EquiTrack/reports/migrations/0021_merge.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
```
robertavram: let's try to move away from merges... see if you can manually sequence these based on what's existing on the develop branch

ilcsik:
Do we still need this field? We have reponsible_persons now on ResultWorkplanProperties.

robertavram:
@ilcsik you're correct we have this in the resultworkplanproperties

PR #124, opened 2016-08-29
feature add result type subactivity

ilcsik:
PL19

Do we need to do anything else other than letting users put an activity to parent and use SubActivity as type? I can see there's some logic based on ResultType in `ProgrammeSynchronizer._save_records(self, records)`.

### DIFF
--- a/EquiTrack/reports/fixtures/initial_data.json
+++ b/EquiTrack/reports/fixtures/initial_data.json
@@ -1,17 +1,22 @@
 [
-  {
-    "pk": 1,
-    "model": "reports.resulttype",
-    "fields": {"name": "Outcome"}
-  },
-  {
-    "pk": 2,
-    "model": "reports.resulttype",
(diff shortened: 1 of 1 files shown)


EquiTrack/reports/fixtures/initial_data.json
```diff
@@ -1,17 +1,22 @@
...
+    },
+    {
+        "pk": 3,
+        "model": "reports.resulttype",
+        "fields": {"name": "Activity"}
+    },
+    {
+        "pk": 4,
+        "model": "reports.resulttype",
+        "fields": {"name": "SubActivity"}
```
ntrncic: change SubActivity to Sub-activity

ilcsik: Done in https://github.com/unicef/etools/pull/124/commits/5b00b856a063065f954af30c5b5fe69e856dc85a

PR #127, opened 2016-08-30
Workplan comments

tejfel:


EquiTrack/workplan/models.py
```diff
@@ -0,0 +1,17 @@
+
+from django.db import models
+from django.contrib.auth.models import User
+from django.contrib.contenttypes.fields import GenericForeignKey
```
robertavram: hmm... good idea using generic relations... with Django tenant schemas i believe the contentType model is on the public schema while the workplan model will be on the tenant... this will create some ugly conflicts... see if there is a way to move the contenttype table on the tenant... let me know what you think

tejfel: I changed the `settings/base.py` based on the installation notes. The `django.contrib.contenttypes` app should be in the `TENANT_APPS` too: [LINK](http://django-tenant-schemas.readthedocs.io/en/latest/install.html#configure-tenant-and-shared-applications)

PR #129, opened 2016-09-02
Partner hact update

ntrncic:


robertavram:
No way.. this is terrible

PR #165, opened 2016-10-03
fix agreement and intervention sequence reference number

ntrncic:


robertavram:
Let's park this until we have a detailed discussion.

PR #175, opened 2016-10-06
Add trigram indexes on Location to speed up autocomplete

ilcsik:
Tested on 10k records, LIKE (__contains) operation went from 12ms to 0.6ms, roughly 20x speedup.


robertavram:
also, test is failing because of the "/" in the branch name.

ilcsik:
Renamed and docs added in https://github.com/unicef/etools/pull/182

PR #182, opened 2016-10-07
Add trigram indexes on Location to speed up autocomplete

ilcsik:
Basically I searched for "postgres autocomplete" in google, and between the 3 options, trigram index seems to be the best practice (LIKE, full text, LIKE with trigram).

How it works, you just simply need to add the extension and then the index on a given column and after that LIKE operations will be much faster and you can also have fuzzy results, which LIKE and fulltext search wont have. Good thing is that the implementation doesnt need to change since it uses the LIKE operator so integation with django is seamless, we already using __icontains for the query.

Trigram will be supported in django 1.10 so that you can fine tune it from django code:

https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/search/#trigram-similarity

Postgres docs:

https://www.postgresql.org/docs/9.5/static/pgtrgm.html

Couple of articles I ran into:

http://rogueleaderr.com/post/50576527241/postgres-fuzzy-search-using-trigrams-django
http://blog.scoutapp.com/articles/2016/07/12/how-to-make-text-searches-in-postgresql-faster-with-trigram-similarity
https://about.gitlab.com/2016/03/18/fast-search-using-postgresql-trigram-indexes/


EquiTrack/locations/migrations/0007_auto_20161004_1655.py
```diff
@@ -0,0 +1,16 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('locations', '0006_auto_20160229_1545'),
+    ]
+
+    operations = [
+        migrations.RunSQL("CREATE EXTENSION IF NOT EXISTS pg_trgm WITH SCHEMA pg_catalog"),
+        migrations.RunSQL("CREATE INDEX IF NOT EXISTS index_locations_on_name_trigram on locations_location USING gin (name gin_trgm_ops)"),
+    ]
```
robertavram: can we add a reverse to this in case we need to roll migrations back?

ilcsik: Sure, done in https://github.com/unicef/etools/pull/182/commits/df3df2cb18876817261de88572a929adaaaa5dd2

PR #183, opened 2016-10-07
Add etag header for Locations

ilcsik:
Tried not to reinvent the wheel and use a 3rd party lib for this (since django's default etag cannot be used with rest framework), and spent quite some time to get it work when realized that there's a bug report for the similar issue I have (no cache invalidation on update).

https://github.com/chibisov/drf-extensions/issues/120

So given no other lib can be found for this, I went with a custom but simple implementation.


EquiTrack/locations/models.py
```diff
@@ -151,6 +155,15 @@ class Meta:
         ordering = ['name']
 
 
+@receiver(post_delete, sender=Location)
+@receiver(post_save, sender=Location)
+def refresh_locations_etag(sender, instance, **kwargs):
```
ntrncic: We don't need to set the tag when something changes in the db, we will just remove the tag. At time of request we will update the e-tag in cache.

EquiTrack/locations/views.py
```diff
@@ -44,6 +45,21 @@ class LocationsViewSet(mixins.RetrieveModelMixin,
     queryset = Location.objects.all()
     serializer_class = LocationSerializer
 
+    def list(self, *args, **kwargs):
+        """
+        Returns list of instances only if there's a new ETag, and it does not
+        match the one sent along with the request.
+        Otherwise it returns 304 NOT MODIFIED.
+        """
+        etag = cache.get("locations-etag")
```
ntrncic: Check if etag exists in cache first, if it doesn't then set it here otherwise get it from cache.

EquiTrack/locations/models.py
ilcsik: I have no objections :) But how is that better than updating the tag right away when data changes? Update or just delete, doesn't really matter, we still have 1 cache operation.

And if we update here, we don't need to do that explicitly on the view.

EquiTrack/locations/views.py
ilcsik: This way we would have 3 operations here instead of 2.

- create if not exist
- check if matches with the requested one, respond with 304 or
- respond with new data

robertavram: the Cache is something that is not guaranteed to persist... we need the following operations on the view:
```
check if etag exists in the cache,
if it doesn't exist:
  set etag on cache,
  get the locations from db
  return locations with newly created etag:

if etag exists in cache:
   check if etag present on the request
         if yes and equal to our cache tag:
             return 304
        if no or if yes and not equal:
            return locations and set new etag on the response

```
On the locations save or update there is no need to set the etag in the cache, we only invalidate it at that point, basically remove the key from tag altogether.

ilcsik: Ah yeah that's true I overlooked that case! Thanks, I'll start with this on Monday.

robertavram:
this looks good so far.

PR #201, opened 2016-10-26
Change locations api

ilcsik:
Work in progress, want to discuss a couple of questions:
- Create a light serializer that return only the pcode and name: 
  Done
- Create a heavy serializer that returns all location fields: 
  Done. I used the old one without renaming to "heavy" so we don't need to refactor where it is used. Haven't added any extra fields, there's only "geom", which is used in property method. I also haven't added the 2 property methods, is this OK?
- Make endpoints for Get and List for both serializers:
  Done. Do we need the etag cache aside mechanism on both heavy and light listings?
- Location Auto-complete function user Light Serializer endpoint:
  Done.
- Location get by pcode or pk with Heavy Serializer:
  Done but nasty. The original implementation here used p_code in place of pk as lookup but returned a list, not a single object. Is this the desired behavior, or should it be a single object (is p_code unique)?

If p_code is unique, to be able to query both ways (pk and p_code) it needs to be indicated somehow, either different url, like "/locations/PCODE1/p_code", or a query parameter on the same url which I currently use, like /locations/PCODE1/?lookup_field=p_code. Which one do you prefer? The latter is a bit nasty since we are passing the pcode in place of the pk.
- Change from Viewsets to custom Views and endpoints:
  This is not yet done. Will be refactored so that we don't use viewsets and routers.

Thanks.


ntrncic:
The geom field should be added to heavy serializer. 
Please add etag for both light and heavy serializer
Pcode should be unique but we do not enforce it at the moment. To filter for one location please use PK. To filter by pcode should pass in a query parameter "pcode="

robertavram:
get should always use heavy serializer.. sorry for the confusion
autocomplete should use light serializer

@ntrncic @ilcsik we should enforce uniqueness on pk codes... let's have a chat on how to handle existing data tomorrow

@ilcsik /locations/[id] and /locations/pcode/[pcode] should do for now

ilcsik:
Done in https://github.com/unicef/etools/pull/201/commits/052e3dd7f6bfc177d857acb0a9872f9edeaddaca

As for moving away from viewsets here, shall we just wait and check what's gonna be the best practice on Agreements as discussed?

robertavram:
@ilcsik if this is ready... remove [WIP] from the header and assign it to me and Nik

ilcsik:
Done

robertavram:
please assign to me and Nik, please update from develop

ilcsik:
Updated from develop. You're already assigned. Thanks!

robertavram:
@ilcsik Assign users by clicking the gear on assignees...
assigning

all good though...

EquiTrack/locations/views.py
```diff
@@ -36,15 +38,36 @@ class LocationTypesViewSet(mixins.RetrieveModelMixin,
...
-class LocationsViewSet(mixins.RetrieveModelMixin,
+class ETagMixin(object):
+
+    def etag_cache_list(self, cls, *args, **kwargs):
+        schema_name = connection.schema_name
+        cache_etag = cache.get("{}-locations-etag".format(schema_name))
+        if not cache_etag:
+            new_etag = uuid.uuid4().hex
+            response = super(cls, self).list(*args, **kwargs)
+            response["ETag"] = new_etag
```
robertavram: I believe we also have to use patch_cache_control unless cache-control is set by one of our middleware:

https://docs.djangoproject.com/en/1.9/ref/utils/#django.utils.cache.patch_cache_control

also.. instead of uuid I think you can use set_response_etag(response) from django.utils.cache. check it out... technically that's the way to go... theoretically that way we could prevent data transmission even if we don;t store our etag in cache.. however the other reason we're doing this is to prevent the extra query and serialization of the response

ilcsik:
@robertavram ah yeah I did a quick look and you were all there but that's just the participants, sorry. Adding you as assignees.

EquiTrack/locations/views.py
ilcsik: @robertavram yeah that's what I tried and the thing is that set_response_etag meant to be used in the middleware _after_ the response is rendered - it throws an error regarding access to its data, saying that the response needs to be rendered before accessing the data (of course I tried render() too but that threw a different error then :))

ilcsik: Cache control added in https://github.com/unicef/etools/pull/201/commits/1d831242fee3eb7dc6399de4935798166ef278ee

Issue #215, opened 2016-11-07
Review all pypi packages for potential upgrades and whether being used


chachra:
Example doesn't look like django-redis-cache is currently used ?

n1207n:
PR for this issue is closed. [https://github.com/unicef/etools/pull/231](url)

caktus-philip:
Confirmed that PR #231 was merged

PR #231, opened 2016-11-09
Python package review and update [#215]

n1207n:
This PR addresses the PyPi dependency reviewing and version updates.

PyPi requirements.txt should be updated to corresponding latest version except the following packages:
* django-autocomplete-light (v3.x package has a different python package naming)
* GDAL (Incompatibility with compiling at GDAL >= 1.10.0)
* celery (django-celery libraries are not compatible with latest celery yet)
* azure-storage (Lack of Azure/production environment access to test this out)


EquiTrack/requirements/base.txt
```diff
@@ -1,66 +1,69 @@
...
-requests==2.8.1
-django-analytical==2.0.0
-xlrd==0.9.3
-dj-static==0.0.6
-django-easy-pdf
-xhtml2pdf==0.0.6
-html5lib==0.9999999 #must be added after xhtml2pdf
-reportlab
+sqlparse==0.2.2  # for django debug toolbar to not break admin
+azure-storage==0.33.0
```
robertavram: have you tested this version of azure-storage?

n1207n: Not yet. However now that I think about it, I don't think I can fully test this library without azure and/or production access.

I'll revert to old version for azure-storage.

PR #247, opened 2016-11-11
Squash migrations

ntrncic:


robertavram:
#247

robertavram:
@ntrncic please update your comment to include all fake migrations for all apps....

ntrncic:
@robertavram some apps are dependent on others therefore only the 5 mentioned need to be run. 
if you run the other apps it just says no migrations to be applied.

robertavram:
cool.. waiting for tomorrow afternoon... let's try together

Issue #244, opened 2016-11-11
Remove django filer


chachra:
Only used here it seems like: https://github.com/unicef/etools/blob/master/EquiTrack/partners/utils.py#L5

Remove from requirements, settings, everywhere.

Migrations use it, but they are being squashed anyway in #214

caktus-philip:
Confirmed that this has been removed

Issue #219, opened 2016-11-07
Upgrade DRF to 3.5.X


chachra:
Depends on #215 of course, but opening separate one here since it'll take some effort from current DRF version being used.

chachra:
@n1207n can you make sure all tests are running in project george branch?

n1207n:
project-george-package-updates branch has the DRF 3.5 version now. Django unit tests are fixed as well.

caktus-philip:
Current version in eTools is 3.5.3 so it looks like this was implemented.

PR #248, opened 2016-11-11
Project george error fixes from package updates

n1207n:
This PR addresses the following changes:
* Reverting some PyPi package versions
    * django-reversions
    * html5lib
* New fix changes for djangorestframework and restframeworkswagger
* DebugToolMiddleware enalbing in local settings
* Removed django-filer package usage
* Django unit test fixes

The corresponding issues:
* #219 
* #244


EquiTrack/requirements/base.txt
```diff
@@ -25,9 +25,8 @@ django-celery==3.1.17
 django-celery-email==1.1.5
 django-suit==0.2.21
 django-suit-ckeditor==0.0.2
-django-filer==1.2.5
 django-import-export==0.5.1
-django-reversion==2.0.7
+django-reversion==1.10.1
```
robertavram: @n1207n, why are we reverting reversion and html5lib

n1207n: @robertavram 

html5lib==1.0b10 breaks xhtml2pdf dependency with `ImportError: cannot import name inputstream`, therefore xhtml2pdf is not ready for html5lib regression.

django-reversion has revamped its low level APIs and new Django migrations. And specifically, the new Django migrations break with our tenant schema with `django.db.utils.ProgrammingError: column "manager_slug" of relation "reversion_revision" does not exist`

For these reasons, I decided to keep old versions for django-reversion and html5lib.

robertavram: sounds good.. i believe that reversion can be updated after we clean up migrations...  i think we're getting away from reversion anyway... thanks

Issue #256, opened 2016-11-16
HACT dashboard doesn't show current year trips


kasfaw:
The HACT dashboard should only show trips completed for the current year...not for the entire country programme

ntrncic:
in prod!

PR #259, opened 2016-11-16
fix hact trip calculation for current year

ntrncic:


ntrncic:
solves #256

robertavram:
looks good.. did you test?

EquiTrack/partners/models.py
```diff
@@ -337,13 +337,11 @@ def cp_cycle_trip_links(self):
         crs = CountryProgramme.current()
         if self.partner_type == u'Government':
             return self.linkedgovernmentpartner_set.filter(
-                        trip__from_date__lt=crs.to_date,
-                        trip__from_date__gte=crs.from_date
+                        trip__from_date__year=datetime.datetime.now().year,
```
robertavram: store this (`datetime.datetime.now().year,`) in a var under line 337 `crs = ` and use it from there

ntrncic:
@robertavram yes tested in my local.

PR #263, opened 2016-11-17
staging-step2

robertavram:
fix hact trip calculation for current year


robertavram:
do not merge until stage 1 has been merged!

PR #268, opened 2016-11-18
Staging-step1-squash-migrations

robertavram:
Develop


robertavram:
@ntrncic just so you know.. you also neeed to run the following:
python manage.py migrate_schemas trips --fake --noinput
python manage.py migrate_schemas  vision --fake --noinput
python manage.py migrate_schemas workplan --fake --noinput

PR #279, opened 2016-11-24
Partners api v2

ntrncic:


ilcsik:
Hi, what are the actual tasks here? Thanks.

ntrncic:
@ilcsik do new serializers, the filters are there but need checking and refining, adding of new partner staff members. Please check the requirements document in case I missed something.

EquiTrack/partners/views/views_v2.py
```diff
@@ -0,0 +1,141 @@
...
+        if self.request.method == "GET":
+            query_params = self.request.query_params
+            if "format" in query_params.keys():
+                if query_params.get("format") == 'csv':
+                    return PartnerOrganizationExportSerializer
+            return PartnerOrganizationSerializer
+        else:
+            return super(PartnerOrganizationListAPIView, self).get_serializer_class()
+
+    def create(self, request, *args, **kwargs):
```
ilcsik: Do we need this create method defined explicitly? Isn't this the same as the default implementation?

ilcsik:
_do new serializers, the filters are there but need checking and refining_

Done in https://github.com/unicef/etools/pull/279/commits/a051be2dd26cc46220797a3913460c0076c8102f, plus added unit tests.

ilcsik:
Blocked field added in https://github.com/unicef/etools/pull/279/commits/65663b1b4e28d7a1b03ef403f6ed3d4b64af26d0

ilcsik:
From the serializer:
        # TODO add missing fields:
        #   Bank Info (just the number of accounts synced from VISION)

Can I have some more explanation on this? Is this a simple counter? How do I suppose to implement?

ilcsik:
_adding of new partner staff members_

Can I have some more explanation here? Currently we have all the endpoints (list, create, retrieve, update) on PartnerStaffMember views, /staff-members and staff-members/pk.

PartnerStaffMember has an FK to PartnerOrganization so we also have the 'staff_members' related field on the PartnerOgranization model, to list members on partner.

What else needs to be implemented?

ntrncic:
@ilcsik we are still not sure if bank info will be provided to us, please leave it for now.
For staff members, they will be added through the partner details page. So when the Partner save api is called and there are partner staff members in the form then it should call the partner staff api.

ilcsik:
So we agreed with Nik that we should support adding staff members through separate calls (this is already in place), and through Partner API as embedded list as well.

ilcsik:
I migrated partnerstaff-api-v2 since I need it for staff members feature. Changed the base of the PR as  well.

ilcsik:
@ntrncic Staff member nested creation and update (add/remove) done in https://github.com/unicef/etools/pull/279/commits/2702a1881e9e754396d23cf554cb005cb49be16a and https://github.com/unicef/etools/pull/279/commits/279f6b1a49c8bd9cf026862ee5042ab11a77d11f

Can you please check?

n1207n:
This PR looks great to me so far. ;)

ilcsik:
Resolves #303

Issue #285, opened 2016-12-05
modify Result Api Endpoint 


robertavram:
one get endpoint that takes in and returns:

nothing in -> all outputs serialized with result-type as string (result_type.name) (not ids)

country_programme (values: year : String) -> all results that apply for that year's country programme

result_type in (string values [Output, Activity, Outcome]) -> all results filtered by that result type

filters can be combined.

robertavram:
@ilcsik can you give this a shot? v2 of the current api endpoint to do this
https://etools-staging.unicef.org/api/reports/results/

robertavram:
by default unless year is specified it will return only the results within the current country programme.

caktus-philip:
Code has changed somewhat since the original PR but this need is satisfied

PR #293, opened 2016-12-07
Result api v2

ilcsik:
Added functionality from #285 

@robertavram Do we only need this listing endpoint in the v2, or all the other create/update etc? 

In the latter case it needs some discussion regarding which nested fields needs create/update, in the former I'll remove all the other endpoints.


EquiTrack/reports/serializers/v2.py
```diff
@@ -0,0 +1,34 @@
+from rest_framework import serializers
+
+from workplan.serializers import ResultWorkplanPropertySerializer
+from reports.serializers.v1 import MilestoneSerializer
+from workplan.models import ResultWorkplanProperty
+from reports.models import Result
+
+
+class ResultCreateSerializer(serializers.ModelSerializer):
```
ntrncic: Anything in the workplan app is not going to be used just yet. We are only looking and the Result Model in the reports app.

ntrncic:
@ilcsik we won't be updating the Result model through the api. We are updating Results via a task nightly that pulls from another web service. So all we need is to return a list with the filters specified.

ilcsik:
Endpoints and fields removed in https://github.com/unicef/etools/pull/293/commits/e7c568ab893a862597e489159a3e22d02f2e9ef6

EquiTrack/reports/views/v2.py
```diff
@@ -0,0 +1,35 @@
...
+        current_cp = CountryProgramme.current()
+        q = Result.objects.filter(country_programme=current_cp)
+
+        query_params = self.request.query_params
+        if query_params:
+            queries = []
+
+            if "country_programme" in query_params.keys():
+                cp_year = query_params.get("country_programme")
+                cp = CountryProgramme.by_year(cp_year)
```
ntrncic: can we filter this way: country_programme__from_date__year=cp_year?

EquiTrack/reports/tests/test_views.py
```diff
@@ -4,9 +4,17 @@
...
 from rest_framework import status
 
-from reports.models import ResultType
-from EquiTrack.factories import UserFactory, ResultFactory, ResultWorkplanPropertyFactory
-from EquiTrack.factories import SectionFactory, LocationFactory, WorkplanFactory, ResultStructureFactory
+from reports.models import ResultType, CountryProgramme
+from EquiTrack.factories import (
+    UserFactory,
+    ResultFactory,
+    ResultWorkplanPropertyFactory,
```
ilcsik: Removed in https://github.com/unicef/etools/pull/293/commits/cbfb2668f85068e8cea5df05a9daf37e7837dfb6

EquiTrack/reports/views/v2.py
ntrncic: @ilcsik can't we just do the filter here and not call the class method?
With the class method you make one call to the database. Then here you make another call, if we have the filter here directly then t'sonly one call to the database. This might be faster, let me know what you think.

ilcsik: I think we cannot apply the order_by part then since we're constructing Q objects and don't really have QuerySet there. Do we need this at all (also used in current(), I copied this behaviour from there)? If not, we're golden.
```python
.order_by('-to_date').first()
```

ilcsik: After all, in this case we are interested in the presence of ANY country programme in a given year so we might not need this here so I'm removing.

ntrncic:
@n1207n please review PR

EquiTrack/reports/views/v2.py
```diff
@@ -0,0 +1,36 @@
...
+from django.db.models import Q
+from rest_framework.generics import ListAPIView
+from rest_framework.permissions import IsAdminUser
+
+from reports.models import Result, CountryProgramme
+from reports.serializers.v2 import ResultListSerializer
+
+
+class ResultListAPIView(ListAPIView):
+    queryset = Result.objects.all()
```
n1207n: This is redundant as we have get_queryset method.

EquiTrack/reports/urls_v2.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import url
+
+from reports.views.v2 import ResultListAPIView
+
+
+urlpatterns = (
+    url(r'^reports/results/$', view=ResultListAPIView.as_view(), name='report-result-list'),
```
ntrncic: please add retrieve by id also.

EquiTrack/reports/views/v2.py
```diff
@@ -0,0 +1,36 @@
...
+    def get_queryset(self):
+        current_cp = CountryProgramme.current()
+        q = Result.objects.filter(country_programme=current_cp)
+
+        query_params = self.request.query_params
+        if query_params:
+            queries = []
+
+            if "country_programme" in query_params.keys():
+                cp_year = query_params.get("country_programme")
```
n1207n: Always provide the default value when using .get() method.

ntrncic:
@ilcsik please checkout silin's comments

EquiTrack/reports/views/v2.py
ilcsik: Done in https://github.com/unicef/etools/pull/293/commits/b8cc40160c642175378ca42b00e871919d8fb5e7

EquiTrack/reports/views/v2.py
ilcsik: Do you think it's still necessary despite that the key is checked before?

Done in https://github.com/unicef/etools/pull/293/commits/b8cc40160c642175378ca42b00e871919d8fb5e7

ilcsik:
Resolves #285

robertavram:
hey @ilcsik can you update this from develop... mark ready when ready... thanks

ilcsik:
@robertavram ready.

robertavram:
@ilcsik fix migrations by removing them and adding them again.. keeps it clean. we're staying away from merge unless things got merged like that in prod

robertavram:
@ilcsik thanks,, this should be good.. we'll merge soon

PR #300, opened 2016-12-14
Remove activityinfo

ilcsik:
I modified the initial migration of reports app since it depends on activityinfo so otherwise I cannot delete activityinfo app. Is this OK?


ntrncic:
No we can't modify the initial migration in reports. Can we just remove the activity info models and leave the app there?

ilcsik:
@ntrncic yeah I guess the best we can do then is to remove the FK from reports, and then remove all models from activityinfo but leave the app in the settings and the migration dependency there. Sounds good?

ntrncic:
@n1207n do you have any idea how to remove an app and not mess up the initial migrations that have dependencies to it? We are trying to remove the activity info app from etools. Please check this PR.

robertavram:
@ilcsik @ntrncic Let's modify the initial migration...  but let's do this:
Split this PR in two parts:
1. add migration to remove all activity info dependent fields and all functions that populate those
We will merge this PR and run that migration then fake it back
2. add second part that does not include the migration but has the initial migration modified to never add those fields. (basically what this PR is doing now)
DONE... make sure to label the PRs as "Part 1" "Part 2"

PR #311, opened 2016-12-19
Government Intervntions API v2

ilcsik:


EquiTrack/partners/models.py
```diff
@@ -1318,11 +1330,11 @@ def save(self, **kwargs):
         if self.pk:
             prev_result = GovernmentInterventionResult.objects.get(id=self.id)
             if prev_result.planned_amount != self.planned_amount:
-                PartnerOrganization.planned_cash_transfers(self.intervention.partner, self)
```
ilcsik: @ntrncic can you please help out here regarding business logic? I removed self from the parameters here since it passes GovernmentIntervention and it expects a budget_log in line 332, and so it doesnt have partnership parameter. But now that I removed, 1 tests fails if you ran it due bad calculations so I don't really know how this piece should work.

ilcsik: @robertavram In absence of Nik can you please check the question above when you have a chance? Thanks

robertavram: hey @ilcsik... So... this code here is to update the json field that calculates the hact dashboard.... basically if someone updates the planned_amount on a government intervention this updates the PartnerOrganization (in this case it will be a government partner) HACT dashboard. the reason it passes self twice has to to with the way Nik designed the logic on it. You can leave it in for now.. We will double check the logic... HACT dashboard logic will need to be re-written since the trips and intervention applications will be updated

ilcsik:
I still have 2 bugs regarding testing. One is when I create Section for GovernmentInterventionResult it won't be available for the backend during GIR create since Section has some tenant related filter in its manager and won't show up. I wasn't able to resolve this yet.

Also, Permissions will only be changed after migrate because we have them implemented on other PR but the v2 views and serializers will be conflicting heavily among 3 PR and don't want to resolve it here

ilcsik:
Test bug fixed, develop merged. Ready for review

robertavram:
@ilcsik please update this from develop

PR #314, opened 2016-12-19
T2f fixes

tejfel:


EquiTrack/t2f/models.py
```diff
@@ -236,44 +263,86 @@ def plan(self):
...
+        html_content = render_to_string(template_name, context)
+
+
+        # TODO what should be used?
+        sender = ''
+        msg = EmailMultiAlternatives(subject, '',
+                                     sender, [recipient])
+        msg.attach_alternative(html_content, 'text/html')
+
+        # Core mailing is broken. Multiple headers will throw an exception
```
robertavram: can you not place the image as a url?

tejfel: then it should stay on the server forever. If we attach them, then the email will always look same - if you remove any file from server, will not affect the mails. And attached images considered more secure as far as I know.

EquiTrack/t2f/models.py
```diff
@@ -236,44 +263,86 @@ def plan(self):
...
+        from t2f.serializers.mailing import TravelMailSerializer
+        serializer = TravelMailSerializer(self, context={})
+
+        url = reverse('t2f:travels:details:index', kwargs={'travel_pk': self.id})
+        approve_url = reverse('t2f:travels:details:state_change', kwargs={'travel_pk': self.id,
+                                                                          'transition_name': 'approve'})
+        approve_certification_url = reverse('t2f:travels:details:state_change',
+                                            kwargs={'travel_pk': self.id,
+                                                    'transition_name': 'approve_certificate'})
+        context = Context({'travel': serializer.data,
```
tejfel: With this approach all logic stays in the code and not in the templates. Could be changed if you wish.

PR #328, opened 2017-01-04
Feature - notification service

n1207n:
This PR addresses a new Django feature for handling notification logic and records.
* New models
* New Python logic to handle notification


n1207n:
@ntrncic Let me know if there are any further comments you have. I'll be happy to address them as soon as possible. Thanks! :)

PR #332, opened 2017-01-06
Fix tests on develop

ilcsik:


robertavram:
@ilcsik merging this... keep at it in the meantime :)

PR #336, opened 2017-01-10
admin changes

ntrncic:


EquiTrack/partners/models.py
```diff
@@ -1603,15 +1603,22 @@ class GovernmentInterventionResult(models.Model):
         verbose_name='Unicef focal points',
         blank=True
     )
+    # TODO remove field after adding sectors field
```
robertavram: remove it in the same migration... but run script to move sector to sectors before.

ntrncic:
@robertavram done! updated current migration to remove sector and section fields

PR #348, opened 2017-01-19
django rest raml

ilcsik:
Toying with RAML, basically the modern swagger. Too bad that I have a JS error so it wont work (yet).


ilcsik:
I found an issue filed regarding this, its already merged to develop but I cant see it on master so...

Also this is only in case of displaying some api error, so it might be that RAML couldnt handle our API but we cannot see the reason due to this issue.

Might try to figure out later.

robertavram:
@ilcsik if this is not suitable anymore.. please close the PR

PR #353, opened 2017-01-25
Update agreement partnership to pdf generation

ntrncic:


robertavram:
add banking details migration to partner

PR #357, opened 2017-01-26
Feature - API for Activity Stream with custom data

n1207n:
This PR addresses a custom activity stream feed which includes action's custom data entry such as snapshot entry for Agreement model.


n1207n:
@robertavram @ntrncic 

Hi,

This PR is ready for review as the custom activity stream feed class seems functional at this point. The custom feed class does returns Activity stream's custom data such as model snapshot if there is one available. Moreover, I trimmed the json item formatting so that only relevant information is there.

This is an example response from custom activity stream.
screen shot 2017-01-26 at 1 06 58 pm

Let me know if you have any comment or further requests on activity stream feed. Thanks!

robertavram:
@n1207n label ready when done

n1207n:
@robertavram,

I just pushed the commit to address an improvement we discussed. From now on, create_snapshot_activity_stream classmethod should create two custom data to Action model instance: `previous` and `changes`.

`previous` data entry should contain the dictionary format of a model instance in previous state.
`changes` data entry is a dictionary of all mutated field values.

The relevant unit test case has been updated too.

robertavram:
@n1207n thanks

PR #372, opened 2017-01-31
T2f second iteration - part 2

tejfel:


EquiTrack/publics/migrations/0007_auto_20170131_1746.py
```diff
@@ -0,0 +1,40 @@
...
+            model_name='currency',
+            name='code',
+            field=models.CharField(max_length=5),
+        ),
+        migrations.AlterField(
+            model_name='dsaregion',
+            name='area_name',
+            field=models.CharField(max_length=120),
+        ),
+    ]
```
ntrncic: please try and merge the migrations?

EquiTrack/publics/models.py
```diff
@@ -3,22 +3,37 @@
...
+
+
 class TravelExpenseType(models.Model):
+    # TODO simon: explain what's this line here
     USER_VENDOR_NUMBER_PLACEHOLDER = 'user'
 
     title = models.CharField(max_length=32)
     vendor_number = models.CharField(max_length=32)
     is_travel_agent = models.BooleanField(default=False)
+    rank = models.PositiveIntegerField(default=100)
```
ntrncic: why is the default 100?

EquiTrack/publics/models.py
```diff
@@ -3,22 +3,37 @@
 from django.db import models
 
 
+class TravelAgent(models.Model):
```
ntrncic: where are we using this? I don't see any relations

EquiTrack/users/migrations/0003_auto_20170123_2009.py
```diff
@@ -58,11 +58,11 @@ class Migration(migrations.Migration):
             name='supervisor',
             field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='supervisee', to='users.UserProfile'),
         ),
-        migrations.AddField(
-            model_name='userprofile',
-            name='vendor_number',
-            field=models.CharField(blank=True, max_length=32, null=True, unique=True),
-        ),
+        # migrations.AddField(
```
ntrncic: fully remove commented code.
why does this need to be moved to a new migration?

ntrncic:
@tejfel update from develop please

EquiTrack/publics/models.py
tejfel: The expense type is weakly related. There should be an expense type per travel agent. Rob mentioned back then to use only this relation to avoid unwanted extra queries. Can be changed to be a FK if you prefer that.

EquiTrack/publics/models.py
tejfel: To be able to put anything after or before. It's used for dropdown ordering.

EquiTrack/users/migrations/0003_auto_20170123_2009.py
tejfel: I put this back, I needed it for local only.

EquiTrack/publics/models.py
ntrncic: ok thanks for the clarification

EquiTrack/publics/models.py
ntrncic: why not start at 0?

tejfel: Because then if you want to put something in front, you have to update all to a bigger number. But can start if you think it's better.

EquiTrack/t2f/vision.py
```diff
@@ -0,0 +1,137 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def run_on_tenants(func):
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        original_tenant = connection.tenant
+        try:
+            for workspace in Workspace.objects.all():
```
robertavram: not all.. just workspaces that have a business_area_code

EquiTrack/users/models.py
```diff
@@ -57,6 +57,9 @@ class Country(TenantMixin):
     offices = models.ManyToManyField('Office', related_name='offices')
     sections = models.ManyToManyField('Section', related_name='sections')
 
+    threshold_tre_usd = models.DecimalField(max_digits=20, decimal_places=4, default=None, null=True)
```
robertavram: what is this field for?

EquiTrack/t2f/vision.py
```diff
@@ -0,0 +1,137 @@
...
+                connection.set_tenant(workspace)
+                func(self, workspace, *args, **kwargs)
+        finally:
+            connection.set_tenant(original_tenant)
+    return wrapper
+
+
+class InvoiceExport(object):
+    def __init__(self, invoice_list):
```
robertavram: this should be consistently named with what is being received (invoice_qs)

EquiTrack/publics/management/commands/load_initial_data.py
```diff
@@ -0,0 +1,663 @@
...
+                                                       long_name=long_name,
+                                                       vision_code=vision_code,
+                                                       business_area=BusinessArea.objects.filter(code=ba_code).first(),
+                                                       iso_2=iso_2,
+                                                       iso_3=iso_3,
+                                                       currency=Currency.objects.filter(code=currency_code).first(),
+                                                       valid_from=valid_from,
+                                                       valid_to=valid_to,
+                                                       threshold_tre_usd=0,
+                                                       threshold_tae_usd=0)
```
robertavram: your country model does not have these fields
THe users.Country model has the threshold fields.. btw.. we should have them on your Country object...
if you want we can foreign key from workspace model to your country model... thoughts?

robertavram: after speaking with K, it makes sense to keep these fields on the workspace model for now. just remove from here.

EquiTrack/publics/serializers.py
```diff
@@ -36,6 +36,10 @@ class Meta:
 
 
 class CurrencySerializer(serializers.ModelSerializer):
+    iso_4217 = serializers.CharField(source='code', read_only=True)
+    exchange_to_dollar = serializers.DecimalField(source='exchange_rates.all.0.x_rate', max_digits=20,
```
robertavram: this just takes the fist one in the list.. unless you have ordering on it... you might get one that is not valid... avoid using .0. anywhere.. as a best practice

EquiTrack/t2f/serializers/export.py
```diff
@@ -56,7 +56,7 @@ class TravelAdminExportSerializer(serializers.Serializer):
     dsa_area = serializers.CharField(source='dsa_region.area_code')
     overnight_travel = serializers.BooleanField()
     mode_of_travel = serializers.CharField()
-    airline = serializers.CharField(source='airlines.0.name')
+    airline = serializers.CharField(source='airlines.all.0.name')
```
robertavram: comment in why you're taking only the first one in the list..

EquiTrack/t2f/models.py
```diff
@@ -159,10 +159,11 @@ class Travel(models.Model):
 
     # When the travel is sent for payment, the expenses should be saved for later use
     preserved_expenses = models.DecimalField(max_digits=20, decimal_places=4, null=True, default=None)
+    approved_cost_traveler = models.DecimalField(max_digits=20, decimal_places=4, null=True, default=None)
```
robertavram: can you comment in the purpose and use of each of these fields?

tejfel: I revised preserved expenses. We need that for certification. I'll leave it there since it's updated on send for payment action and not on approval (basically we will compare preserved expenses with the actual expenses after certification to see if there is any difference).

EquiTrack/publics/serializers.py
```diff
@@ -37,7 +37,7 @@ class Meta:
 
 class CurrencySerializer(serializers.ModelSerializer):
     iso_4217 = serializers.CharField(source='code', read_only=True)
-    exchange_to_dollar = serializers.DecimalField(source='exchange_rates.all.0.x_rate', max_digits=20,
+    exchange_to_dollar = serializers.DecimalField(source='exchange_rates.all.last.x_rate', max_digits=20,
```
robertavram: is the sorting defined on the meta class? otherwise last won't work as expected

EquiTrack/t2f/serializers/export.py
```diff
@@ -56,7 +56,7 @@ class TravelAdminExportSerializer(serializers.Serializer):
     dsa_area = serializers.CharField(source='dsa_region.area_code')
     overnight_travel = serializers.BooleanField()
     mode_of_travel = serializers.CharField()
-    airline = serializers.CharField(source='airlines.all.0.name')
+    airline = serializers.CharField(source='airlines.all.last.name')
```
robertavram: same here

EquiTrack/t2f/vision.py
```diff
@@ -24,7 +24,7 @@ def run_on_tenants(func):
     def wrapper(self, *args, **kwargs):
         original_tenant = connection.tenant
         try:
-            for workspace in Workspace.objects.all():
+            for workspace in Workspace.objects.exclude(business_area_code=''):
```
robertavram: also exclude bubsiness_area_code__isnull

robertavram:
@tejfel also.. update from develop

PR #392, opened 2017-02-06
Feature - API endpoint for parternship dashboard

n1207n:
This PR addresses a new REST API endpoint which returns a set of aggregation data for Intervention and GovernmentIntervention objects.


anusenan:
#392

PR #440, opened 2017-02-10
Add filter for Partnership Managers

### LINKED ISSUE #343: Endpoint for unicef Partnership Managers
List of users that are in the partnership management group

ilcsik:
Resolves #343


EquiTrack/users/views.py
```diff
@@ -86,7 +87,7 @@ def get_queryset(self):
         return self.model.objects.filter(
             country=user.profile.country,
             user__is_staff=True
-        ).order_by('user__first_name')
```
robertavram: why are we removing the ordering?

ilcsik: Good catch! By mistake, first I started tinkering there but then I realized that it's another endpoint. Fixed.

EquiTrack/users/views.py
```diff
@@ -202,6 +203,10 @@ def get_queryset(self):
         # we should only return workspace users.
         queryset = super(UserViewSet, self).get_queryset()
         queryset = queryset.prefetch_related('profile', 'groups', 'user_permissions')
+        # Filter for Partnership Managers only
+        if "managers" in self.request.query_params.keys():
```
robertavram: check if the value of managers is true, also . field should be "partnership_managers"

ilcsik: Done in https://github.com/unicef/etools/pull/440/commits/c51a2b1af2b3a8287d8f08de65ad5426c09779d6

PR #456, opened 2017-02-15
Delete endpoints for results and activities for Government intervention

ntrncic:


EquiTrack/partners/views/government.py
```diff
@@ -144,3 +146,25 @@ def update(self, request, *args, **kwargs):
...
+
+class GovernmentInterventionResultDeleteView(DestroyAPIView):
+    """
+    Delete Government Intervention Result Activities.
+    Returns a .
+    """
+    serializer_class = GovernmentInterventionResultNestedSerializer
+    permission_classes = (IsAdminUser,)
+    filter_backends = (PartnerScopeFilter,)
+    queryset = GovernmentInterventionResult.objects.all()
```
robertavram: add some validation on delete.. make sure that the GovernmentIntervention has not been signed.. return error if it has. Also... Let's add a permission class that includes only partnershipMangers and Rep Office

PR #457, opened 2017-02-15
Add monitoring visits api

ilcsik:
@robertavram said:

Look at this screen:
https://projects.invisionapp.com/share/XQ90VSSC7#/screens/205455281
notice MonitoringVisits.
Build an api endpoint that takes in a partner organization ID, Gets all the travels with those fields.  Basically what you'll have to do is:
get all The travel_activities that connect to the Partner Oganization.
In order to get the travel you'll have to get all the Travels where the traveller is equal to TravelActivity primary_traveler

### DIFF
--- a/EquiTrack/t2f/filters/__init__.py
+++ b/EquiTrack/t2f/filters/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 from django.db.models.query_utils import Q
+from django.db.models import F
 from rest_framework.filters import BaseFilterBackend
 
 from t2f.serializers.filters import SearchFilterSerializer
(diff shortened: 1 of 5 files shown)


ilcsik:
@robertavram can you please check current implementation. Got a couple of questions yesterday.

It returns TravelActivites by Partner Organization ID and a list of related Travels from M2M field.

1, How the response format should look like? Currently 1 TravelActivity contains a list field for M2M travels. Should this be the way around, for e.g. expand all items and somehow merge Travel + TravelActivity info like on the wireframe?

2, Shall I further filter travels by primary traveler, does that make a difference (for e.g. are there Travels that are not related to primary_traveler?)

robertavram:
So.. the endpoint needs to retrieve a list of Travels
the list of travels you send to the frontend should look like this:
```
[
   {
    id: 1,
    reference_number: 123123,
    traveler: John Smith,
    travel_type: programmatic_visit,
    locations: {id: 1, p_code: 12, name:BlahBlah},
    status: completed
   },

]
```

in order to get this list...
1 only travel_activities that connect to the Partner Oganization
2. get all the travels connected to those travel_activities.. but only get the travels where the traveller is the primary traveller 

for example:
(don't think it works out of the box but around those lines)

```
Travel.objects.filter(travel_activity__primary_traveller=F(traveller), travel_activity__partner=mypartner)
```

ilcsik:
@robertavram OK thanks I understand, just wanted to check if it can be solved the way around, from travel activities (since most of the fields are related to travel activities).

ilcsik:
@robertavram I implemented the version discussed in slack. Please check: https://github.com/unicef/etools/pull/457/commits/c38020167fb39f7ad91366bfb314c27f51e54b3e

PR #460, opened 2017-02-16
Pmp delete api

ntrncic:


ntrncic:
@ilcsik please write tests for new delete api endpoints

PR #476, opened 2017-02-21
- add action point export

tejfel:


EquiTrack/t2f/tests/test_action_points.py
```diff
@@ -228,3 +231,23 @@ def test_invalid_status(self):
         response_json = json.loads(response.rendered_content)['action_points']
         self.assertEqual(response_json,
                          [{'status': ['Invalid status. Possible choices: cancelled, ongoing, completed, open']}])
+
+    @skip('')
+    def test_export(self):
```
ewheeler: while this tests that the exported column names are correct, it would be good to do better validation

created an [issue](https://github.com/unicef/etools/issues/487) to track this generally

for now, lets at least leave a `TODO` comment here so we can revisit

tejfel: Agree, the main problem (and the reason why it's skipped) because I still don't know how to send csv request. Can you help me with this maybe? Any time when I send, even if the `request_type='csv'` it returns json.

ewheeler: ahh ok. not sure what is needed. 
@robertavram any ideas?

robertavram: i believe there needs to be a format query param set to csv;
?format=csv
take a peek https://github.com/unicef/etools/blob/develop/EquiTrack/partners/views/interventions_v2.py#L158

robertavram:
@tejfel please update from base / fix conflicts. also did I answer your question?

PR #488, opened 2017-02-22
T2f fixes pt4

tejfel:


EquiTrack/t2f/models.py
```diff
@@ -22,6 +22,13 @@
 log = logging.getLogger(__name__)
 
 
+class TransitionError(Exception):
```
ewheeler: i think this particular custom exception should subclass `RuntimeError`

but probably fine for now until we document coding conventions/guidelines for the project as a whole. and maybe even a separate namespace for custom etools exceptions

tejfel: Done.

EquiTrack/t2f/models.py
```diff
@@ -175,6 +182,10 @@ def cost_summary(self):
     def check_completion_conditions(self):
         if self.status == Travel.SUBMITTED and not self.international_travel:
             return False
+
+        if not self.report_note or len(self.report_note) < 1:
```
tejfel: Changed and added some tests

ewheeler: thanks. looks good

EquiTrack/users/serializers.py
```diff
@@ -157,6 +157,7 @@ class UserCreationSerializer(serializers.ModelSerializer):
     groups = serializers.SerializerMethodField()
     user_permissions = serializers.SerializerMethodField()
     profile = UserProfileCreationSerializer()
+    t2f = T2FUserDataSerializer(source='*', read_only=True)
```
robertavram: this would make this extremely heavy... and other apps are using this that don;t need this information
as an alternative you could create a sparate serializer for t2f_list users and use that if a query param like "t2f=true" what do you think?

tejfel: It affects only the detail view, not the list.

PR #492, opened 2017-02-22
Feature - Partner organization overview API

n1207n:
This PR addresses an API endpoint for getting partner organization overview info including a list of its interventions serialized objects.


n1207n:
@ntrncic this is ready for review. Let me know if you have any questions! ;)

EquiTrack/partners/views/partner_organization_v2.py
```diff
@@ -210,3 +221,41 @@ def delete(self, request, *args, **kwargs):
...
+
+
+class PartnerOrganizationOverviewAPIView(APIView):
+
+    def get(self, request, partner_pk=None):
+        partner = get_object_or_404(PartnerOrganization, pk=partner_pk)
+        hact_min_reqs = partner.hact_min_requirements
+
+        if isinstance(partner.hact_values, str):
+            partner.hact_values = json.loads(partner.hact_values)
```
ewheeler: is it worth catching `JSONDecodeError`? is there any chance we might get malformed json here?

if so, not sure whether it would be more appropriate to 
1) return a 500 with a nice message so front-end can show an error to the user
2) return a 200 with `0`s for `prog_visit_done`, `spot_checks_done`,  `audit_done`, `audit_mr` and `planned_cash_transfer` along with remaining values that aren't from `partner.hact_values`

i'm guessing number 1? in either case we should log the error explicitly with a `logger.error` message

EquiTrack/partners/views/partner_organization_v2.py
```diff
@@ -210,3 +221,41 @@ def delete(self, request, *args, **kwargs):
...
+            interventions = InterventionListSerializer(interventions, many=True)
+
+        else:
+            interventions = GovernmentIntervention.objects.filter(partner=partner)
+
+            interventions = GovernmentInterventionListSerializer(interventions, many=True)
+
+        return Response({
+            'partner': partner_data,
+            'interventions': interventions.data,
```
ewheeler: what do `InterventionListSerializer` and `GovernmentInterventionListSerializer` return if called with an `EmptyQuerySet` from the prior `.objects.filter` call?

will `interventions.data` definitely exist?

EquiTrack/partners/views/partner_organization_v2.py
```diff
@@ -210,3 +221,41 @@ def delete(self, request, *args, **kwargs):
...
+            'spot_checks_mr': hact_min_reqs.get('spot_checks', 0),
+            'audit_done': partner.hact_values.get('audit_done', 0),
+            'audit_mr': partner.hact_values.get('audit_mr', 0),
+            'total_ct_cp': partner.total_ct_cp,
+            'total_ct_cy': partner.total_ct_cy,
+            'planned_cash_transfer': partner.hact_values.get('planned_cash_transfer', 0)
+        }
+
+        if partner.partner_type != PartnerType.GOVERNMENT:
+            interventions = Intervention.objects.filter(agreement__partner=partner)
```
ntrncic: may have to filter interventions by status or date. waiting for business feedback

EquiTrack/partners/views/partner_organization_v2.py
robertavram: hact values should be there always in a proper json format (jsonField with defaults on the model).. it gets automatically updated whenever we sync a partner... 
on another topic: @ntrncic  since you updated the functions this reminds me .. can you check the sync functions are correct on partner org sync

n1207n: Thanks for your inputs here ;)

First of all, I believe PartnerOrganization object has `hact_values` as JSON-encoded format when we are querying from database. Therefore, JSON decoding is needed here. 

Whenever we update PartnerOrganization object, JSON decoding/encoding is happening automatically as we overrode its save method. In this API endpoint, we don't modify anything for PartnerOrganization object. Therefore, JSON handling doesn't happen automatically.

For malformed data concern, that should have been catched when we call save method on PartnerOrganization object. The fact that we're querying from database assumes that all PartnerOrganization objects have a valid JSON encoded string for `hact_values` field.

Overall, I do not think this is an issue to think about JSON exceptions.

EquiTrack/partners/views/partner_organization_v2.py
n1207n: `.data` attribute comes from REST framework serializer class, and by default it will output an empty list as the serializer has been instantiated with `many=True`

EquiTrack/partners/views/partner_organization_v2.py
n1207n: Nice to know here. It would be matter of adding GET query parameters to add filtering support ;)

ntrncic: filter interventions by status = active

n1207n: @ntrncic I added a GET parameter support for intervention status filtering.

EquiTrack/partners/views/partner_organization_v2.py
ewheeler: @n1207n ok cool. thanks for the explanation!

EquiTrack/partners/views/partner_organization_v2.py
ewheeler: @n1207n makes sense. thanks

EquiTrack/partners/views/partner_organization_v2.py
```diff
@@ -210,3 +221,44 @@ def delete(self, request, *args, **kwargs):
...
+            interventions = InterventionListSerializer(interventions, many=True)
+
+        else:
+            interventions = GovernmentIntervention.objects.filter(partner=partner)
+
+            interventions = GovernmentInterventionListSerializer(interventions, many=True)
+
+        return Response({
+            'partner': partner_data,
+            'interventions': interventions.data,
```
robertavram: @n1207n I'm just realizing this... this shouldn't be a different endpoint, can we just add the 'interventions' key read-only to the PartnerDetailSerializer and PartnerCreateUpdateSerializer. This way we wouldn't need a separate endpoint for the overview tab. (partner_data is already serialized for the partner)

ntrncic: @robertavram I knew you would say that. We have to make another serializer for interventions to have only the fields needed for partner overview. Then add to existing partners endpoint.

EquiTrack/partners/views/partner_organization_v2.py
ntrncic: @robertavram task functions for hact are working fine.

n1207n:
@robertavram requested changes are made:

- InterventionSummaryListSerializer & GovernmentInterventionSummaryListSerializer

- PartnerOrganizationDetailSerializer now has partnership_overview method

Let me know if there are any other requests!

ntrncic:
made a simple rename from partnership_overview to interventions. otherwise looks good!

PR #495, opened 2017-02-23
T2f fixes pt5

tejfel:
A lot of file movement happened, that's why there are a lot of changed files. (Views and serializers were split up into smaller files)

This PR is on the top of #488


robertavram:
@tejfel can you comment in exactly what fixes are being addressed by each PR... it would increase visibility by the Product Team

EquiTrack/t2f/serializers/invoices.py
```diff
@@ -0,0 +1,49 @@
+from __future__ import unicode_literals
+
+from decimal import Decimal, getcontext, InvalidOperation
+
+from rest_framework import serializers
+
+from t2f.models import Invoice, InvoiceItem
+
+
+def rount_to_currency_precision(currency, amount):
```
robertavram: rount?

EquiTrack/t2f/management/commands/dsa_test_setup.py
```diff
@@ -0,0 +1,82 @@
+from __future__ import unicode_literals
```
robertavram: why do we need this as a management command?

EquiTrack/t2f/serializers/invoices.py
tejfel: Typo

EquiTrack/t2f/management/commands/dsa_test_setup.py
tejfel: It's added by my pycharm automatically. I prefer this because then I use unicode everywhere instead of bytestring.

PR #509, opened 2017-02-28
T2f fixes pt6

tejfel:
This is blocked by the pr #495 

!!! CAUTION !!!
Migrations cannot be applied at once, only in 2 steps.
First the t2f app has to be migrated and then the publics (and anything else).

`python manage.py migrate_schemas t2f`
`python manage.py migrate_schemas`


lordfuoco:
After this is merged please merge this to align the frontend

https://github.com/unicef/etools-t2f/pull/91

robertavram:
@tejfel can we not make dependencies on migrations so they run in proper order

robertavram:
also @tejfel do you mind describing the fixes on the PR? the product team could benefit greatly from it

EquiTrack/publics/tasks.py
```diff
@@ -99,47 +99,97 @@ def import_exchange_rates(xml_structure):
...
+   wbs_to_create = wbs_code_set - existing_wbs_codes
+   bulk_wbs_list = []
+   for wbs_code in wbs_to_create:
+       business_area_code = wbs_code[:4]
+       wbs = WBS(name=wbs_code, business_area=business_area_cache[business_area_code])
+       bulk_wbs_list.append(wbs)
+
+   WBS.objects.bulk_create(bulk_wbs_list)
+
+   wbs_mapping = {wbs.name: wbs for wbs in WBS.objects.filter(name__in=wbs_code_set)}
```
robertavram: since most of the time we'll have nothing or very little to create, we can just use existing `existing_wbs_codes` and if to_create just update the wbs_mapping with objects from another query base on all the names of the objects that were just created. 

these are thousands of records.. we need to be extra careful with the resources on the db

EquiTrack/publics/tasks.py
```diff
@@ -99,47 +99,97 @@ def import_exchange_rates(xml_structure):
...
+
+   grant_to_create = grant_code_set - existing_grants
+   bulk_grant_list = []
+   for grant_code in grant_to_create:
+       grant = Grant(name=grant_code)
+       bulk_grant_list.append(grant)
+
+   Grant.objects.bulk_create(bulk_grant_list)
+   
+   grant_mapping = {g.name: g for g in Grant.objects.filter(name__in=grant_code_set)}
```
robertavram: the previous comment is valid for grants as well

EquiTrack/publics/tasks.py
```diff
@@ -99,47 +99,97 @@ def import_exchange_rates(xml_structure):
...
+
+   fund_to_create = fund_code_set - existing_funds
+   bulk_fund_list = []
+   for fund_code in fund_to_create:
+       fund = Fund(name=fund_code)
+       bulk_fund_list.append(fund)
+
+   Fund.objects.bulk_create(bulk_fund_list)
+   
+   fund_mapping = {f.name: f for f in Fund.objects.filter(name__in=fund_code_set)}
```
robertavram: same comment as above applies for fund_mapping

EquiTrack/publics/tasks.py
```diff
@@ -99,47 +99,97 @@ def import_exchange_rates(xml_structure):
...
+def _fetch_business_areas(wbs_set):
+   business_area_codes = {wbs_code[:4] for wbs_code in wbs_set}
+   business_area_qs = BusinessArea.objects.filter(code__in=business_area_codes)
+   return {ba.code: ba for ba in business_area_qs}
+
+
+def create_wbs_objects(wbs_code_set):
+   business_area_cache = _fetch_business_areas(wbs_code_set)
+
+   existing_wbs_objects = WBS.objects.filter(name__in=wbs_code_set).select_related('business_area')
```
robertavram: why are we using select_related here?

EquiTrack/publics/tasks.py
```diff
@@ -99,47 +99,97 @@ def import_exchange_rates(xml_structure):
...
+    fund_mapping = create_fund_objects(fund_code_set)
+
+    wbs_grant_mapping = defaultdict(list)
+    grant_fund_mapping = defaultdict(list)
+    for g in groups:
+        wbs = wbs_mapping[g['wbs_code']]
+        grant = grant_mapping[g['grant_code']]
+        fund = fund_mapping[g['fund_code']]
+
+        wbs_grant_mapping[wbs].append(grant)
```
robertavram: we should not do this unless the record has changed or does not exist... if we don;t use our current objects and validate against new objects coming before save... this will be extremely expensive on the db

EquiTrack/publics/tasks.py
```diff
@@ -99,47 +99,97 @@ def import_exchange_rates(xml_structure):
...
+        grant_fund_mapping[grant].append(fund)
+
+    for wbs, grants in wbs_grant_mapping.items():
+        wbs.grants.set(grants)
+
+    for grant, funds in grant_fund_mapping.items():
+        grant.funds.set(funds)
```
robertavram: previous comment valid here as well.. please revisit

tejfel:
The reason why it has dependencies to run in a proper order. If the wbs/grant/fund related models are not cleared before the wbs/grant/fund entries are deleted, the db will throw an integrity error.
In general what's the problem with dependencies?

EquiTrack/publics/tasks.py
tejfel: Some remaining code from previous version. Removed.

EquiTrack/publics/tasks.py
tejfel: This is a dict mapping. Nothing is happening with the db here.

EquiTrack/publics/tasks.py
tejfel: Django does the check under the hood. It adds only the new ones and removes the not existing ones. Everything else will be untouched.

EquiTrack/publics/tasks.py
tejfel: I don't get how you want. How can I make the mapping if I don't have the objects?

And on the other hand this will run per workspace so in total there will be not so much wbs at once. And in my opinion even 1-2k models in memory is not so bad.

EquiTrack/publics/tasks.py
robertavram: sorry this was meant for the .set

EquiTrack/publics/tasks.py
robertavram: so .set won;t call the db if no changes?

EquiTrack/publics/tasks.py
robertavram: You already have the objects, you can do the mapping on those.
 existing_wbs_objects = WBS.objects.filter(name__in=wbs_code_set)
(line 111) 
after you did that mapping.. if wbs_to_create you can do a much smaller db call and filter by the wbs_to_create
and add that result to your mapping
I might be missing something.. does this make sense?

EquiTrack/publics/tasks.py
tejfel: It will because it has to check which ones are already in. Otherwise it would not know what to do. Otherwise I would have to check which to create, leave there or remove. Does it make so big overhead which would require a lot of custom code?

robertavram: I'm pretty sure that you'll have at least a db query here... under the hood I suspect it will at least get get the existing funds on the grant. somehow I wonder if we could prefetch_related for funds and grants and manually check before calling the db.

robertavram: just saw that you answered before I posted my response 
(i removed my response), I just think that for 70k records doing this for every record is not a good thing. what do you think? especially because we're going over all wbs, then all grants

tejfel: After sync and unique check I got these numbers (over the whole system):

```
>>> WBS.objects.all().count()
6539
>>> Grant.objects.all().count()
8287
>>> Fund.objects.all().count()
28
```

So in total max 15000 elements, but here we will fetch only per workspace, so probably will be around 1500 WBS, 2000 grant and max 28 funds.

Issue #504, opened 2017-02-27
url to item in export serializers


ntrncic:

robertavram:
https://github.com/unicef/etools-issues/issues/5

robertavram:
closes:https://github.com/unicef/etools-issues/issues/5

robertavram:
fixes:https://github.com/unicef/etools-issues/issues/5

robertavram:
closes https://github.com/unicef/etools-issues/issues/5

robertavram:
closes robertavram/etools-issues#5

PR #513, opened 2017-02-28
Add URL field on export

ilcsik:


ilcsik:
Resolves https://github.com/unicef/etools/issues/504

PR #530, opened 2017-03-04
T2f fixes pt8

tejfel:


robertavram:
@tejfel can you split the other business logic from the synchronizer logic.. I need to look deeper in the synchronizer... but it don;t want to hold off the other code... they don;t seem to be dependent on eachother.

PR #541, opened 2017-03-06
Fix trips dashboard no section error

ilcsik:
Frontend needs to be aligned.

For travels_by_section that have no Section associated, section_id is null and section_name is "no_section_selected". Frontend should say "No Section Selected" instead of Section name.


ilcsik:
section_name is "No Section Selected" now, so section_id can be checked for null and section_name can be used right away as row name.

PR #553, opened 2017-03-10
Fix travel reference number generator

tejfel:


EquiTrack/users/models.py
```diff
@@ -71,6 +71,43 @@ def __unicode__(self):
...
+
+        setattr(counter_model, counter_type, counter_value + 1)
+        counter_model.save()
+
+        return counter_value
+
+    @classmethod
+    def create_counter_model(cls, sender, instance, created, **kwargs):
+        """
+        Signal handler to create user profiles automatically
```
robertavram: Signal handler to create  workspace counter models automatically

PR #558, opened 2017-03-13
T2f fixes pt10

tejfel:


robertavram:
@tejfel Please always do 1PR / Fix / Feature... this could be split into 2. I'm merging this, but for the future there should never be more than 1 Fix / PR

PR #579, opened 2017-03-24
Fix mailing serializer

simon-the-sloth:
fixes unicef/etools-issues#96
fixes unicef/etools-issues#98
fixes unicef/etools-issues#100


robertavram:
@simon-the-sloth can you look at the tests.. NonExistingTime Error is happening randomly throughout all PRs... it seems to be connected with the trips tests... Can you look at it?

PR #581, opened 2017-03-24
update iteneraryitem model to enforce ordering

ewheeler:
- migration to correct improperly ordered items
- update serializer to enforce ordering

also these misspellings are driving me insane!

fixes unicef/etools-issues#148


ntrncic:
the ordered migration is so cool. learned something new today!

PR #591, opened 2017-03-30
Add validation fixes

ilcsik:
Resolves https://github.com/unicef/etools-issues/issues/204 and https://github.com/unicef/etools-issues/issues/180


ntrncic:
@ilcsik please resolve conflicts

EquiTrack/partners/tests/test_views.py
```diff
@@ -231,6 +231,25 @@ def test_api_partners_update_with_members(self):
         self.assertEquals(response.status_code, status.HTTP_200_OK)
         self.assertEquals(len(response.data["staff_members"]), 2)
 
+    def test_api_partners_update_assessments_invalid(self):
```
ewheeler: need to test for valid dates (today, yesterday, long ago) and for `completed_date` equal to tomorrow

EquiTrack/partners/validation/agreements.py
```diff
@@ -58,6 +58,19 @@ def amendments_ok(agreement):
     # To be Continued
     return True
 
+def amendments_signed_amendment_valid(agreement):
```
ewheeler: Django's FileField is implemented as CharField with empty strings or nulls when a file isn't present.

Something like this should be faster and easier on the db
`return all(agreement.amendments.values_list('signed_amendment', flat=True))`

ilcsik: Done in https://github.com/unicef/etools/pull/591/commits/dd3d9f2b06f245fed961dd4e0d85c33d57e93f0b

EquiTrack/partners/tests/test_views.py
ilcsik: Done in https://github.com/unicef/etools/pull/591/commits/dd3d9f2b06f245fed961dd4e0d85c33d57e93f0b

EquiTrack/partners/serializers/partner_organization_v2.py
```diff
@@ -179,6 +180,12 @@ class Meta:
         model = Assessment
         fields = "__all__"
 
+    def validate(self, data):
+        today = datetime.date.today()
```
ilcsik: Done in https://github.com/unicef/etools/pull/591/commits/dd3d9f2b06f245fed961dd4e0d85c33d57e93f0b

EquiTrack/partners/validation/agreements.py
robertavram: For signed amendment if we're looking to satisfy the required field.. that should be on the serializer or the model. However there seems to be a business issue here.. originally the request was to not add the signature or the file as required fields because we wanted to be able to generate the document for signing.. 
I will check with the BO and get back to you

EquiTrack/partners/validation/agreements.py
```diff
@@ -58,6 +58,16 @@ def amendments_ok(agreement):
...
+    return all(agreement.amendments.values_list('signed_amendment', flat=True))
+
+def amendments_signed_date_valid(agreement):
+    today = date.today()
+    for amendment in agreement.amendments.all():
```
robertavram: two things here:
1. since you're already going through the "signed_date_valid" you could easily add the check if the file is present as well rather than the "amendments_signed_amendment_valid" function.
2. I believe that we discovered that .all() here has been previously cached.. so.. all new records added might not be present in .all() call... to avoid that.. you can just use .filter()

ilcsik: 1, I believe that the main reason I did it separately was to be able to return separate error messages for the different fields, which was the goal of the last refactor.
2, Yeah true, replaced in https://github.com/unicef/etools/pull/591/commits/17feb269bbd81a09ec3e39756d4c24d86cbfea4c

EquiTrack/partners/serializers/partner_organization_v2.py
```diff
@@ -179,6 +181,12 @@ class Meta:
         model = Assessment
         fields = "__all__"
 
+    def validate(self, data):
+        today = timezone.now().date()
```
robertavram: why timezone.now? shouldn;t we use datetime?

ilcsik: In this comment Evan advised that it should be UTC

https://github.com/unicef/etools/pull/591#discussion_r109687342

PR #596, opened 2017-04-04
Fix cost assignments

simon-the-sloth:


EquiTrack/t2f/helpers/cost_summary_calculator.py
```diff
@@ -3,61 +3,88 @@
...
         travel_agent_expenses = sorted(travel_agent_expenses, key=lambda o: o.vendor_number)
         expenses.extend(travel_agent_expenses)
 
         # explicit None checking should happen here otherwise the 0 will be filtered out
-        if parking_money is not None:
-            expenses.append(self.ExpenseDTO('', parking_money))
+        for parking_money_expense in parking_money:
+            expenses.append(self.ExpenseDTO('', parking_money_expense))
+
+        # TODO fix this
```
ewheeler: what needs fixing here?

EquiTrack/t2f/tests/test_exports.py
```diff
@@ -115,7 +115,7 @@ def test_finance_export(self):
                           'No',
                           'Yes',
                           '0.00',
-                          '500.00',
+                          '500 USD',
```
ewheeler: is this correct? shouldn't decimal places be included like `500.00 USD`?

additionally, it would likely be more useful to have amount (`500.00`) and currency (`USD`) in separate columns in the export. is there a spec for the export requirements?

EquiTrack/t2f/models.py
```diff
@@ -328,7 +329,8 @@ def plan(self):
     @threshold_decorator
     @transition(status, source=[APPROVED, SENT_FOR_PAYMENT, CERTIFIED], target=SENT_FOR_PAYMENT)
     def send_for_payment(self):
-        self.preserved_expenses = self.cost_summary['expenses_total']
+
+        self.preserved_expenses_local = self.cost_summary['expenses_total'][0]['amount']
```
ewheeler: i don't see a place where presence of expenses is enforced.

if there are no expenses, then there will be no item `[0]` and no `[amount]` and this will raise an exception

EquiTrack/t2f/helpers/cost_summary_calculator.py
simon-the-sloth: Nothing. Was a comment for myself during development.

EquiTrack/t2f/tests/test_exports.py
simon-the-sloth: The decimal places are controlled by currency (`decimal_places` attribute). This is formatted based on that (the test factory sets `decimal_places` to 0.

EquiTrack/t2f/models.py
simon-the-sloth: The setting of the value happens before in during the approval process and this state cannot be reached without that.

EquiTrack/t2f/helpers/cost_summary_calculator.py
```diff
@@ -3,61 +3,87 @@
...
 
-        total_expense = sum(expense_mapping.values(), Decimal(0))
-        total_expense = total_expense.quantize(Decimal('1.0000'))
+        local_expenses = [e.amount for e in chain.from_iterable(expense_mapping.values())
+                          if e.document_currency == self.travel.currency]
+        total_expense_local = sum(local_expenses, Decimal(0))
+        total_expense_local = total_expense_local.quantize(Decimal('1.0000'))
+
+        usd_expenses = [e.amount for e in chain.from_iterable(expense_mapping.values())
+                        if e.document_currency != self.travel.currency]
```
ntrncic: Is this summing USD currency if the two currencies don't match? Could you clarify what the intended outcome is?

EquiTrack/t2f/helpers/cost_summary_calculator.py
```diff
@@ -3,61 +3,87 @@
...
+        total_expense_usd = total_expense_usd.quantize(Decimal('1.0000'))
 
         # Order the expenses
         expenses = []
         if 'user' in expense_mapping:
-            expenses.append(self.ExpenseDTO('user', expense_mapping.pop('user')))
-        parking_money = expense_mapping.pop('', None)
+            for expense in expense_mapping.pop('user'):
+                expenses.append(self.ExpenseDTO('user', expense))
+        parking_money = expense_mapping.pop('', [])
```
robertavram: what is parking money? and when are we using an empty string as a key. Are you trying to just assign an empty list to `parking_money`?

simon-the-sloth: Parking money is a placeholder. When someone is traveling but does not know the exact cost, can make a parking money entry with an approximate amount and this can be adjusted later.
The parking money has empty vendor code (since it does not belong to any) and that's the empty key (mapping is `vendor_code`: `aggregated value`)
I just use the default value for pop. In case no parking money was defined by user, I'll get an empty list.

robertavram: Is it ok to use empty string as keys in a dictionary?

PR #604, opened 2017-04-06
Add threshold check for more state changes

simon-the-sloth:
fixes unicef/etools-issues#193


EquiTrack/t2f/tests/test_treshold.py
```diff
@@ -194,3 +194,28 @@ def test_multi_step_reach(self):
         travel = Travel.objects.get(id=travel_id)
         self.assertEqual(travel.approved_cost_traveler, 0)
         self.assertEqual(travel.approved_cost_travel_agencies, 120)
+
+    def test_threshold_check_on_complete(self):
```
ewheeler: should have tests for both cases where `settings.DISABLE_INVOICING = True` and `settings.DISABLE_INVOICING = False` as well as cases for `self.check_threshold()` returning `True` and case for `False`

simon-the-sloth: Tests added.

EquiTrack/t2f/models.py
```diff
@@ -110,7 +110,7 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
-def threshold_decorator(func):
+def send_for_paymnet_threshold_decorator(func):
```
robertavram: send for "payment"

PR #623, opened 2017-04-20
Change wording

simon-the-sloth:
fixes unicef/etools-issues#108


EquiTrack/t2f/models.py
```diff
@@ -13,7 +13,7 @@
 from django.db import models, connection
 from django.template.context import Context
 from django.template.loader import render_to_string
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import ugettext, ugettext_lazy as _
```
robertavram: as throughout the app _ is used for unused variables, until we come up with the coding guide, let's keep full text (I know that the convention for ugettext is to import it as _)

PR #629, opened 2017-04-24
Dsa region values

simon-the-sloth:
Does the backend part of unicef/etools-issues#218 and unicef/etools-issues#58


ntrncic:
could you reference a card or an issue that this PR reflects on

EquiTrack/publics/models.py
```diff
@@ -209,3 +212,55 @@ def unique_name(self):
 
     def __unicode__(self):
         return self.label
+
+    @memoize(maxsize=50)
```
robertavram: let's take this out for now.. not sure this is worth the complication... we can revisit later.

EquiTrack/publics/serializers.py
```diff
@@ -1,29 +1,53 @@
...
     long_name = serializers.CharField(source='label')
 
     class Meta:
         model = DSARegion
-        fields = ('id', 'country', 'area_name', 'area_code', 'dsa_amount_usd', 'dsa_amount_60plus_usd',
-                  'dsa_amount_local', 'dsa_amount_60plus_local', 'room_rate', 'finalization_date', 'eff_date',
-                  'unique_id', 'unique_name', 'label', 'long_name')
+        fields = ('id', 'country', 'area_name', 'area_code', 'unique_id', 'unique_name', 'label', 'long_name')
+
+    def to_representation(self, instance):
```
robertavram: It seems like we're using a ModelSerializer to serialize data from 2 models... Can we define a separate Serializer for this purpose as it's a bit counterintuitive.

EquiTrack/publics/models.py
simon-the-sloth: Removed already.

EquiTrack/publics/models.py
```diff
@@ -209,3 +212,55 @@ def unique_name(self):
...
+                    'room_rate', 'finalization_date', 'effective_from_date']:
+            return getattr(self.get_rate_at(), item)
+        return super(DSARegion, self).__getattr__(item)
+
+
+class DSARate(models.Model):
+    DEFAULT_VALID_TO = date(2999, 12, 31)
+
+    region = models.ForeignKey('DSARegion', related_name='rates')
+    effective_from_date = models.DateField(auto_now_add=True)
```
simon-the-sloth: This is removed too.

EquiTrack/publics/serializers.py
robertavram: Based on our phone conversation, I'll merge this in, but let's see if we can figure out a better solution. In the meantime a docstring explaining why this approach was taken works. @simon-the-sloth

PR #633, opened 2017-04-26
Alter origin and destination fields to FK, migrate and backup data

ilcsik:
Connects https://github.com/unicef/etools-issues/issues/216


ilcsik:
This PR contains a refactor for ItineraryItem origin and destination fields, to make them foreign key instead of charfield.

It also contains a migration sequence which tries to match existing origin and destination fields and replace them with the relevant Location ids. If the name is not found in the Location table, the field remains empty. The old values are still accessible in origin_old and destination_old fields.

I have ran a migration on the 04.21. staging dump locally and 46k locations were resolved and migrated to FK, but 65k locations were not matched any Location. 

One solution we suggest is that whenever an FK is missing, we display a readonly field just to show what's the original data, and let the user/admin fill the dropdown with the actual location. Once it's filled, the readonly field disappears.

I'm not sure if this solution is feasible or is there something else we can do in the current situation. Just let me know.

robertavram:
@ilcsik let us get back to you.
@anusenan please make the business owners aware of the potential data issue in regards to moving from free-text fields for Itinerary locations to location selector.
I am against handling old data in a special way as it's complex to develop and presents the opportunity for extra burden to improve on / change these elements in the future.

robertavram:
closing until we have further clarity on the business requirements.

PR #635, opened 2017-04-28
Improve publics admin

ilcsik:
General mixin for include fields and search etc into publics admin. To be refined next round.


robertavram:
I like it.. please resolve conflicts first

PR #651, opened 2017-05-04
Fix travel admin status field

ilcsik:
Makes the field read only so travel can be saved.


ntrncic:
we want to be able to change the status in the admin.

simon-the-sloth:
@ntrncic Not possible right now since the state machine does not allow direct state modification. If you need this functionality, maybe can be achieved with some custom admin actions.

PR #652, opened 2017-05-04
Fix travel activity export

simon-the-sloth:
fixes unicef/etools-issues#76


EquiTrack/t2f/serializers/export.py
```diff
@@ -23,19 +23,40 @@ def to_representation(self, value):
...
     class Meta:
-        model = Travel
-        fields = ('id', 'reference_number', 'traveler', 'purpose', 'start_date', 'end_date', 'status', 'created',
-                  'section', 'office', 'supervisor', 'ta_required', 'ta_reference_number', 'approval_date', 'is_driver',
-                  'attachment_count')
+        fields = ('reference_number', 'traveler', 'office', 'section', 'status', 'trip_type', 'partner', 'partnership',
+                  'results', 'locations', 'when', 'is_secondary_traveler',	'primary_traveler_name')
+
+    def get_locations(self, obj):
+        return ', '.join([l.name for l in obj.activity.locations.all()])
```
ewheeler: better to use `.values_list('name', flat=True)` than list comprehension (see https://docs.djangoproject.com/en/1.11/ref/models/querysets/#values-list)

EquiTrack/t2f/serializers/export.py
```diff
@@ -23,19 +23,40 @@ def to_representation(self, value):
...
+
+    def get_is_secondary_traveler(self, obj):
+        if self._is_secondary_traveler(obj):
+            return 'YES'
+        return ''
+
+    def get_primary_traveler_name(self, obj):
```
ewheeler: why is this called `get_primary_traveler_name` if it only returns the primary traveler name if a secondary traveler is present? should always return primary or should have a more descriptive  name

EquiTrack/t2f/serializers/export.py
```diff
@@ -23,19 +23,40 @@ def to_representation(self, value):
...
+        if self._is_secondary_traveler(obj):
```
ntrncic: why are you checking for secondary traveler here?

EquiTrack/t2f/serializers/export.py
```diff
@@ -23,19 +23,40 @@ def to_representation(self, value):
...
 
 
-class TravelListExportSerializer(TravelListSerializer):
-    traveler = serializers.CharField(source='traveler.get_full_name')
-    section = serializers.CharField(source='section.name')
-    office = serializers.CharField(source='office.name')
-    ta_reference_number = serializers.CharField(source='reference_number')
-    approval_date = serializers.DateTimeField(source='approved_at')
-    attachment_count = serializers.IntegerField(source='attachments.count')
+class TravelActivityExportSerializer(serializers.Serializer):
```
ntrncic: We are removing the Travel Export functionality and fixing TravelActivity export. I don't think the waffle card reflects this.

EquiTrack/t2f/serializers/export.py
simon-the-sloth: The spec shows that we should only include the primary traveler name if it's different than the main traveler. https://docs.google.com/spreadsheets/d/1E4qOdZWxtPPCxf3WIH7e8AqKu8-A8hpQr6iDj-uakeo/edit#gid=144405922

EquiTrack/t2f/serializers/export.py
simon-the-sloth: This was called travel export back, but then the requirement changed and this is the proper name of it now.

EquiTrack/t2f/serializers/export.py
simon-the-sloth: It's because of rest framework naming convention. For `SerializerMethodField` I has to use `get_<field_name>` format and here the field called `primary_traveler_name`.

EquiTrack/t2f/serializers/export.py
simon-the-sloth: `.select_related()` is not working with values list, that's why list comprehension was used. It can be changed, but then the query count will be 5+N instead of 6.

EquiTrack/t2f/serializers/export.py
ntrncic: ok if it's according to specs

EquiTrack/t2f/serializers/export.py
ewheeler: ahh ok. makes sense

EquiTrack/t2f/serializers/export.py
ewheeler: gotcha. good call

EquiTrack/t2f/serializers/export.py
ntrncic: ok cool

PR #676, opened 2017-05-22
Fix expense currency integrity issue

simon-the-sloth:
fixes unicef/etools-issues#76
fixes unicef/etools-issues#330


simon-the-sloth:
Migrations might fail on some DB if it has corrupted entries!

simon-the-sloth:
@ntrncic All invoice records which will fail has a traveler vendor number so it's possible to fix easily. All what has to be done is:
- get the records
- assign the currency from the travel to it
- save them

```python
invoice = Invoice.objects.get(currency__isnull=True)
invoice.currency = invoice.travel.currency
invoice.save()
```

In total 3 tenants are affected: Lebanon (1 entry), Kenya (2 entries) and Kyrgyzstan (3 entries)

ntrncic:
@simon-the-sloth can't we write a data migration then

PR #678, opened 2017-05-25
Fix permission matrix check

simon-the-sloth:


EquiTrack/t2f/__init__.py
```diff
@@ -1 +1,23 @@
-from __future__ import unicode_literals
\ No newline at end of file
+from __future__ import unicode_literals
+
+from django.utils.translation import ugettext_lazy
+
+
+class UserTypes(object):
```
robertavram: Add a todo to revise when we have a guide on where to place choices

PR #683, opened 2017-05-26
Refactor DSA upload to use Azure

ilcsik:
Blind coding - to be tested


ilcsik:
Connects https://github.com/unicef/etools-issues/issues/58

ntrncic:
@ilcsik i removed the codecs line. the file is already in memory and open for reading.
could you refactor the task to not call the db on every record. Maybe grab all the existing DSA Rates in a list and then compare them to the input file.

ilcsik:
@ntrncic I wouldn't add that complexity to the code, get_or_create() is a 1liner, convenient and simple, comparing lists would be more verbose and error prone, just to save a couple of seconds on a task that run once in a few weeks. Let me know if you still want to refactor, though.

EquiTrack/publics/tasks.py
```diff
@@ -221,12 +223,12 @@ def __init__(self, dsa_rate_upload):
         self.warnings = {}
         self.dsa_rate_upload = dsa_rate_upload
 
-    def read_input_file(self, input_file_path):
-        # Utilize codecs module to remove BOM if present
-        with codecs.open(input_file_path, 'r', encoding='utf-8-sig') as input_file:
-            utf_8_file = codecs.iterencode(input_file, encoding='utf-8')
+    def read_input_file(self, filename):
+        storage = AzureStorage()
+        with storage.open("{}".format(filename)) as input_file:
```
simon-the-sloth: Instead of format just simply filename can be passed.

ilcsik: Done in https://github.com/unicef/etools/pull/683/commits/6b6726c950fba9652a0ba6f4cbd2de200e96cf72

PR #684, opened 2017-05-30
Users api merge

ntrncic:
connects unicef/etools-issues#349

New Users api version 3


EquiTrack/EquiTrack/urls.py
```diff
@@ -35,20 +29,13 @@
     LocationsLightViewSet,
 )
 from trips.views import TripsViewSet, TripFileViewSet, TripActionPointViewSet
-
 from partners.views.v1 import (
-    PartnerOrganizationsViewSet,
```
robertavram: this should be part of a different PR... removing stale viewsets

EquiTrack/users/urls.py
```diff
@@ -1,29 +1,37 @@
...
+    'get': 'retrieve',
+})
 
 urlpatterns = patterns(
     '',
 
     # api
     url(r'^api/profile/$', UserAuthAPIView.as_view()),
     url(r'^api/changecountry/$', ChangeUserCountryView.as_view(http_method_names=['post'])),
+    url(r'^api/(?P<pk>[0-9]+)/$', UsersDetailAPIView.as_view(http_method_names=['get'])),
```
robertavram: the new direction we're going is:
/api/v[n]/users/etc

EquiTrack/users/urls.py
```diff
@@ -1,29 +1,37 @@
...
+    url(r'^api/viewset/(?P<pk>\d+)/$', user_detail, name='user-detail'),
+    url(r'^api/viewset/', user_list, name='user-list'),
```
robertavram: we should definitely not include "viewset" in the url

EquiTrack/users/urls.py
```diff
@@ -1,29 +1,37 @@
...
     url(r'^api/', UsersView.as_view()),  # TODO: staff required , partners should not be able to hit this
-    url(r'^api/(?P<pk>\d+)/$', UsersDetailAPIView.as_view(http_method_names=['get'])),
     url(r'^myprofile/$', MyProfileAPIView.as_view(), name="myprofile-detail"),
-
-    # url(r'^api/(?P<pk>\d+)/$', UsersDetailAPIView.as_view()),
-
+    url(r'^country/$', CountryView.as_view(http_method_names=['get']), name="country-detail"),
```
robertavram: is this a new api ep?

EquiTrack/EquiTrack/urls.py
ntrncic: @robertavram Please explain why? I want to get an understanding.

EquiTrack/users/urls.py
ntrncic: this was from url_v2. i moved it to v1 urls

EquiTrack/users/tests/test_views.py
```diff
@@ -109,7 +99,7 @@ def test_api_groups_list(self):
     def test_api_users_retrieve_myprofile(self):
         response = self.forced_auth_req(
             'get',
-            '/api/v2/users/myprofile/',
+            '/api/v3/users/profile/',
```
robertavram: are the v2 eps still valid?
if so.. leave the tests for v2 as well... (copy them for v3) we should only remove v2 tests when we deprecate the endpoints...

ntrncic: @robertavram yeah i thought about that too. tried to get one past you... see if you're still in vacation mode

PR #689, opened 2017-06-01
[BREAKING] Change trip activity validation

ilcsik:
Connects https://github.com/unicef/etools-issues/issues/391


EquiTrack/t2f/serializers/travel.py
```diff
@@ -194,6 +195,14 @@ def validate(self, attrs):
         elif 'government_partnership' in attrs:
             attrs['partnership'] = None
 
+        if attrs.get('partner', None) and \
+                attrs['partner'].partner_type == PartnerType.GOVERNMENT and \
+                'travel_type' in attrs and \
+                (attrs['travel_type'] == TravelType.PROGRAMME_MONITORING or
+                attrs['travel_type'] == TravelType.SPOT_CHECK) and \
+                'result' not in attrs:
+            raise ValidationError({'result': serializers.Field.default_error_messages['required']})
```
simon-the-sloth: In case it's a partial update, one key might be in the attrs but the other not. Still possible that this combination occurs (example: partner was set in a previous call and now only the travel_type gets updated).

```python
partner = attrs.get('partner', getattr(self.instance, 'partner'))
travel_type = attrs.get('travel_type', getattr(self.instance, 'travel_type'))

if partner and travel_type and partner.partner_type == PartnerType.GOVERNMENT and travel_type in [...]:
    raise ValidationError(...)
```

ilcsik: Done in https://github.com/unicef/etools/pull/689/commits/418d811efb322dd09a534ccf3bd1c31e4a9dd62e

PR #690, opened 2017-06-01
Fix search bar exact match error

ilcsik:
Connects https://github.com/unicef/etools-issues/issues/152


EquiTrack/t2f/filters/__init__.py
```diff
@@ -34,7 +34,7 @@ def filter_queryset(self, request, queryset, view):
         if search_str:
             q = Q()
             for field_name in self._search_fields:
-                constructed_field_name = '{}__iexact'.format(field_name)
+                constructed_field_name = '{}__icontains'.format(field_name)
```
ntrncic: have you tested the performance of this call?

ilcsik: @ntrncic I tested the API on staging DB and it feels instant so I havent measured. Whenever we need to make it more faster we can put trigram indexes on the fields like we did in the case of locations.

ntrncic: @ilcsik ok cool we do it this way for now.

PR #696, opened 2017-06-07
Fix finance export

simon-the-sloth:
fixes unicef/etools-issues#76

### DIFF
--- a/EquiTrack/t2f/models.py
+++ b/EquiTrack/t2f/models.py
@@ -275,7 +275,7 @@ class Travel(models.Model):
         if self.ta_required and self.itinerary.all().count() < 2:
             raise TransitionError(ugettext('Travel must have at least two itinerary item'))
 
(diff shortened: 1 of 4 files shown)


EquiTrack/t2f/models.py
```diff
@@ -271,13 +271,17 @@ def check_travel_count(self):
 
         return True
 
-    def check_itinerary_count(self):
+    def validate_itinerary(self):
         if self.ta_required and self.itinerary.all().count() < 2:
             raise TransitionError(ugettext('Travel must have at least two itinerary item'))
+
+        if self.itinerary.filter(dsa_region=None).exists():
```
ntrncic: but what if it's ta not required, does it still need a dsa region?

simon-the-sloth: No, then can be optional. Updated the code.

PR #711, opened 2017-06-14
Generic attachments models

roman-razortheory:
Generic models for attachments.  

Attachments types can be assigned dynamically from database using FileType model. 
Attachments are support both files and hyperlinks.


EquiTrack/attachments/models.py
```diff
@@ -0,0 +1,50 @@
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ValidationError
+from django.db import models
+from django.utils import six
+from django.utils.encoding import python_2_unicode_compatible
+from django.utils.translation import ugettext as _
+
+from model_utils.models import TimeStampedModel
+from ordered_model.models import OrderedModel
```
robertavram: what's orderedModel used for? is there any way we can do without?

EquiTrack/attachments/models.py
```diff
@@ -0,0 +1,50 @@
...
+@python_2_unicode_compatible
+class Attachment(TimeStampedModel, models.Model):
+    file_type = models.ForeignKey(FileType)
+
+    file = models.FileField(upload_to='files', blank=True, null=True)
+    hyperlink = models.CharField(max_length=255, blank=True, null=True)
+
+    content_type = models.ForeignKey(ContentType)
+    object_id = models.IntegerField()
+    object = GenericForeignKey()
```
robertavram: can we name this content_object? to not confuse with python object

EquiTrack/attachments/models.py
th13f: we need ordered model here to order allowed file types in choices

PR #712, opened 2017-06-14
Audit organizations models

roman-razortheory:
Audit organization model with organization staff members.

Base organization functionality moved to abstract classes in separate app for future usage in TPM.


EquiTrack/EquiTrack/settings/base.py
```diff
@@ -428,6 +428,9 @@
     't2f',
     'workplan',
     'actstream',
+    'audit',
+
+    'utils.organizations',
```
robertavram: let's not nest Organizations under utils... raise it one level so that all apps are at the same level.
rename app from Organizations to "Firms"

pitsevich:
@robertavram We can do a review of the whole thing together - might indeed be easier this way. 

We'll need a few more days to sort out/polish some internal to-dos in remaining places of the code, but I think we can schedule this at some point next week (maybe after Wed).

I'll send an email to do the scheduling of a review session!

PR #715, opened 2017-06-14
Add guidelines document

vkurup:
connects unicef/etools-issues#448

This is super barebones. I mostly copied a portion of a similar doc we have in a different project. We'll be updating this as we go along, and suggestions for content are welcome. It may be useful to split this into multiple docs, depending on how much we cover.

I've added Sphinx so that we can make it easy to push to readthedocs.io, if that's wanted.


EquiTrack/requirements/local.txt
```diff
@@ -1,2 +1,12 @@
 -r base.txt
 django-extensions
+
```
ntrncic: do we need all these packages?

vkurup: All of the indented packages (starting at alabaster) are requirements of sphinx. So doing `pip install sphinx` automatically installs alabaster, etc. I think it's a good practice to pin every package, so that everyone's environment is as identical as possible.

So, the short answer is that if we want Sphinx, then yes, we need all these packages.

Which leads to the question if we want Sphinx. Sphinx makes it easy to publish docs to different formats (HTML, PDF, etc), and specifically to publish to readthedocs.io. I think that's a very useful feature, but it's also something we could add later, so I'd be fine removing it for now until we decided to go down that path.

ntrncic: @vkurup we should use Sphinx. didn't know the others are requirements.

PR #718, opened 2017-06-14
Fix agreement reference number - cp to agreement

robertavram:
Fixes unicef/etools-issues#452


EquiTrack/partners/serializers/agreements_v2.py
```diff
@@ -149,6 +150,7 @@ class AgreementCreateUpdateSerializer(serializers.ModelSerializer):
     partner_name = serializers.CharField(source='partner.name', read_only=True)
 
     amendments = AgreementAmendmentCreateUpdateSerializer(many=True, read_only=True)
+    country_programme = serializers.PrimaryKeyRelatedField(queryset=CountryProgramme.objects.all(), required=True)
```
ntrncic: why is this needed on amendment type?

robertavram: It's not on type, it's on the agreement

robertavram: Although, if the type is MOU, I believe this is not required. I'll have to double check with the spec.

ntrncic: according to specs it says show country programme only for PCA.

robertavram: ok.. I'll fix it and add some tests

PR #721, opened 2017-06-15
Add coverage config locally and to CI

vkurup:
This adds coverage reporting and sets a threshold under which builds will be failed.

* Move runtests.sh and .coveragerc into the EquiTrack directory, so that they will be in the docker container and can be used on CircleCI.
* purposefully set coverage threshold too high, to test failure on CircleCI


ntrncic:
I don't understand why we have to move to EquiTrack folder? The tests were already running in CircleCi.

vkurup:
The tests were running in CircleCI, but the .coveragerc file was having no effect since it is outside the EquiTrack folder and never gets copied into the docker image. I also like the idea of having a single script (runtests.sh in this case) that runs all testing and linting exactly as it would be run on CI.

PR #725, opened 2017-06-19
Set country_programme for agreements and interventions

ntrncic:
fixes unicef/etools-issues#470
to run locally, you must comment out `raise exception` in the migration file


EquiTrack/partners/migrations/0034_auto_20170619_1822.py
```diff
@@ -0,0 +1,65 @@
...
+                                                    start__lte=cp.to_date,)
+        for intervention in interventions:
+            wrong_cp = []
+            for rl in intervention.result_links.all():
+                if rl.cp_output.country_programme != cp:
+                    wrong_cp.append(rl.cp_output.wbs)
+            if len(wrong_cp) > 0:
+                print ("PD [{}] STATUS [{}] CP [{}] has wrongly mapped outputs {}".format(intervention.id, intervention.status, cp.wbs, wrong_cp))
+            intervention.country_programme = cp
+            intervention.save()
```
robertavram: do not save the intervention in this case. Raise an error... otherwise we introduce bad data. these records should be fixed before we migrate in prod

robertavram:
looks good just one comment

PR #730, opened 2017-06-21
Auditor portal

maximlomakin:
Auditor portal backend code.


EquiTrack/audit/serializers/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .auditor import *
+from .risks import *
+from .engagement import *
```
ewheeler: can we remove these? i can only find 2 places where this saved anyone from typing one of these words :stuck_out_tongue_winking_eye:

since we don't usually put these kinds of shortcuts in `__init__.py` then removing might reduce confusion in the future

EquiTrack/audit/models.py
```diff
@@ -0,0 +1,595 @@
...
+from utils.permissions import has_action_permission
+from utils.permissions.models import StatusBasePermission
+from utils.permissions.models.query import StatusBasePermissionQueryset
+from .transitions.conditions import AuditSubmitReportRequiredFieldsCheck, ValidateAuditRiskCategories, \
+    EngagementHasReportAttachmentsCheck, SPSubmitReportRequiredFieldsCheck, ValidateMARiskCategories, \
+    EngagementSubmitReportRequiredFieldsCheck
+from .transitions.serializers import EngagementCancelSerializer
+
+
+class AuditOrganization(BaseOrganization):
```
ewheeler: I'm thinking we should refactor this to `AuditorFirm` and use `Firm` or `firm` wherever currently we have `Organization` or `organization`

This way it will be clearer that these are things based on things from the `firms` app rather than from other places in etools. 

There doesn't seem to be a UNICEF 'term of art' to describe the category of 'vendors that perform audits' and 'vendors that perform third party monitoring activities', but i think we can safely use `firm` as it probably won't be shown on the frontend in many places but will help reduce confusion for backend developers

EquiTrack/EquiTrack/settings/base.py
```diff
@@ -401,13 +406,19 @@
     'mptt',
     'easy_pdf',
     'django_hstore',
+    'ordered_model',
```
robertavram: do we still need this? have you guys looked into the django builtin :
https://docs.djangoproject.com/en/1.11/ref/models/options/#order-with-respect-to

EquiTrack/users/views.py
```diff
@@ -362,3 +363,14 @@ def get_queryset(self):
             else:
                 queryset = queryset.filter(id__in=ids)
         return queryset
+
+
+class ModuleRedirectView(RedirectView):
```
robertavram: 🥇

EquiTrack/EquiTrack/settings/base.py
maximlomakin: Yes, we looked into it. There are some problem with using it.

- Django builtin "order_with_respect_to" have non-obvious functional. For example it implicitly add "_order" field to model.
- Django builtin "order_with_respect_to" can be used correctly only in related queryset. We need to define ordering to file types with respect to internal field (code).

With the above we think that using django-ordered-models package give more flexible and explicit functional.

PR #733, opened 2017-06-21
Change 2 tests to not rely on hardcoded PKs of ResultTypes.

caktus-philip:
Change 2 tests to not rely on hardcoded PKs of `ResultTypes`.

New code is more robust and also clearly states that subactivities aren't valid for the test.

Question -- 
There are 4 ResultTypes in the database, these are the typical ids --
`{1: u'Outcome', 2: u'Output', 3: u'Activity', 4: u'Sub-activity'}`

The original code excluded ResultType with id ==4 (Sub-activity) from being used in the test. I don't know whether this was done by design or by accident. **When I use the Sub-activity ResultType, tests fail.** My new code mimics the old behavior of excluding that ResultType.

**Is it correct to exclude ResultType 'subactivity' from being used as self.result_type in these tests?**


robertavram:
you found something here... we need to get rid of sub-activity completely as it's an invalid result-type... 
Answer: it's correct to assume that we're not using sub-activity

EquiTrack/reports/tests/test_views.py
```diff
@@ -24,7 +20,8 @@ class TestReportViews(APITenantTestCase):
 
     def setUp(self):
         self.unicef_staff = UserFactory(is_staff=True)
-        self.result_type = ResultType.objects.get(id=random.choice([1, 2, 3]))
+        # Grab a random result type that's not Sub-Activity
+        self.result_type = ResultType.objects.exclude(name__iexact=ResultType.SUBACTIVITY).order_by('?').first()
```
robertavram: this was a misunderstanding to begin with.. this result_type needs to be "Output" (thanks)

EquiTrack/workplan/tests/test_views.py
```diff
@@ -24,8 +22,8 @@ def setUp(self):
 
         self.workplan_project = WorkplanProjectFactory(workplan=self.workplan)
         self.labels = [LabelFactory() for x in xrange(3)]
-
-        self.result_type = ResultType.objects.get(id=random.choice([1,2,3]))
+        # Grab a random result type that's not Sub-Activity
```
robertavram: cool!

EquiTrack/reports/tests/test_views.py
caktus-philip: OK, thanks, I suspected there might be something more specific going on here. Will change to always use "Output" result type. I feel more comfortable not using randomness in tests anyway.

caktus-philip:
OK @robertavram, updated -- I removed the subactivity type entirely, created a corresponding migration, and adjusted the tests to always use resulttype == output. Ready for your re-review.

caktus-philip:
Just a thought — since this removes a result type, wouldn’t it be prudent to check the existing database to ensure no records exist that use that subactivity result type?

robertavram:
@caktus-philip .. I'll check before I merge this in.. looks good otherwise! thanks!

robertavram:
I confirmed.. 0 records with sub-activity

PR #740, opened 2017-06-27
Clean up activityinfo app

vkurup:
I had hoped to remove the app completely, but its migrations are referenced by the 'reports' app, so removing it would make it impossible to run `migrate` on a fresh database, unless we were to edit those migrations.


robertavram:
@vkurup feel free to edit the migrations!

vkurup:
Always happy to remove code!
I tested that I could run migrate_schemas on a new database, and that an existing migrated database didn't complain.

PR #747, opened 2017-06-30
Begin location app cleanup

vkurup:
The first commit has flake8 fixes and no significant code changes. The second commit has some minor code changes.

Here are the tasks that I believe remain to clean up this app:
- [ ] Update models.Location docstring: It says p_code should be unique, but it's only listed as unique_together. which is right?
- [ ] Do we need to address the 'Make generic' comment on the models.Location unicode method? or can we just remove the comment?
- [ ] Some CharFields and TextFields have null=True, which is [not recommended](https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.Field.null)
- [ ] We should [avoid relative imports](https://google.github.io/styleguide/pyguide.html?showone=Imports#Imports) (`from .models import blah`). Using absolute imports prevents ever accidentally importing the same module twice.
- [ ] Task parameters should be [simple, not complex](http://celerytaskschecklist.com/)
- [ ] We should create an issue for the "TODO: test performance" in urls.py (if that's still a priority)
- [ ] Management command fixes: use self.stdout instead of print, remove bare exception, limit the amount of code in the try block, use argparser
- [ ] Add docstrings
- [ ] Add tests
- [ ] Update guidelines docs with above recommendations, as needed.

That's a lot of (admittedly mostly trivial) stuff. Wondering if I should start creating PRs to fix, or just create issues for now and move on to the other apps, while you all prioritize which of those issues we should work on.


ntrncic:
@vkurup you can check off all PR tasks that have been completed

ntrncic:
please make the PR against pmp-redesign branch

vkurup:
Closed in favor of #761

PR #751, opened 2017-07-06
Ssfa set agreement dates

ntrncic:


ntrncic:
already completed in another PR

PR #753, opened 2017-07-06
Replace 3rd party library django-hstore

caktus-philip:
Replace 3rd party library `django-hstore` with native Django functionality. (hstore has been native since django 1.8)

Removing this library removes an impediment to upgrading to Django 1.10. django-hstore isn't compatible with Django 1.10. The author is aware of this but has no more time to maintain the module and is looking for a new  maintainer.
https://github.com/djangonauts/django-hstore/issues/154
https://github.com/djangonauts/django-hstore/issues/161

Also tidied imports and fixed all flake8 complaints in the file I touched (`partners/models.py`).

### DIFF
--- a/EquiTrack/EquiTrack/settings/base.py
+++ b/EquiTrack/EquiTrack/settings/base.py
@@ -398,7 +398,6 @@ SHARED_APPS = (
     'analytical',
     'mptt',
     'easy_pdf',
-    'django_hstore',
 
     'vision',
     'management',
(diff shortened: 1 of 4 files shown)


robertavram:
i think we might have removed the need for HStore altogether in pmp-redesign branch.. could you update this from that branch? looks good though.
@caktus-philip 
thanks.

caktus-philip:
Replaced by #769

PR #754, opened 2017-07-07
Add assertions to several of the test_urls() methods in t2f test cases

caktus-philip:
Add assertions to several of the `test_urls()` methods in t2f test cases and refactor existing ones to make them more declarative in style (to minimize repetition of invariant pieces). This PR lays the groundwork for another PR I have in the pipeline to refactor many urls.py files for Django 1.10 compatibility. Also fixed all flake8 complaints in all files I touched.

**I have 2 questions.**

There was one URL I couldn't figure out where to test. **Where should I test this URL, or should I just skip it?**
`url(r'^invoice_calculations/(?P<travel_pk>[0-9]+)/$', TravelEditView.as_view(), name='invedit')`

Also, in `t2f/urls.py`, `action_points_patterns` contains this -- 
`url(r'^dashboard/', action_points_dashboard_list, name='dashboard')`

The URL regex doesn't include a '$' at the end, so while this works --
`resolve('/api/t2f/action_points/dashboard/')`
This also works, which maybe shouldn't be the case -- 
`resolve('/api/t2f/action_points/dashboard/sdfasdfsd')`

**I'm happy to fix that as part of this PR, just let me know if I should.**

PR #756, opened 2017-07-10
Refactor url declarations

caktus-philip:
No functional change, just a series of cascading changes to prep for Django 1.10 compatibility.

**First**, Django changed the preferred construction of URL patterns in `urls.py`. Deprecation of the old method (which eTools uses) happened in Django 1.8, it's not available at all in Django 1.10. **Second**, Django changed how URL namespaces are applied. I applied relevant changes for all of the above in 2ed41fb5e926e023ba125658fdfdf5ce4792f0ef.

**Third**, I removed `admin.autodiscover()` which has been superfluous since Django 1.7, isn't available in Django 1.10, and was causing downstream PEP8 complaints. 

**Fourth**, I made all the files I touched `flake8` clean.

**Last but not least**, the removal of unused imports revealed that `djgeojson` wasn't being used anywhere, so I removed the reference to it in `SHARED_APPS`.

Since I made a lot of changes to the URL namespacing, I don't want to proceed with this until #754 is merged. That adds the tests that fully exercise the changes I'm making here.


caktus-philip:
Closing to be replaced by new PR against pmp-redesign branch.

PR #757, opened 2017-07-10
Move import for Django 1.10 compatibility

caktus-philip:
No functional change. This import was working by luck; the class `Value` is available via `django.db.models.functions` only incidentally. (It's not available there in Django 1.10.) Proper location is `django.db.models`.
https://docs.djangoproject.com/en/1.10/ref/models/expressions/#value-expressions

Also fixed a PEP8 complaint; this file is now flake8 clean.


robertavram:
awesome @caktus-philip thanks!

any new work.. could you pr it against pmp-redesign branch? we're doing a bunch of work there...

PR #764, opened 2017-07-12
Clean up management app

vkurup:
* Fixes flake8 errors
* moves `permissions.IsSuperUser` to EquiTrack since it might be more generally useful than just in this app
* Fixes possible bug in `LoadResultStructure`, if you were to call it without sending a country.
* Adds tests for some of the views (more to do!)
* Updates runtests.sh to allow you to run coverage and reports on just one app (for convenience while developing) e.g.:
```
./runtests.sh partners
```


EquiTrack/management/views/reports.py
```diff
@@ -14,7 +13,7 @@ class LoadResultStructure(APIView):
     def get(self, request, format=None):
         try:
             workspace = Workspace.objects.get(name=request.query_params.get('country').title())
-        except Workspace.DoesNotExist:
+        except (Workspace.DoesNotExist, AttributeError):
```
robertavram: 👍

PR #768, opened 2017-07-13
Cleanup notifications

vkurup:
First commit is only for flake8 changes. Second commit makes a simple change to use `get_or_create`, and then removes what I think is a duplicate entry in TEMPLATE_NAME_CHOICES. (But really, I can't see where TEMPLATE_NAME_CHOICES is being used, so not sure about that...)

I've created personal notes to do the following in a later pass:
* use more specific exception in send_mail
* Add docstrings
* Fix that lambda in Notification.get_template_context_entries
* Figure out how TEMPLATE_NAME_CHOICES is supposed to be used
* Add more tests


EquiTrack/notification/email.py
```diff
@@ -57,17 +58,14 @@ def get_current_site(cls):
...
-        except EmailTemplate.DoesNotExist:
-            template = EmailTemplate.objects.create(
-                name=cls.template_name,
-                description=cls.description,
-                subject=cls.subject,
-                content=cls.content
-            )
+        template, created = EmailTemplate.objects.get_or_create(
+            name=cls.template_name,
+            defaults={
```
robertavram: not sure why it was done separately.. but i think that unless there is a unique on name,  it will create other templates with the same name. maybe Silin was trying to avoid that... any thoughts?

vkurup: Hmm, this should have the [exact same behavior as the previous code](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#get-or-create), whether `name` is unique or not. The underlying implementation is actually identical. Django does a `.get()` using only the `name` attribute. (It does not look at anything in the `defaults` dictionary when doing the `.get()`). If it can't find a record, only then does it create it, and now it uses the values in the `defaults` dictionary to populate the new record. 

Example code:
```
>>> et, created = EmailTemplate.objects.get_or_create(name='trips/trip/created/updated', defaults={'description': 'vinod'})
>>> created
False
>>> et
<EmailTemplate: trips/trip/created/updated >
>>> et.description
u'The email that is sent to the supervisor, budget owner, traveller for any update'
>>> 
```

But, I should write a test to prove it, since this code is currently uncovered. :)

vkurup: Actually, after doing a little 'git archeology', it looks like the BaseEmail class is an obsolete piece of code that is no longer used anywhere. It looks like it was made obsolete in PR #328, which switched from using BaseEmail to notification.models.Notification. Silin's name is on this code only because he moved it from the Equitrack.utils module to this module.

That makes me think that it's not worth spending time writing tests for it. I'd prefer to remove it, but let me know, if you think we should keep it around. If so, I'll just undo my 'fix'.

@robertavram ^

PR #777, opened 2017-07-18
Don't allow use of default secret key

vkurup:
This prevents the accidental use of the SECRET_KEY value from the public git repo. If SECRET_KEY is not provided in the environment, the server will throw an error instead.

As I don't have access to the deployed servers' configurations, please make sure that SECRET_KEY is properly provided in the environment or else this PR will break those servers. :)

Note: I expect that this PR will break the Circle CI test build. Once I confirm that it breaks, I will add SECRET_KEY to CircleCI's environment at which point it should pass.


Dockerfile
```diff
@@ -49,4 +49,5 @@ ADD EquiTrack /code/
 WORKDIR /code/
 
 ENV DJANGO_SETTINGS_MODULE EquiTrack.settings.production
```
robertavram: doesn't this effectively set the SECRET_KEY as an env var in the docker container to "not-so-secret-key-just-for-collectstatic" .. if we're not careful we in production we might have the same issue?

vkurup: Yikes! In fact this might be worse, since it might override any env var set outside the container. Sorry about that... this is my Docker inexperience showing. I tried to find a way to use an env var from outside the container with `docker build` like I did with `docker run`, but didn't find a clear example. I'll keep looking. Alternatively, I could probably do something like just set `SECRET_KEY=blah python manage.py collectstatic` which would set `SECRET_KEY` just for that command.

robertavram: no worries...
`Alternatively, I could probably do something like just set SECRET_KEY=blah python manage.py collectstatic which would set SECRET_KEY just for that command.`
I don't know if you can do that.. and have the env just stick for that command, but I'm sure there must be a way to do it on the command... you can try building the image like that.. "ssh" into the container and check the value of the env var....
if you pass -env on run for the container.. it will override what you declared. however not a good idea to keep the env var there..
another way of doing it could be to keep the env var there, keep the run collectstatic line, set the envar again to a default value and then make sure in the settings file that the value is not the default value.. explode on default value match... (not sure how elegant it is)

vkurup: I actually already implemented the `SECRET_KEY=blah python manage.py collectstatic` and it worked in this [build](https://circleci.com/gh/unicef/etools/4118) . I cancelled the build after I saw that it was working, but the important part worked:

image

I temporarily added a 'RUN env' command to print out the environment after the collectstatic command to be sure that the SECRET_KEY didn't persist in the environment after that command, and then I undid that debug line in the next commit.

Let me know what you think.

robertavram: 👍 awesome. Thanks @vkurup !

vkurup:
@ewheeler @ntrncic The reason this was failing CodeFresh was because we needed to add SECRET_KEY to the environment, which I did in CodeFresh, so now it passes:
image

PR #788, opened 2017-07-21
Clean up settings, part 1

vkurup:
This only organizes the current settings, into the following high level sections: Core Django, Contrib Django, 3rd Party, and eTools settings. There should be no deletions or changes of existing settings (though I did delete some things that were commented out)

I've listed the settings that I think are deprecated. In a future PR, I'll remove those and fix up imports.

A few questions:
- [ ] Is postgis still being used?
- [ ] [django-celery](https://github.com/celery/django-celery) is no longer being maintained and is only needed if you're using the result backend or the admin interface. Do you know if those are being used for this project?
- [ ] Can django-allauth be removed? I think I have that in my notes.
- [ ] Anything else that you see on quick glance that I can work on removing/updating?


ntrncic:
@vkurup postgis I think is being used for locations.
django-celery is still being used
django-allauth was used for social accounts authentication. At the moment we don't have any social account users accessing tools, but I would keep it for now.

vkurup:
Thanks! Closing in favor of #808

PR #791, opened 2017-07-24
flake8 fixes for trips

vkurup:
This only fixes flake8 errors for the trips module. IIRC, trips is being deprecated, so I didn't spend any time trying to identify other issues in this app.

If that's not correct, let me know and I'll spend some time identifying issues.


vkurup:
Confirmed during meeting that trips is being deprecated, so I won't go digging for problems there.

PR #797, opened 2017-07-25
Try to cache the docker image on CI

vkurup:


vkurup:
So, just to explain this 22-commit fiasco ... :) 

I was getting frustrated by the long build times on CircleCI and I noticed that the largest chunk of time is rebuilding the docker image each time. I tried multiple approaches to get CircleCI to cache the layers of the docker image, but each attempt failed. There is something broken about Docker caching on the CircleCI 1.0 platform. I didn't test the 2.0 platform, though I saw more anecdotal reports of success with it.

My final approach was to do the entire build directly on the CircleCI VM (and not inside the docker container). If the branch is a branch that we want to push to Docker Hub, then we build it during the 'deploy' step and push it to the hub. This drops build times for PRs from 30 minutes down to about 10.

But, it looks like CodeFresh DOES properly cache Docker layers, as evidenced by how speedy those builds are. So, an alternative approach might just be to do everything on CodeFresh.

vkurup:
* ee1786e I added this branch to the list of branches that we try to deploy, by building the docker image and pushing it to Docker Hub [Success!](https://hub.docker.com/r/unicef/etools/tags/)
* cc94c9b I added a bogus command to the Dockerfile so that docker build would fail. I wanted to see that the build command failed, and that the push command didn't get run. [Success!](https://circleci.com/gh/unicef/etools/4332)

In summary, I think this approach will allow us to run tests faster, while also pushing deployable docker images to Docker Hub only if the build succeeds. (The caveat is that if we make a bad change to the Dockerfile, we will only know that the build breaks once we merge that change to a deployable branch, unless we add a conditional check to build the docker image whenever the Dockerfile changes, maybe using [something like this](https://discuss.circleci.com/t/how-to-detect-commited-file-at-test-phase/8436/2)?)

@robertavram  If this looks like a good approach to you, i'll create a new PR against develop.

robertavram:
assuming the CI will fail if build will fail... however it would be good to have some tests run in the docker container as well... if for some reason requirements don;t get properly installed (maybe unaligned) or there are some other issues that don;t fail the build but could fail the app in that container... could be catastrophic :(

maybe we can run the tests outside docker container... but on the important branches build and run the tests inside the containers as well...

the branch name is passed as an env var in $CIRCLE_BRANCH

in theory it can just be a few ifs

..let me know what you think

vkurup:
@robertavram that makes total sense. I've updated #822 with that approach, so I'll close out this one.

PR #808, opened 2017-08-04
Cleanup settings dev

vkurup:
This is a replacement for #788, just against the develop branch.

@ntrncic, thanks for the answers to my questions on that PR.

This PR should not remove or add any settings.... it only reorganizes them.

Once this is merged, I'll work on a separate PR to try removing django-suit.


vkurup:
Just to confirm what I mentioned in our meeting. This PR is a NO-OP from a functional standpoint. It re-organizes settings but does not remove, add or change any.

PR #811, opened 2017-08-04
Fix intermittent test failure

vkurup:
As @caktus-philip mentioned, there is a test in the partners app named `test_agreement_export_api` which intermittently fails because the ordering returned from the API call is not consistent.

This fixes that problem, but I'm not 100% sure that the ordering I picked (ordering by 'creation' ascending) is the right one. I chose it because it made the existing test succeed, but maybe it's better to order by 'creation' descending and adjust the test?


ntrncic:
@vkurup I would do descending order.

vkurup:
@ntrncic cool, updated.

PR #813, opened 2017-08-07
Fix t2f flake8 violations

vkurup:


robertavram:
sorry.. I merged a few things before this.. created a conflict.. could you fix it please?

labeling wip until then..

vkurup:
No worries! It's updated, but wouldn't be surprised if there are more conflicts when this is merged :)

EquiTrack/t2f/management/commands/import_dsa.py
```diff
@@ -55,7 +55,7 @@ def handle(self, *args, **options):
             if not line or not all(line):
                 continue
 
-            country, region, amount_usd, amount_local, amount_60plus_local, room_rate, eff_date, finalization_date = line
+            country, region, amount_usd, amount_local, amount_60plus_local, room_rate, eff_date, finalization_date = line  # noqa: E501
```
ntrncic: what is this `# noqa: E501`?

EquiTrack/t2f/management/commands/import_permission_matrix.py
```diff
@@ -26,7 +23,7 @@ def add_arguments(self, parser):
     @atomic
     def handle(self, *args, **options):
         country_name = options['country_name'][0]
-        import_file_path = options['import_file_path'][0]
+        # import_file_path = options['import_file_path'][0]
```
ntrncic: should we remove this completely

EquiTrack/t2f/management/commands/import_dsa.py
vkurup: That tells flake8 to ignore an E501 violation on this line. E501 is the '120 character limit'. In the rare cases that we add a `# noqa` line, we should always add the specific rule that we're allowing to break, so that other rules will still apply.

Now that I look at it though, I can probably make that comply without making it too ugly, so let me do that.

vkurup:
@ntrncic good comments! this is updated.

PR #817, opened 2017-08-09
Migrate from old model _meta API

caktus-philip:
As of Django 1.8, Django changed what's available on `model._meta` and deprecated some of the methods there. eTools still uses two of those deprecated methods (`get_field_by_name()` and  `get_all_field_names()`). They're completely gone in Django 1.10. This PR replaces the deprecated methods with Django's recommended alternatives.

Django's recommended code for replacing `get_all_field_names()` is so ugly that I felt it was important to hide it in a utility function, so I added it to `EquiTrack/utils/common/utils.py`. Feel free to tell me if I should move it elsewhere. 

https://docs.djangoproject.com/en/1.10/ref/models/meta/#migrating-from-the-old-api


EquiTrack/utils/common/utils.py
```diff
@@ -1,3 +1,28 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+from itertools import chain
+
+from django.contrib.contenttypes.fields import GenericForeignKey
+
+
+def get_all_field_names(TheModel):
```
ewheeler: would it be worth including a simple test for this? 
maybe just to confirm that fields are returned without internal-only names? 
if django changes behavior of Model._meta API in a future version we should catch even if we don't read the release notes 😛

caktus-philip: +1 to that idea. I added a test in bfff6fd189803fbf96070e3d826c566e80b3c190. Please confirm that it covers what you were thinking of.

ewheeler: 👍

PR #820, opened 2017-08-09
Log tenant to sentry if available

czue:
https://trello.com/c/OUcrwR8K/5-add-custom-variables-to-sentry-logging-to-capture-tenant-user-etc

I don't have access to the production sentry but this worked locally. The user data appears to be already there, but this adds tenant data as well. This will only work when sentry handles a request exception so won't be available in celery tasks etc. but seems a reasonable start?

This is what my sentry data looked like:

image

(the only new bit is the very last line)

Not sure if there is more I should to do verify this won't break anything, but seems reasonably safe to me (famous last words... :crossed_fingers: ).


robertavram:
Beautiful!

PR #821, opened 2017-08-11
Clean up settings (final?)

vkurup:
This PR is the final (I hope!) in the settings clean-up series. It's only really reviewable commit-by-commit. If any commits are not approved, I can create a new PR without them, but I think the end result of this is a cleaner, more maintainable settings package.

The TL;DR; of this PR:
- Removed the settings that I identified as deprecated in the previous PR #808 
- Removed an unnecessary addition to sys.path (It was adding the Equitrack/Equitrack/ folder to the path, which shouldn't be needed)
- Adds a str2bool helper function to allow consistent use of Boolean environment variables
- Moves ALLOWED_HOSTS, CACHES, AUTHENTICATION_BACKENDS, CORS_ORIGIN_ALLOW_ALL, allauth settings, AZURE settings, and saml2 settings to base.py, with overrides in local.py only, if needed
- Moves JWT_AUTH to base, with production override
- Removed settings from production which were already set properly in base.py
- Removed commented code
- Added guidelines about these changes, to help guide future changes


vkurup:
To easily review commit-by-commit, start with this link: https://github.com/unicef/etools/pull/821/commits/10c4b7aa2e6c051bac265c7181af1d16b0568aa7
and then click the 'Next >' button in the right side of the interface.

vkurup:
Sounds good. I'll move them back. It's sometimes good to have prod-only settings like this in base (inside a conditional), so that as a developer, you could easily set a couple env vars and then have your local server mimic staging or production. But if that itself is a rare need, then having them in production.py is fine. And, in any case, copy-pasting them from production.py to local.py is not difficult if the need arises.

vkurup:
@ntrncic This is updated.

EquiTrack/EquiTrack/settings/base.py
```diff
@@ -211,6 +210,9 @@
 INSTALLED_APPS = SHARED_APPS + TENANT_APPS + ('tenant_schemas',)
 
 # DJANGO: SECURITY
+ALLOWED_HOSTS = [
+    os.environ.get('DJANGO_ALLOWED_HOST', '127.0.0.1'),
```
ewheeler: is there a good reason for pulling this from the env rather than just specifying in settings files?

vkurup: It depends on how your deployment scripts are set up. At Caktus, we allow this to be specified in the environment, to make it easier to maintain multiple projects using a similar codebase: https://github.com/caktus/django-project-template/blob/master/project_name/settings/deploy.py#L59 

I don't recall how deployments are set up. Is DJANGO_ALLOWED_HOST not set in the environment of staging and prod? If that is the case, I'd be happy to hardcode this to 127.0.0.1.

ewheeler: ok. we don't really need to make it easy to deploy to different hosts so this seems like an avoidable attack surface we should close. it would be easy to mistakenly set to `*.unicef.org` somewhere in the deployment chain--which would mean that csrf protections could be undermined by anything else running on the domain... i'd be surprised if there _wasn't_ an old wordpress blog hanging around waiting to be pwnd

probably ok to keep as-is for base settings but for prod we should hardcode as `etools.unicef.org`

vkurup: Sounds good, but we also need to be able to accept staging's domain, which I think is etools-staging.unicef.org, right? Are there any other domains using the production settings?

vkurup: I got the full list of deployed servers from @ntrncic and have now updated the production ALLOWED_HOSTS setting to include all of them.

PR #822, opened 2017-08-11
Speed up feature branch CI

vkurup:
This is an attempt to speed up CI builds for feature branches, but keep the ability to run tests in the docker container for deployable branches.


vkurup:
In the first commit, I included this branch as a 'branch-to-be-built-and-test-in-docker', and then I removed it from that list in the second commit.

PR #824, opened 2017-08-14
Quieter audit tests

vkurup:
This...

1) removes extraneous output from our test output, and
2) makes things about 4 seconds faster because the fixture is loaded once per class, instead of per-testcase.


vkurup:
I've updated the PR with 2 small changes which should now silence all of the non-essential test output.

PR #826, opened 2017-08-17
Auditor portal updates

roman-karpovich:
Fields labels update with Django 1.10 writable serializers fix.


ntrncic:
@caktus-philip could you try pulling this branch in to your Django 1.10 changes and see if it's working.

PR #835, opened 2017-08-22
Add SessionAuthenticationMiddleware

caktus-philip:
Add `django.contrib.auth.middleware.SessionAuthenticationMiddleware` in prep for Django 1.10. I wasn't able to test this; please read. 

According to the Django 1.10 doc, this is a change from Django 1.9 to 1.10 -- 

> Session verification is enabled regardless of whether or not
> 'django.contrib.auth.middleware.SessionAuthenticationMiddleware' is in 
> MIDDLEWARE_CLASSES. SessionAuthenticationMiddleware no longer has any purpose 
> and can be removed from MIDDLEWARE_CLASSES.

https://docs.djangoproject.com/en/1.10/releases/1.10/

I don't expect adding this middleware to change behavior (other than seamlessly improving security), but if it does change something, I'd rather it be the result of this explicit PR rather than an implicit change in Django's behavior caused by an upgrade to 1.10.

These notes suggest that the only concerns with enabling this middleware don't apply to eTools --
https://docs.djangoproject.com/en/1.11/topics/auth/default/#session-invalidation-on-password-change

The eTools Web UI has never worked very well for me, so although I was able to log in and click around a little, I wasn't able to test something like changing my password. Please test this.


ntrncic:
tested changing password locally and it works. We don't do password management in Etools anyways so we should be good. merging this

PR #846, opened 2017-08-29
add environment-specific way to disable existing loggers

czue:
I believe setting this explicitly to `False` will allow flower logs to be visible in docker cloud.


czue:
on thing that's not clear to me - when this gets merged will it automatically update the docker image the next time i deploy it or is there a manual step for that?

EquiTrack/EquiTrack/settings/base.py
```diff
@@ -110,7 +110,7 @@
 # DJANGO: LOGGING
 LOGGING = {
     'version': 1,
-    'disable_existing_loggers': True,
+    'disable_existing_loggers': os.environ.get('DJANGO_DISABLE_EXISTING_LOGGERS', True),
```
vkurup: Just a note that this might not work as expected because environment variables are always strings, so if you set `DJANGO_DISABLE_EXISTING_LOGGERS=False` in the environment, this will result in `disable_existing_loggers: 'False'`, which is a truthy value. There is a helper function `str2bool` near the top of base.py that helps you avoid this problem.

czue: ah nice catch! my bad, should have tested this more carefully before pushing. have updated

vkurup: No worries! I only caught it because I had just worked on this recently :) 

PR looks good to me.

ntrncic:
@czue since we moved the docker dev environment to another account we will have to deploy it manually. I'll inform you when that is done.

PR #847, opened 2017-08-29
Pmp Redesign fixes

ntrncic:


EquiTrack/partners/tests/test_views.py
```diff
@@ -1367,7 +1367,7 @@ def setUp(self):
         response = self.forced_auth_req(
             'post',
             '/api/v2/interventions/',
-            user=self.unicef_staff,
```
ewheeler: i think it would be worthwhile to include at least one test case to demonstrate the current expected behavior for the previously required user role (`self.unicef_staff`)

EquiTrack/partners/validation/interventions.py
```diff
@@ -78,7 +78,7 @@ def start_date_signed_valid(i):
 def start_date_related_agreement_valid(i):
     # i = intervention
     if i.document_type in [i.PD, i.SHPD] and not i.contingency_pd and i.start and i.agreement.start and\
```
ewheeler: lots of checks in this line. are there tests that need updating now that `signed_pd_document` is required here?

EquiTrack/reports/models.py
```diff
@@ -425,7 +425,7 @@ class Indicator(models.Model):
     view_on_dashboard = models.BooleanField(default=False)
 
     class Meta:
-        ordering = ['name']
+        ordering = ['active', 'name']
```
ewheeler: a comment explaining need/usage for ordering by `active` would be nice to have

EquiTrack/partners/tests/test_views.py
ntrncic: you're right. done!

PR #853, opened 2017-08-30
Replace deprecated carto library 

caktus-philip:
Replace deprecated carto library with a non-deprecated one. The packages are similarly named so it's a little hard to tell them apart.

The "old" library (`cartodb==0.8.1`) was this one --
https://github.com/Vizzuality/cartodb

The new version of that library is written in Ruby and doesn't have a Python installer.

I replaced it with `carto-python` (a.k.a. just "carto", version 1.0.1) -- 
https://github.com/CartoDB/carto-python

This PR probably needs a little more work but I can't do it. Authentication requires credentials to which I don't have access, as documented here --
https://github.com/CartoDB/carto-python#authentication

I don't know what that is for this project (I guessed 'equitrack'). 

**Can someone else pick this up and complete it?**
 - Find out what the proper auth name is
 - Test it
 
This PR also changes `.flake8` to remove some inline comments. The new carto library included in this PR includes `pyrestcli` which includes `future` which breaks flake8's ability to handle comments. It's explained in the URLs added in 3be718748611d43be1b65c8a6f5f620f07f7e65b

PS - the new `carto-python` library is Python 3-compatible.


EquiTrack/locations/tasks.py
```diff
@@ -106,9 +107,9 @@ def update_sites_from_cartodb(carto_table):
                 carto_table.pcode_col,
                 carto_table.table_name)
 
-        sites = sql_client.sql(qry)
-    except CartoException:
-        logging.exception("CartoDB exception occured")
+        sites = sql_client.send(qry)
+    except CartoException as exc:
+        logging.exception("CartoDB exception occured {}".format(exc))
```
caktus-philip: @ntrncic FYI, when calling `logging.exception()`, "[e]xception info is always added to the logging message."
https://docs.python.org/2/library/logging.html#logging.Logger.exception

It's automatic. :-)

PR #854, opened 2017-09-01
Remove unused things

vkurup:
Connects unicef/etools-issues#641

There are items in that list that I am not sure about, so I didn't remove.
1. EquiTack.saml.stspem-old.cer (It seems old based on the name, but I don't know for sure)
2. vision.management.commands.country_import_results: refers to nonexistent command 'import_results', which was renamed to 'vision_sync'. Should this be changed?
3. partners.exports is the old django-import-export code. It seems that code has been changed to use DRF's CSV renderer. Can the old ones be removed?


ntrncic:
@vkurup 
1. yes please remove the old cer
2. Please remove the non existent command
3. Remove the old export code

vkurup:
@ntrncic This is updated. I decided to remove the export resources which I was sure were not being used. There are still some that are referenced only in the Admin, or only in v1 API urls. I can remove those in a subsequent PR if we're sure they're not needed.

PR #858, opened 2017-09-05
Change how object is unicode-d (Python 3 compatibility)

caktus-philip:
Switch to `unicode(object)` rather than `object.__unicode__()`. The latter can't be automatically translated by the 2to3 tool. No functional change.


ntrncic:
@caktus-philip please fix conflicts

caktus-philip:
@ntrncic conflicts resolved

caktus-philip:
Just realized that the only change in this PR was made to a file that was deleted in another branch, so the PR has no changes. Closing.

PR #866, opened 2017-09-12
add example monitoring endpoint for db and celery

czue:
Picking this back up. The idea here is to provide a URL endpoint (in this case https://etools-whatever.unicef.org/monitoring/) that exercises a bunch of different services that we can then connect to New Relic to do application monitoring and be notified if a single service goes down.

This is just scaffolding for a couple simple checks. We could obviously build out a bunch more over time.

This is largely inspired by [how we did monitoring at dimagi](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/hqadmin/service_checks.py).

Once this is merged in I can hook it up to new relic in the develop environment and see if it's working properly.


czue:
@robertavram fyi I switched this to a CBV. let me know if you have any other feedback!

PR #867, opened 2017-09-12
Improve partners organization views tests

caktus-philip:
This PR reorganizes and improves test coverage for the partner organization views. Prior to this they were all lumped into one large test case. I've broken that test case up into 4 -- a test for the list view, the CSV portion of the list view, the create view, and everything else (retrieve/update/delete).

Reorganized tests are a bit faster (less setup), provide better coverage, and have docstrings.

Main changes were --
 - Add the new list test case (mostly using tests that were already written) (90543bcfa408ad7de15f0becb1cc4195418d0033)
 - Add the new CSV test case (two new tests) (ee020f3b9ee08679074732651c90812b71d6e397)
 - Add the new create test case (using tests that were already written) (f9d3e5a53fa4f55cd329d989ad7ca93acd1de1e5)


caktus-philip:
Not sure why this failed. CircleCI seems broken at the moment; that might be the problem. The only failure I get locally is the one fixed in PR #865 so maybe it's important that that gets merged first. :-)

EquiTrack/partners/tests/test_views.py
```diff
@@ -66,8 +70,255 @@ def test_urls(self):
...
+class TestPartnerOrganizationListViewForCSV(APITenantTestCase):
+    '''Exercise the CSV-generating portion of the list view for PartnerOrganization.
+
+    This is a separate test case from TestPartnerOrganizationListView because it does some monkey patching in
+    setUp() that I want to do as infrequently as necessary.
+    '''
+    def setUp(self):
+        # Monkey patch the serializer that I expect to be called with a wrapper that will set a flag here on
+        # my test case class before passing control to the normal serializer.
+        class Wrapper(PartnerOrganizationExportSerializer):
```
ntrncic: why is this needed?

caktus-philip: It allows the test to verify that the view is using the serializer it expects the view to use. Later in the test it asserts this --
```
self.assertTrue(self.wrapper_called)
```

It allows the view test to ignore serializer-specific details (e.g. which fields are included in the output, were all serialized correctly, etc.) which more properly belong in a dedicated serializer test. As long as the view test knows the correct serializer is used, it can assume that the serialization went OK. (Unfortunately this serializer doesn't have a dedicated test, but the view tests weren't testing it properly before, so this is still a step forward.)

It's conceptually similar to Django's `assertTemplateUsed()`. Neither Django nor DRF offer `assertSerializerUsed()` so I rolled my own.

Would you like better comments/docstrings in the code to explain this? It sounds like they're needed.

ntrncic: thank you for explaining, it's definitely a step forward! Maybe update the comments.

PR #881, opened 2017-09-18
Management module sketch

czue:
@ewheeler I started sketching out just the data modeling side for how we could start tracking data/compliance issues. as a proof of concept i just took the script that @ntrncic wrote and tried to see if I could get it to generate appropriate issues in the database instead of just printing them out. I just hooked up one for a proof of concept and to make sure it seemed like it would work well, and then stopped to check in before going farther down this path. You can see the created issues in the admin here:

image

If this looks good I imagine next steps would be:

- [x] Validate the data model by using it for the entire script and confirm I can generate an equivalent to the print out from the database
- [x] Start thinking through the plumbing / glue code that would be able to make it easy for developers to drop in new "checks", and then port the script to use that framework.
- [x] Add code/logic to recheck any issue (including enforcing IssueCheck ID uniqueness)
- [x] Add tests for all functionality
- [x] Add another task to automatically recheck all flagged issues
- [x] Fix CircleCI
- [x] Replace the management command with a generic periodic task in celery that runs all checks on a regular schedule
- [ ] Start making some of this data accessible via APIs for admins, etc.

One big design decision is whether `FlaggedIssue` should be a tenant schema or not. I opted to make it one for two reasons: 1) to be able to use a `GenericForeignKey` against the offending model, and 2) to simplify the use case of country offices wanting to view/resolve their own issues. However this does mean that doing stuff in the aggregate over multiple tenants gets more complicated. Am very open to any thoughts on whether that should change.


EquiTrack/management/management/commands/pmp_redesign_prep.py
```diff
@@ -0,0 +1,89 @@
+from django.core.management import BaseCommand
```
czue: (this is basically copy/pasted from a script that @ntrncic sent me with a few relevant changes in https://github.com/unicef/etools/pull/881/commits/b48ae9505dc68ef3d3a05fbd8c923e115835e00f

ntrncic:
@czue Hey this is definitely the way forward. I'm just not sure about the status and updates to the records. Let's say the product team fixes some of the records and does not mark them as completed in the list. How do we know if they have been resolved.  If we rerun the script will these records just be removed or do we update the status automatically? 
Maybe we don't have a status just a date as when the script ran last.

For `FlaggedIssue` being a tenant schema I think that's how we would want it to be, in order for each country to manage their own data.

I see the task you have added to make it more generic for developers to add new "check" scripts. That's definitely needed as we will have model and process changes all the time.

ewheeler:
@czue  good start and next steps sound great. agree that `FlaggedIssue` as a tenant schema makes the most sense

for inspiration around glue/plumbing, this situation reminded me of how casepro implements 'rules' for categorizing sms messages:
basically, a particular `Rule` has some number of `Test`s along with `Action`s to take on records that are matched by any of the `Rule`'s `Test`s. 
https://github.com/rapidpro/casepro/blob/master/casepro/rules/models.py
Not directly applicable as our tests will often not operate on just one instance of one model, but its a nice, extensible way to model the problem

on @ntrncic's point on verification & resolution of flagged issues, two approaches come to mind:
1) use django signals to rerun checks/rules on save. management app ignores most signals and just reruns checks on known issues. new issues would only be noticed when the periodic job runs but resolved issues could be removed from the issue list almost as soon as they're fixed
2) if the issue has a user-facing banner/warning message, add a button to 'recheck' and offer similar buttons in the management interface

czue:
Nice, thanks for the feedback @ewheeler and @ntrncic!

Regarding verification and resolution - I 100% agree there should be a "recheck" functionality built into the framework, so that any `FlaggedIssue` should be able to check if it is still "active". I think we'll be able to design it in such a way that that is possible. Then, however we want to implement the resolution (signal, user-triggered, a periodic task that cycles through all `FlaggedIssue`s, or all of the above) it should be possible to arbitrarily check any `FlaggedIssue` whenever we want.

@ewheeler thanks for passing along that link. One thing that's different about how I've been thinking about our situation is that I'm imagining rules are only added by developers and might be arbitrarily complex, so likely wouldn't live in the DB. In my head a closer analogy was something like django middleware, where there is some global registration mechanism that builds up a set of functions that get called. Were you thinking that we would ever have rules that would be user-generated and so should try and come up with something that doesn't require code changes to deploy a new rule?

czue:
@ewheeler @ntrncic okay I've taken an initial pass on the first two items from the checklist, sketched out the "glue" and porting the entire management command over to the new architecture to validate that everything is possible and works. It does!

I ended up modeling the checks architecture after other things in django like middleware and such, though the glue is not too much so we could easily play with it if we prefer a different architecture or want something that's DB-backed. This just seemed simplest to me as a first pass.

So the way you would add new checks would be:

- Write a new subclass of `BaseIssueCheck` ([examples here](https://github.com/unicef/etools/pull/881/files#diff-dd779c176be40bc59dfa5172eeab7e27))
- Add the class to `settings.py` ([here](https://github.com/unicef/etools/pull/881/files#diff-50fe3d759ea19799c6282578ebde3f4cR541))

I've also updated the task list above based on the latest requirements.

Keen to get your input on this again! Once I've got your feedback I can keep pushing ahead.

ewheeler:
@czue this is great!

definitely cool with having these defined in code. nice approach you've taken here


we get a lot of change requests around error message text content. in an ideal future, we would have a way for product leads to edit directly.. perhaps in the management module? maybe this is a good opportunity to move incrementally in that direction--and extract these `ImproperlyConfigured` messages to a separate file? Not high priority by any means.. just noting bc it occurred to me

czue:
@ewheeler interesting - sounds worth talking through some of the details of how that should work. e.g. who's allowed to update messages, and where. obviously editing individual messages against a `FlaggedIssue` would be very easy to expose, but backing the default message that gets spit out by the various `IssueCheck`s might be trickier to do from the database. But probably something we can figure out if it's an important requirement.

czue:
@ntrncic @ewheeler Ok, I think this is basically ready. I haven't build the API yet, but that doesn't seem like it needs to be a blocker for review and/or merge. Would be great to get your feedback!

czue:
Yep, sounds good. I'm bad at naming things.

EquiTrack/management/README.md
```diff
@@ -0,0 +1,88 @@
+# Management App
```
czue: read this first

EquiTrack/management/issues/checks.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+ModelCheckData = namedtuple('ModelCheckData', 'object metadata')
+
+
+class BaseIssueCheck(object):
+    """
+    Base class for all Issue Checks
+    """
+    __metaclass__ = ABCMeta
```
czue: just so we can use the `@abstractmethod` decorator

EquiTrack/management/issues/checks.py
```diff
@@ -0,0 +1,131 @@
...
+    model = None
+    issue_id = None
```
czue: can we come up with a better name?

EquiTrack/management/issues/checks.py
```diff
@@ -0,0 +1,131 @@
...
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+from django.db.models import Model
+from django.utils.module_loading import import_string
+from EquiTrack.util_scripts import run
+from .exceptions import IssueFoundException, IssueCheckNotFoundException
+from management.models import FlaggedIssue, ISSUE_STATUS_RESOLVED
+
+
+ModelCheckData = namedtuple('ModelCheckData', 'object metadata')
```
czue: just used to lump these two things together and pass them around.

EquiTrack/management/issues/checks.py
```diff
@@ -0,0 +1,131 @@
...
+    model = None
+    issue_id = None
+
+    def __init__(self):
+        if self.model is None or not issubclass(self.model, Model):
+            raise ImproperlyConfigured('Issue checks must define a model class that subclasses models.Model!')
+        if not self.issue_id:
+            raise ImproperlyConfigured('Issue checks must define a unique ID!')
+
+    def check_all(self):
```
czue: this is the main entry point when running all checks against an individual issue check type.

EquiTrack/management/issues/checks.py
```diff
@@ -0,0 +1,131 @@
...
+    import_path is the full Python path to the class.
+    """
+    IssueCheck = import_string(import_path)
+    if not issubclass(IssueCheck, BaseIssueCheck):
+        raise ImproperlyConfigured('Issue Check "%s" is not a subclass of "%s"' %
+                                   (IssueCheck, BaseIssueCheck))
+    return IssueCheck()
+
+
+def run_all_checks():
```
czue: main entry point 1: run all checks against all data.

EquiTrack/management/issues/checks.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    Run all configured issue checks. Note that this function might take a long time to complete on a large
+    database.
+    """
+    for issue_check in get_issue_checks():
+        issue_check.check_all()
+
+
+def recheck_all_open_issues():
```
czue: main entry point 2: recheck all issues that have already been flagged by previous runs.

EquiTrack/management/issues/project_checks.py
```diff
@@ -0,0 +1,167 @@
...
+from django.db.models import Q
+from .checks import BaseIssueCheck, ModelCheckData
+from management.issues.exceptions import IssueFoundException
+from partners.models import Agreement, Intervention, InterventionAmendment, AgreementAmendment
+from partners.validation.interventions import InterventionValid
+from reports.models import CountryProgramme
+
+
+# todo: these can probably move closer to the models they are associated with, but just
+# starting with them here as a proof of concept
```
czue: this whole file is a 1:1 port of the scripts that @ntrncic wrote. each issue check class maps to a single function that his function called.

EquiTrack/management/issues/project_checks.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class ActivePCANoSignedDocCheck(BaseIssueCheck):
```
czue: https://github.com/unicef/etools/pull/881/files#diff-a16effa7ba948efccfd20949580c740eR47

EquiTrack/management/issues/project_checks.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def run_check(self, model_instance, metadata):
+        if not model_instance.attached_agreement:
+            raise IssueFoundException(
+                '{} Agreement [{}] does not have a signed PCA attached'.format(model_instance.agreement_type,
+                                                                               model_instance.id)
+            )
+
+
+class PdOutputsWrongCheck(BaseIssueCheck):
```
czue: https://github.com/unicef/etools/pull/881/files#diff-a16effa7ba948efccfd20949580c740eR55

EquiTrack/management/issues/project_checks.py
```diff
@@ -0,0 +1,167 @@
...
+            if rl.cp_output.country_programme != cp:
+                wrong_cp.append(rl.cp_output.wbs)
+        if len(wrong_cp) > 0:
+            raise IssueFoundException(
+                "PD [P{}] STATUS [{}] CP [{}] has wrongly mapped outputs {}".format(
+                    model_instance.id, model_instance.status, cp.wbs, wrong_cp)
+            )
+
+
+class InterventionsAssociatedSSFACheck(BaseIssueCheck):
```
czue: https://github.com/unicef/etools/pull/881/files#diff-a16effa7ba948efccfd20949580c740eR70

EquiTrack/management/issues/project_checks.py
```diff
@@ -0,0 +1,167 @@
...
+        if not validator.is_valid:
+            raise IssueFoundException(
+                'intervention {} of type {} is invalid: (Status:{}), Errors: {}'.format(
+                    model_instance.id, model_instance.document_type, model_instance.status,
+                    ', '.join(validator.errors)
+                )
+            )
+
+
+class PDAmendmentsMissingFilesCheck(BaseIssueCheck):
```
czue: https://github.com/unicef/etools/pull/881/files#diff-a16effa7ba948efccfd20949580c740eR80

EquiTrack/management/issues/project_checks.py
```diff
@@ -0,0 +1,167 @@
...
+        if not model_instance.signed_amendment:
+            raise IssueFoundException(
+                'intervention {} type {} status {} has missing amendment file'.format(
+                    model_instance.intervention.id,
+                    model_instance.intervention.document_type,
+                    model_instance.intervention.status)
+            )
+
+
+class PCAAmendmentsMissingFilesCheck(BaseIssueCheck):
```
czue: https://github.com/unicef/etools/pull/881/files#diff-a16effa7ba948efccfd20949580c740eR87

EquiTrack/management/models.py
```diff
@@ -0,0 +1,79 @@
...
+    # generic foreign key to any object in the DB
+    # https://docs.djangoproject.com/en/1.11/ref/contrib/contenttypes/#generic-relations
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField(db_index=True)
+    content_object = GenericForeignKey('content_type', 'object_id')
+
+    date_created = models.DateTimeField(auto_now_add=True)
+    date_updated = models.DateTimeField(auto_now=True)
+    issue_category = models.CharField(max_length=32, choices=ISSUE_CATEGORY_CHOICES, default=ISSUE_CATEGORY_DATA,
+                                      db_index=True)
```
czue: not really using this field yet - was meant to be able to differentiate between "data" and "compliance" down the road.

EquiTrack/management/models.py
```diff
@@ -0,0 +1,79 @@
...
+        help_text='A readable ID associated with the specific issue, e.g. "pca-no-attachment"',
+        db_index=True,
+    )
+    message = models.TextField()
+
+    def recheck(self):
+        from management.issues.checks import get_issue_check_by_id  # noqa
+        check = get_issue_check_by_id(self.issue_id)
+        try:
+            check.run_check(self.content_object, metadata=check.get_object_metadata(self.content_object))
```
czue: this is why `get_object_metadata` is needed - for rechecks

EquiTrack/management/issues/project_checks.py
```diff
@@ -0,0 +1,167 @@
...
+    issue_id = 'pd_outputs_wrong'
+
+    def get_objects_to_check(self):
+        cps = CountryProgramme.objects.filter(invalid=False, wbs__contains='/A0/')
+        for cp in cps:
+            interventions = Intervention.objects.filter(
+                start__gte=cp.from_date, start__lte=cp.to_date
+            ).prefetch_related('result_links')
+            for intervention in interventions:
+                yield ModelCheckData(intervention, {'cp': cp})
```
czue: example of metadata being passed when checking all issues.

EquiTrack/management/issues/project_checks.py
```diff
@@ -0,0 +1,167 @@
...
+# todo: these can probably move closer to the models they are associated with, but just
+# starting with them here as a proof of concept
+
+
+class ActivePCANoSignedDocCheck(BaseIssueCheck):
+    model = Agreement
+    issue_id = 'active_pca_no_signed_doc'
+
+    def get_queryset(self):
+        return Agreement.objects.filter(agreement_type=Agreement.PCA).exclude(status='draft')
```
czue: simple use case with no metadata, can just be handled by a queryset

EquiTrack/management/tests/test_issue_checks.py
```diff
@@ -0,0 +1,131 @@
+from django.core.exceptions import ImproperlyConfigured
```
czue: this test file is a good place to look for super simple implementations of issue checks

EquiTrack/management/issues/checks.py
czue: I renamed to `check_id` which I think is slightly better (though still not great)

czue:
@ewheeler @ntrncic I updated this so that the checks are configurable from the database. Just did the simplest thing which was add a config model with the check ID and an `is_active` field. Could do something fancier if you thought was worth it.

Also I wasn't sure where to put it - it had to be in a new app since it's not a `tenant` app. I chose to make a new app called `environment` - representing data relevant to a particular env of etools - but very open to other suggestions / conventions you can come up with.

EquiTrack/management/issues/checks.py
```diff
@@ -0,0 +1,155 @@
+from abc import ABCMeta, abstractmethod
+from collections import namedtuple
+import logging
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+from django.db.models import Model
+from django.utils.module_loading import import_string
+from EquiTrack.util_scripts import run
```
ewheeler: ha! i've been bugging Rob to get rid of this dodgy `util_scripts` file for a while...

`utils.common.utils` would be a much better home for a new and improved `every_country` class

czue: sure thing, will update. did you want me to move the whole contents of the file or just that function for now?

ewheeler: Thanks. Just that function

czue: alright moved. wasn't sure if you also wanted to move `run` - I didn't because it was a bit harder to grep for, but I definitely can if desired

ewheeler: sorry lets move `run` too. 
no need to extract/refactor util_scripts. just duplicate `run`+`every_country` to a better place (either utils.common.utils or management.utils?) so we don't rely on `util_scripts` and i can delete it while Rob isn't looking

czue: alrighty, done. I took the liberty of renaming it to `run_on_all_tenants` as well since that seemed a bit clearer....

czue:
:joy:

PR #890, opened 2017-09-20
Redesign fixes

ntrncic:


EquiTrack/partners/models.py
```diff
@@ -1183,9 +1182,9 @@ class Intervention(TimeStampedModel):
         ENDED: [CLOSED]
     }
     TRANSITION_SIDE_EFFECTS = {
-        SIGNED: [side_effect_one, side_effect_two],
+        SIGNED: [side_effect_ssfa_signed],
```
robertavram: i think this is done on save.. no need for these side effects I don't think

EquiTrack/partners/validation/agreements.py
```diff
@@ -46,6 +46,9 @@ def agreement_transition_to_ended_valid(agreement):
 
 
 def agreements_illegal_transition(agreement):
+    if agreement.agreement_type == agreement.SSFA and agreement.status == agreement.SUSPENDED:
+        raise TransitionError(['Please unsuspend the SSFA on the PD/SSFA Page and the record on the Agreement page '
```
robertavram: this transition error will never hit the front-end on save... are you using this to update some error string on the recurring script?

EquiTrack/partners/models.py
ntrncic: yup missed that save code somehow even though I looked for it.
in the save we have
`if self.status == self.ACTIVE and self.agreement.status != Agreement.SIGNED:`
we should change to `if self.status == self.SIGNED`

EquiTrack/partners/validation/agreements.py
robertavram: also.. nevermind the agreement's status for SSFA.... the agreement status actually gets set by the intervention on save.... if you keep this validation will always hit after you suspend.... what are you trying to achieve?

robertavram: oh I see. this is on the agreement... nvmd...

ntrncic: It would have hit the code if transition was to `draft`.  Anyways the unsuspend button is removed for SSFA agreements in the frontend  therefore no need for this. will remove

EquiTrack/reports/serializers/v1.py
```diff
@@ -130,9 +131,12 @@ class Meta:
 class CountryProgrammeSerializer(serializers.ModelSerializer):
     expired = serializers.ReadOnlyField()
     active = serializers.ReadOnlyField()
-    special = serializers.ReadOnlyField()
+    special = serializers.SerializerMethodField()
```
robertavram: this would work if we're not checking on the front-end.. probably better like this... however... when we return these CPs to PRP frontend (partners.views.v2 L213) we don't use the serializer.. so this will not help....

EquiTrack/partners/validation/agreements.py
robertavram: coming back to this after thinking about it... we never initiate the transition from the front end... so this validation will never hit on save, also our recurring scripts should never attempt to transition SSFA agreements.. not sure why you needed this.... brain is messed up.. i might be missing something :)

EquiTrack/reports/serializers/v1.py
ntrncic: this view is only used in agreements.

ntrncic: in the partners.view.v2 we see all active Cps even special ones

EquiTrack/partners/validation/agreements.py
ntrncic: removing it. there was an unsuspend button in agreements that sends a PATCH request to change status. It resulted in an invalid something something error. They wanted the error changed. But now I've asked frontend to remove the unsuspend button for SSFAs

EquiTrack/reports/serializers/v1.py
robertavram: I don't believe this is used in agreements... I think the agreement CPs are populated with the locally stored data from PMPDropdownsListAPIView....  and I think that's where we'd need to label the specials as non-special for Palestine...

ntrncic: I thought it was in the `PMPDropdownListAPIView` too but checked with Daniel and it's definitely this view. Also tested it locally after making the change and in Palestine the CP comes up now.

robertavram: awesome... thanks for going through the pain...

PR #893, opened 2017-09-21
Add Disaggregation Tests (and cleanup others)

vkurup:
This adds some simple tests of the serializer and views. There may be other tests that were appropriate, if there are business rules regarding them (i.e., Might make sense that you need at least 2 DisaggregationValues per Disaggregation, else you're not really dis-aggregating anything), but I didn't implement anything like that.

I also cleaned up some of the existing tests to use `reverse` and to speed up tests by moving model creation to the `setUpTestData` class method.


robertavram:
@vkurup thanks.. :)

vkurup:
You're welcome!

PR #898, opened 2017-09-22
Convert partners models to Python 3-compatible str() methods

caktus-philip:
Add `@python_2_unicode_compatible` decorator to all 24 models in partners that had a `__unicode__()` method. (03f4d1eac0ed2d0e67b8089760ad14670a686f5e)

Add tests that use non-ASCII to challenge the `__str__()/__unicode__()` methods of all partners models that have those methods. Added lots of factories to support those tests. Lightly reorganized the imports in `test_models.py` to conform with the 4 sections rule (python/django/3rd party/project). (48f71a359dd9bc9c7f11e8a6c1f44a1cde853ace)

I recently added [EquiTrack/tests/test_all_models.py](https://github.com/unicef/etools/blob/develop/EquiTrack/EquiTrack/tests/test_all_models.py) to ensure that all eTools models eventually implement `@python_2_unicode_compatible`. The idea was that as I implemented it for each app (funds, partners, etc.) that I would also change `test_all_models.py` to include that app in its tests. However, while doing the work for this PR I realized that the code in `test_all_models.py` isn't robust enough to handle the variety of models in `partners`, so `test_all_models.py` needs an update. I will commit that in a separate PR, though, because this one is already long.

This is part of https://github.com/unicef/etools-issues/issues/746

### DIFF
--- a/EquiTrack/EquiTrack/factories.py
+++ b/EquiTrack/EquiTrack/factories.py
@@ -15,6 +15,7 @@ from trips import models as trip_models
 from reports import models as report_models
 from locations import models as location_models
 from partners import models as partner_models
+from supplies import models as supplies_models
 from funds.models import (
     Donor,
     FundsCommitmentHeader,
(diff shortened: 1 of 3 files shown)


caktus-philip:
I can remove those tests, but they all pass. I feel that if the models are present, they ought to be tested. If they're not used, the models ought to be removed.

It looks like the PCA models (`PCA`, `PCAFile`, `PCAGrant`, `GwPCALocation`, `PCASector`, and `PCASectorGoal`) are referenced quite a lot, but only in the V1 API code. Those URLs are still accessible (at least as far as `partners/urls.py` is concerned), so theoretically those models are still in use. Here's an example --
https://github.com/unicef/etools/blob/develop/EquiTrack/partners/urls.py#L91

If the models aren't in use, I'd rather withdraw this PR, delete the models, and then re-submit this PR against the slimmed-down `partners/models.py`. I know it will be a bunch of work to remove those models, but I don't mind doing it.

Please let me know if you'd like for me to remove the models or just remove the str/unicode tests.

ntrncic:
@caktus-philip remove all the models! thanks.

caktus-philip:
@ntrncic   The delete key is my favorite key! 

For models related to supplies, I see `SupplyPlan` and `DistributionPlan`. Do those sound right? If so, I'll close this PR and open a new one to remove the following models and everything that references them (views, serializers, URLs, etc.) -- 
 - PCA
 - PCAFile
 - PCAGrant
 - GwPCALocation
 - PCASector
 - PCASectorGoal
 - SupplyPlan
 - DistributionPlan

caktus-philip:
Closing for now. Will remove PCA-related models and come back to this one.

PR #904, opened 2017-09-26
Prp refactoring flake8 fixes

csabadenes:


csabadenes:
The pull request has two parts:
 - changes the flake8 rules to be less restrictive
 - fixes the current flake8 errors through the project, if it is decided to keep the rules restrictive

EquiTrack/.flake8
```diff
@@ -1,6 +1,7 @@
 [flake8]
 max-line-length = 120
-ignore = F405
+# ignore = F405
+ignore = F405,E121,E123,E126,E226,E24,E704,W503,W504
```
vkurup: JMHO, but I would prefer not to ignore these rules. They make things more readable for me. If we do change this, then the guidelines document should be updated with appropriate explanations: https://github.com/unicef/etools/blob/master/docs/guidelines.rst#flake8-exceptions

csabadenes: As far as i can see, each of these rules(except F405), were actually ignored until now, because of flake's default IGNORE settings. Adding F405 to the .flake8 ignore list overwrote the default settings, and a bunch of linting errors popped up.
I have added back the default IGNORE list + F405 as a quickfix, but i have also fixed the related flake errors.. i also think we could remove the default rules(all but F405) from the ignore list, they make the code more readable indeed, they seem quite useful.

If it is decided to keep the default IGNORE's, i will update the guidelines.

csabadenes: I have removed the default ignores(E121,E123,E126,E226,E24,E704,W503,W504) from the flake8 settings for now.

vkurup: Ahh, now I see. Yes, I missed the point about F405. Adding that rule to the prp-refactoring branch causes the default ignore list to be overwritten. (I was still looking at the develop branch which has no 'ignore' specified)

My preference: I am OK with adding F405 to the ignore list. Of the 'default ignores', I'd like to reinstate E226, E24, and E704, since I think those make code more readable. (JMHO, again)

csabadenes: I have actually re-enabled/reinstated all of the ignored defaults before reading your reply. Right now, only F405 is on ignore.

I can change them again if you or anyone else wants. (As far as i can recall, most of the errors i have fixed were  E12*, i think these rules are also nice to keep for readability.)

vkurup: Even better! Thank you.

robertavram:
soo... even though I agree in an ideal world we'd make those requirements that flake8 defaults to ignore to not ignore them, for the purpose of this branch I'd like to keep them ignored... those spacing fixes could potentially be done against develop... but I'm trying to avoid conflicts on this PR as much as possible when it comes to merging code from develop, and given the number of files changed I'd prefer to handle the flake8 fixes after pmp-remediation production push and after prp-refactoring merged into develop....

For now i'll just add the default ignores back in - directly in prp-refactoring. @csabadenes we'll keep your branch on opened but parked until all the code merges.

robertavram:
👍  on the changes though!

caktus-philip:
@csabadenes It all makes sense once I read the back and forth between you and Vinod. Thanks for doing this. I'm +1 on leaving the ignore list as short as possible, even if I don't always agree with flake8's opinions.

csabadenes:
Thanks all for the input/clarifications. I can reapply this on develop once prp-refactoring is finished/merged.

robertavram:
closing this.. can we redo later? too many conflicts...

Issue #908, opened 2017-09-27
Remove Hstore


robertavram:
- [x] Remove all [hstore fields](https://github.com/unicef/etools/pull/911)
- [x] Remove dependencies from old migrations
- [ ] Run migrations on prod
- [ ] Remove references in DB init scripts and other places

caktus-philip:
@czue FYI, I did some HStore-related work in a previous PR. https://github.com/unicef/etools/pull/769

czue:
nice thanks for the heads up! :raised_hands:

czue:
@robertavram so it appears that there's [one place in the code still using `hstore`](https://github.com/unicef/etools/blob/develop/EquiTrack/partners/models.py#L1869-L1871) and there is a small amount of data in that field. You had mentioned on slack that we need to do a migration - would this be to migrate the data to a new place or just a schema migration to remove the field? If the former, where should the data go?

robertavram:
@czue we can get rid of the data altogether.

PR #911, opened 2017-09-28
remove "activities" field

czue:
https://github.com/unicef/etools/issues/908

we need to remove this field first, then when migrations have run we can remove the HStore dependency entirely.

@ntrncic @robertavram is this what you had in mind?


robertavram:
yep...

PR #915, opened 2017-09-28
Remove the trips app and all references to it

caktus-philip:
Non-obvious changes --
 - trips was the only thing referencing the `django-datetime-widget`, so I removed that library from the requirements.
 - `t2f/models.py` contained an inaccurate docstring that mentioned trips; I corrected it.
 - Remove the property `trips` from `partners.PartnerOrganization`
 - Remove the properties `programmatic_visits` and `spot_checks` from `partners.PCA`

The commit comment isn't correct; I got confused about the location of the `programmatic_visits` and `spot_checks` properties.


EquiTrack/partners/templatetags/intervention_tags.py
```diff
@@ -69,42 +68,6 @@ def show_work_plan(value):
...
-        data.headers = trip_summary[0].keys()
-        for row in trip_summary:
-            data.append(row.values())
-
-        return data.html
-
-    return '<p>No trips</p>'
-
-
-@register.simple_tag
```
ntrncic: we shouldn't remove this line of code `@register.simple_tag`

EquiTrack/trips/models.py
```diff
@@ -1,802 +1,14 @@
...
-
-
-post_save.connect(ActionPoint.send_action, sender=ActionPoint)
+# Models removed so that migration can run to drop the associated tables. Once
+# that migration has been run on production, this entire app can be deleted and
+# removed from INSTALLED_APPS.
+from django.db import connection
 
 
+# get_report_filename() is used in the trips initial migration.
```
ntrncic: when we squash migrations then can we remove this?

EquiTrack/partners/templatetags/intervention_tags.py
caktus-philip: Good catch! But the diff is a little misleading. The `@register.simple_tag` decorator was associated with the function that I removed (`show_trips()`) and also with the function below it (`show_fr_fc()`). I deleted `show_trips()` and its decorator but according to the diff I left behind the decorator associated with `show_trips()` and deleted the one associated with `show_fr_fc()`. Either way, it amounts to the same thing. `show_fr_fc()` still has the appropriate decorator. It's visible if you view the whole file.

EquiTrack/trips/models.py
caktus-philip: Yes.

caktus-philip:
@ntrncic questions answered, I believe. Ready for your re-review when you get a chance.

PR #916, opened 2017-09-29
Convert reports models to Python 3-compatible str() methods

caktus-philip:
Convert all models in `reports` to Python 3-compatible `__str()__` methods using Django's `@python_2_unicode_compatible decorator` while retaining compatibility with Python 2.

Add factories as necessary for tests.

In `EquiTrack/tests/test_all_models.py`, remove `reports` from the list of modules not tested for Python 3-compatibility since it's now compliant.


ntrncic:
@caktus-philip please resolve conflicts.

caktus-philip:
@ntrncic Branch is up-to-date; please re-review when you have a chance.

PR #919, opened 2017-10-02
API endpoints for retreiving and updating Disaggregations

czue:
Took a crack at what I thought these two items meant:

- Endpoint for getting/patching/deleting disaggregations (disabled deleting)
- Test endpoint for getting/patching/deleting disaggregations

@robertavram cc @vkurup since you wrote the other Disaggregations endpoint / tests.

### DIFF
--- a/EquiTrack/reports/serializers/v2.py
+++ b/EquiTrack/reports/serializers/v2.py
@@ -85,9 +85,11 @@ class DisaggregationSerializer(serializers.ModelSerializer):
         except KeyError:
             pass
         else:
+            found_values = []
             for value_data in values_data:
                 value_id = value_data.get('id', None)
                 if value_id:
+                    found_values.append(value_id)
                     try:
                         value = DisaggregationValue.objects.get(disaggregation=instance, id=value_id)
                         for k, v in value_data.items():
@@ -100,7 +102,10 @@ class DisaggregationSerializer(serializers.ModelSerializer):
                             )
                         )
                 else:
-                    DisaggregationValue.objects.create(disaggregation=instance, **value_data)
+                    value = DisaggregationValue.objects.create(disaggregation=instance, **value_data)
+                    found_values.append(value.id)
+            # delete any values that weren't specified
+            instance.disaggregation_values.exclude(id__in=found_values).delete()
         return super(DisaggregationSerializer, self).update(instance, validated_data)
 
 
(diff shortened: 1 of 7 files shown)


EquiTrack/reports/serializers/v2.py
```diff
@@ -67,9 +72,29 @@ def create(self, validated_data):
         values_data = validated_data.pop('disaggregation_values')
         disaggregation = Disaggregation.objects.create(**validated_data)
         for value_data in values_data:
+            if 'id' in value_data:
+                # todo: is this valid? silently create new DisaggregationValues if IDs are passed in
+                del value_data['id']
```
robertavram: complain if Ids are passed in on create..

EquiTrack/reports/serializers/v2.py
```diff
@@ -67,9 +72,29 @@ def create(self, validated_data):
...
             DisaggregationValue.objects.create(disaggregation=disaggregation, **value_data)
         return disaggregation
 
+    def update(self, instance, validated_data):
+        try:
+            values_data = validated_data.pop('disaggregation_values')
+        except KeyError:
+            pass
+        else:
+            for value_data in values_data:
```
robertavram: A few scenarios needed here:
1. If someone passes in a list of values but removes one of the existing elements.. we need to be able to pop the value.
2. We need to make sure that we don't change values that are not associated with that particular disaggregation... like patching a disaggregation with a disaggregation value that has an id but does not relate to this disaggregation
3. Do not allow editing of disaggregations or disaggregation values if the disaggregations are already associated with any indicator.

czue: @robertavram just to make sure I'm following can you confirm:

> If someone passes in a list of values but removes one of the existing elements.. we need to be able to pop the value.

This should remove the value from the Disaggregation but keep it in the DB?

> Do not allow editing of disaggregations or disaggregation values if the disaggregations are already associated with any indicator.

if any `AppliedIndicator`s are referencing this `Disaggregation` it should not be editable.

robertavram: @czue because of this:
    if any AppliedIndicators are referencing this Disaggregation it should not be editable.
we don't need to keep value in the DB

czue:
ok, @robertavram I think I got all your feedback and added tests for each of the conditions you mentioned

robertavram:
@czue One test failed. otherwise looks good.

czue:
@robertavram fix tests. :white_check_mark:

PR #922, opened 2017-10-04
Auditor portal updates

roman-karpovich:
Bugfixes & improvements for reported issues.


roman-karpovich:
@ntrncic 0007 migration [already exists](https://github.com/unicef/etools/blob/develop/EquiTrack/audit/migrations/0007_auto_20170807_1158.py). I've just changed text in choice for action points (recieve -> receive) and updated migration instead of creating new one.

EquiTrack/audit/tests/factories.py
```diff
@@ -16,6 +18,14 @@ def fuzz(self):
         return bool(random.getrandbits(1))
 
 
+class AgreementFactory(OriginalAgreementFactory):
```
ntrncic: maybe rename to AuditorAgreementFactory as it's confusing with the original factory name.

EquiTrack/partners/urls_v2.py
```diff
@@ -74,9 +74,6 @@
     url(r'^partners/(?P<partner_pk>\d+)/staff-members/$',
         view=PartnerStaffMemberListAPIVIew.as_view(http_method_names=['get']),
         name='partner-staff-members-list'),
-    url(r'^partners/(?P<partner_pk>\d+)/authorized-officers/$',
```
ntrncic: why is this being removed?

ntrncic:
ok cool for migrations. we have to roll back to 0005 and redo migrations when we deploy.

EquiTrack/partners/urls_v2.py
```diff
@@ -74,9 +74,6 @@
...
-        view=PartnerAuthorizedOfficersListAPIVIew.as_view(http_method_names=['get']),
-        name='partner-authorized-officers-list'),
     # url(r'^staff-members/$', view=PartnerStaffMemberListAPIVIew.as_view(), name='staff-member-list'),
```
ewheeler: please remove unused code

EquiTrack/audit/serializers/engagement.py
```diff
@@ -62,6 +63,18 @@ class Meta(WritableNestedSerializerMixin.Meta):
             'id', 'description', 'due_date', 'person_responsible', 'comments',
         ]
 
+    def validate(self, attrs):
+        if not self.instance and attrs.get('description') == _('Escalate to Investigation')\
+                and 'person_responsible' not in attrs:
+            email = 'integrity1@un.org'
+            attrs['person_responsible'] = User.objects.get_or_create(username=email, defaults={
```
ewheeler: would it make more sense to create this user with a fixture? 

then for this code, we could determine this user based on something like `email_for_user_responsible_for_investigation_escalations` in settings that would hold the email for the user created by the fixture.

user only needs to be created once and if done via fixtures then this can be altered without code changes (we just create and load a new fixture with different user details and change the email in settings file)

EquiTrack/partners/urls_v2.py
pitsevich: This endpoint is no longer used. 
We've been asked to use `partners/(?P<partner_pk>\d+)/staff-members/` instead to generate list of authorized officers, as no additional filtering is needed.

FYI @roman-karpovich  @maximlomakin

roman-karpovich:
> we have to roll back to 0005 and redo migrations when we deploy.

no need to roll back now. 0006 changed moved to 0008 with proper data migration.

EquiTrack/partners/urls_v2.py
roman-karpovich: please specify what do you mean here. we removed own code. should we clean commented urls below?

ewheeler: oops, sorry my mistake. for some reason i thought that this pr was adding the commented urls

roman-karpovich:
@ntrncic @ewheeler branch was updated according to requested changes.

roman-karpovich:
deployment additionals
`python EquiTrack/manage.py tenant_loaddata audit_risks_blueprints`

PR #925, opened 2017-10-05
Redesign fixes

ntrncic:


EquiTrack/funds/tests/test_views.py
```diff
@@ -112,15 +112,12 @@ def test_get_fail_with_one_bad_value(self):
                          'One or more of the FRs selected is either expired, has been used by another '
                          'intervention or could not be found in eTools')
 
-    def test_get_fail_with_expired_fr(self):
+    def test_get_success_with_expired_fr(self):
```
robertavram: @ntrncic  i think there is another validation that needs to be scrapped... give me 1 min

robertavram: funds.views.FRsView... that's where the validation is executed... line 30... end_date_gte==today

ntrncic: yes I changed that line and tested adding an expired FR.

robertavram: blind today...

PR #928, opened 2017-10-05
Add InterventionReportingPeriod API

vkurup:
Some questions:
- I put this model in `EDITABLE_FIELDS` in partners.tests.test_api_interventions. Is that correct?
- I'm validating that start < end < due, not allowing them to be equal. Is that correct?
- I'm not allowing the intervention to be changed after an instance is created (just because it seemed like there would be better ways to accomplish that type of operation, like delete and re-create), but let me know if that is too strict.
- Did I put the URL, view and serializer in the correct files? (I saw some labelled `prp_`... maybe this goes there?)


robertavram:
@vkurup , there is a file in assests.partners called "intervention_permissions.csv"
That file dictates what is required for which status... and what is editable for each status...

see if you can figure out how to add this field ("reporting_periods") so that it's editable in Draft by Partnership Manager, not required in Draft by UNICEF User.

vkurup:
@robertavram does this look right?

EquiTrack/partners/migrations/0054_interventionreportingperiod.py
```diff
@@ -0,0 +1,33 @@
...
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created', model_utils.fields.AutoCreatedField(default=django.utils.timezone.now, editable=False, verbose_name='created')),
+                ('modified', model_utils.fields.AutoLastModifiedField(default=django.utils.timezone.now, editable=False, verbose_name='modified')),
+                ('start_date', models.DateField(verbose_name='Reporting Period Start Date')),
+                ('end_date', models.DateField(verbose_name='Reporting Period End Date')),
+                ('due_date', models.DateField(verbose_name='Report Due Date')),
+                ('intervention', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='reporting_periods', to='partners.Intervention')),
+            ],
+            options={
+                'ordering': ['-created'],
```
robertavram: can we set ordering by due_date?

EquiTrack/partners/serializers/interventions_v2.py
```diff
@@ -283,6 +284,38 @@ def update(self, instance, validated_data):
...
+
+    def validate_intervention(self, value):
+        """
+        Changing the intervention is not allowed. Users should delete this
+        reporting period and create a new one associated with the desired
+        intervention.
+        """
+        if self.instance and value != self.instance.intervention:
+            raise ValidationError(
+                'Cannot change the intervention that this reporting period is associated with.')
```
robertavram: 👍

EquiTrack/partners/serializers/interventions_v2.py
```diff
@@ -283,6 +284,38 @@ def update(self, instance, validated_data):
...
+    def validate(self, data):
+        """
+        Validate that start_date < end_date < due_date.
+        """
+        # If we're creating, we'll have all these values in ``data``. If we're
+        # patching, we might not, so get missing values from the existing DB instance
+        start_date = data.get('start_date') or self.instance.start_date
+        end_date = data.get('end_date') or self.instance.end_date
+        due_date = data.get('due_date') or self.instance.due_date
+        if start_date >= end_date:
```
robertavram: it theoretically can be the same day if the reporting period is one day. We should probably remove the "=" on both end_date and due_date validation

EquiTrack/partners/serializers/interventions_v2.py
```diff
@@ -283,6 +284,38 @@ def update(self, instance, validated_data):
...
+            raise ValidationError('end_date must be after start_date')
+        if end_date >= due_date:
+            raise ValidationError('due_date must be after end_date')
+        return data
+
```
robertavram: can we also make sure that for this intervention there are no overlapping reporting periods. Maybe add another validation that checks all the reporting periods for this intervention and makes sure none are overlapping. (end_dates can be equal to start_dates never before the start_dates)

robertavram:
Some questions:
- I'm validating that start < end < due, not allowing them to be equal. Is that correct?
> due could be equal to end

- I'm not allowing the intervention to be changed after an instance is created (just because it seemed like there would be better ways to accomplish that type of operation, like delete and re-create), but let me know if that is too strict.
> that makes sense

- Did I put the URL, view and serializer in the correct files? (I saw some labelled `prp_`... maybe this goes there?)
> yep, you used the correct files.

👍

vkurup:
Thanks for the feedback, @robertavram. This is updated now.

PR #935, opened 2017-10-11
Fix two broken imports that kill all tests (prp-refactoring)

caktus-philip:
Two import problems were killing all tests. One was a missing import (`CountriesViewSet `) in the top-level `urls.py`. The other was a bad import in `partners/views/prp_v1.py`. (0b616e1ff84fa5fef6df32dc17c14fd291a598e9)

I also fixed 3 tests that were failing due to the upstream view changing. (2b3ebad8fa41d5b14fee083ca38ab4a9b2146cb7 and f6b710d4a042573c49cf146ac7e38f8c5a94a957). 

Last but not least, a test that counted queries was failing; I increased the number of expected queries from 17 to 18. (12e9689c5da068cca82f37cc0de996b3249c4733 ) I don't know if this is correct but I want to get this merged so the branch can get un-broken.


czue:
:clap:

PR #895, opened 2017-09-21
Fix three spelling typos in partners (PRP-refactor)

caktus-philip:
In `partners/permissions.py`, changed this --
`PartneshipManagerPermission`
to this --
`PartnershipManagerPermission`   (<== added 'r')

And this --
`PartneshipManagerRepPermission`
to this --
`PartnershipManagerRepPermission`    (<== added 'r')

And in `partners/models.py`, changed this --
`ASSESMENT_TYPES`
to this --
`ASSESSMENT_TYPES`    (<== added 'S')

Naturally this affected a lot of other files...

Sorry, I know this is annoying but I figure it's best to correct these before they spread even further.

PR #937, opened 2017-10-11
CSV flat exports on a number of models

reinbach:
Fixes https://github.com/unicef/etools-issues/issues/799


EquiTrack/partners/exports_v2.py
```diff
@@ -147,6 +221,140 @@ class InterventionCvsRenderer(r.CSVRenderer):
...
+        "country_programme": "Country Programme",
+        "result_type": "Result Type",
+        "sector": "Section",
+        "name": "Name",
+        "code": "Code",
+        "from_date": "From Date",
+        "to_date": "To Date",
+        "parent": "Parent",
+        "humanitarian_tag": "Humanitarian Tag",
+        "wbs": "WSB",
```
ewheeler: typo- label should be `WBS`

EquiTrack/partners/exports_v2.py
```diff
@@ -147,6 +221,140 @@ class InterventionCvsRenderer(r.CSVRenderer):
...
+        "vision_id": "Vision Id",
```
ewheeler: VISION should be all caps (VISION is UNICEF's WINGS -- iirc WFP uses WINGS to refer to their SAP system)

EquiTrack/funds/renderers.py
```diff
@@ -0,0 +1,325 @@
...
+        "actual_amt",
+        "outstanding_amt",
+        "start_date",
+        "end_date",
+    ]
+
+    labels = {
+        "intervention": "Reference Number",
+        "vendor_code": "Vendor Code",
+        "fr_number": "Number",
```
ewheeler: i suspect folks will want this labelled `FR Number`

EquiTrack/partners/exports_flat.py
```diff
@@ -0,0 +1,423 @@
...
+        "agreement_type": 'Agreement Type',
+        "start": 'Start Date',
+        "end": 'End Date',
+        "partner_manager_name": 'Signed By Partner',
+        "signed_by_partner_date": 'Signed By Partner Date',
+        "signed_by_name": 'Signed By UNICEF',
+        "signed_by_unicef_date": 'Signed By UNICEF Date',
+        "staff_members": 'Partner Authorized Officer',
+        "amendments": 'Amendments',
+        "country_programme_name": "Country Programme Name",
```
ewheeler: label can just be "Country Programme"

EquiTrack/partners/exports_flat.py
```diff
@@ -0,0 +1,423 @@
...
+        "country_programme": "Country Programme",
+        "result_type": "Result Type",
+        "sector": "Section",
+        "name": "Name",
+        "code": "Code",
+        "from_date": "From Date",
+        "to_date": "To Date",
+        "parent": "Parent",
+        "humanitarian_tag": "Humanitarian Tag",
+        "wbs": "WSB",
```
ewheeler: WBS

EquiTrack/partners/exports_flat.py
```diff
@@ -0,0 +1,423 @@
...
+        "vision_id": "Vision Id",
```
ewheeler: VISION

EquiTrack/reports/renderers.py
```diff
@@ -0,0 +1,117 @@
...
+        "description",
+        "code",
+        "subdomain",
+        "disaggregatable",
+        "disaggregation_logic",
+    ]
+
+    labels = {
+        "intervention": "Reference Number",
+        "lower_result": "Lower Result",
```
ewheeler: i'm reasonably confident that this should read "Lower Level Result"

EquiTrack/reports/renderers.py
```diff
@@ -0,0 +1,117 @@
...
+        "code",
+        "subdomain",
+        "disaggregatable",
+        "disaggregation_logic",
+    ]
+
+    labels = {
+        "id": "Id",
+        "intervention": "Reference Number",
+        "lower_result": "Lower Result",
```
ewheeler: 'Lower Level Result' i think

EquiTrack/reports/renderers.py
```diff
@@ -0,0 +1,117 @@
...
+        "assumptions": "Assumptions",
+        "means_of_verification": "Means of Verification",
+        "total": "Total",
+        "name": "Name",
+        "unit": "Unit",
+        "description": "Description",
+        "code": "Code",
+        "subdomain": "Subdomain",
+        "disaggregatable": "Disaggregatable",
+        "disaggregation_logic": "Logic",
```
ewheeler: probably will make more sense to folks as "Disaggregation Logic"

EquiTrack/reports/views/v2.py
```diff
@@ -7,12 +7,38 @@
...
+    LowerResultSerializer,
+    MinimalOutputListSerializer,
+    OutputListSerializer,
+)
 from reports.serializers.v1 import IndicatorSerializer
+from partners.filters import PartnerScopeFilter
 from partners.models import Intervention
-from partners.permissions import PartneshipManagerRepPermission
+from partners.permissions import (
+    PartneshipManagerPermission,
```
ewheeler: FYI this spelling mistake has been fixed by @caktus-philip in https://github.com/unicef/etools/tree/prp-refactoring (via https://github.com/unicef/etools/pull/895)

EquiTrack/partners/exports_v2.py
reinbach: Ah good old WINGS... grrrrr

EquiTrack/funds/tests/test_api.py
```diff
@@ -3,16 +3,547 @@
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
+
+from rest_framework import status
+from tablib.core import Dataset
 from unittest import TestCase
```
caktus-philip: `from unittest import TestCase` should go up with the other Python imports (`from __future__ ...`)

EquiTrack/funds/renderers.py
```diff
@@ -0,0 +1,325 @@
+from rest_framework_csv import renderers as r
+
+
+class FundsReservationHeaderCsvRenderer(r.CSVRenderer):
```
caktus-philip: Per PEP8, "When using abbreviations in CapWords, capitalize all the letters of the abbreviation. Thus HTTPServerError is better than HttpServerError." So this should be `FundsReservationHeaderCSVRenderer`. Sorry, I know it's  a lot to change since this occurs in many places.

EquiTrack/funds/tests/test_api.py
```diff
@@ -3,16 +3,547 @@
...
         names_and_paths = (
+            ('funds-donor', 'donor/', {}),
             ('frs', 'frs/', {}),
-            )
+            ('funds-commitment-header', 'commitment-header/', {}),
+            ('funds-commitment-item', 'commitment-item/', {}),
+            ('funds-reservation-header', 'reservation-header/', {}),
+            ('funds-reservation-item', 'reservation-item/', {}),
+            ('funds-grant', 'grant/', {}),
+        )
```
caktus-philip: Thanks for keeping this up-to-date!

EquiTrack/funds/tests/test_api.py
```diff
@@ -3,16 +3,547 @@
...
+class TestFundsReservationHeaderExportList(APITenantTestCase):
+    def setUp(self):
+        super(TestFundsReservationHeaderExportList, self).setUp()
+        self.unicef_staff = UserFactory(is_staff=True)
+        self.frs = FundsReservationHeaderFactory()
+
+    def test_invalid_format_export_api(self):
+        response = self.forced_auth_req(
+            'get',
+            '/api/v2/funds/reservation-header/',
```
caktus-philip: I would use Django's `reverse()` to construct this URL, and perhaps do it in `setUp()`. Using the URL alias makes the test more robust in case `urls.py` changes, and it makes it easier to find references to URLs (if, say, you want to change or delete an endpoint).

EquiTrack/partners/tests/test_export_agreement.py
```diff
@@ -0,0 +1,259 @@
...
+            core_values_assessment_date=datetime.date.today(),
+            total_ct_cp=10000,
+            total_ct_cy=20000,
+            deleted_flag=False,
+            blocked=False,
+            type_of_assessment="Type of Assessment",
+            last_assessment_date=datetime.date.today(),
+        )
+        partnerstaff = PartnerStaffFactory(partner=partner)
+        with tempfile.NamedTemporaryFile(
```
caktus-philip: Thanks for updating this to Django 1.10-friendly code. As I look at the code now, I wonder if the tempfile is even needed. IIUC, it's deleted as soon as the `with` statement exits which is pretty much immediately. The existence of the file can't mean much to the test if that's the case. Could you just as well use `attached_agreement='my_fake_attachment.pdf'` in place of `attached_agreement=attachment.name` and skip the whole tempfile? Just a suggestion.

caktus-philip:
I added one more comment regarding the tempfile; other than that this looks OK to me. 👍

EquiTrack/partners/tests/test_export_agreement.py
reinbach: I believe with the `delete=False` parameter the file sticks around when it closes after the `with` statement exits. But removing the `tempfile` code, the tests still pass, so I agree, it is better to just remove that part, and make things simpler.

EquiTrack/EquiTrack/factories.py
```diff
@@ -173,6 +173,32 @@ class Meta:
     country_programme = factory.SubFactory(CountryProgrammeFactory)
 
 
+class AssessmentFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = partner_models.Assessment
+
+    partner = factory.SubFactory(PartnerFactory)
+    type = u"Micro Assessment"
```
ntrncic: can do a factory for AssessmentType

EquiTrack/funds/renderers.py
```diff
@@ -0,0 +1,325 @@
+from rest_framework_csv import renderers as r
```
ntrncic: If we have a renderer for each model and a field changes or gets added we will have to remember to add to the renderer. Can we just make a generic one that takes all fields?

EquiTrack/funds/tests/test_api.py
```diff
@@ -3,16 +3,547 @@
...
+            "Document Text",
+            "Amount",
+            "Total Amount",
+            "Actual Amount",
+            "Outstanding Amount",
+            "Start Date",
+            "End Date",
+        ])
+        self.assertEqual(dataset[0], (
+            u"{}".format(self.frs.intervention.pk),
```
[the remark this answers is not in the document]
ntrncic: could we test for all fields in the model but not explicitly naming them? Like the number of columns equals number of fields.

ntrncic:
This is a lot of code in our codebase just for exports. Maybe use something more generic or we move exports into it's on service if the business needs fully customizable exports.

reinbach:
A generic serializer would be nice, a few caveats though are;

1. that we do have certain fields where we don't actually use the default value, but change the source;

eg: Based on the format we make use of a different field for the object instead of it's primary key. This field does not necessarily match the unicode(object) value. Maybe the unicode(object) should be set, if missing or corrected

2. In other instances we do have some custom serializer fields, but these could be applied across the board for specific field types maybe?

eg: Hact values are a jsonfield, those get custom rendered, maybe I did that incorrectly?

3. For certain models, we pull in reverse foreign keys

So we do have a number of exceptions to the norm, and would probably have to inherit/abstract from the generic serializer for these instances. I'm happy to work on this, just wanted to raise these thoughts.

EquiTrack/funds/renderers.py
reinbach: The renderers do handle the ordering of the fields, and the heading labels.

So if we're not concerned about ordering, that makes things simpler.

For heading labels, we could make use of a mapping (may get ugly), or we ensure that all fields have a verbose_name label and attempt to use that?

reinbach:
For testing just the count of headers and fields.

There are some fields that are rendered/serialized differently, that we may want to ensure happen correctly. Maybe have unique tests for that serialization. But for most of these we test just the header/field count?

reinbach:
I've added a more generic way of handling the renderer header/labels, and applied it to a couple of the exports to see what the affect is.

If you want to take quick look at that and see if there are any suggestions on it, before I go ahead and apply it to the rest of the exports, that would be great.

See commit https://github.com/unicef/etools/pull/937/commits/af96aa42cd3fd3d976214c0e3c2e25f107d8895f

reinbach:
Please note that there is no longer any ordering if a custom renderer is not provided.
The ordering of fields is consistent between requests, so I think that this is ok then?

ntrncic:
@reinbach please remove conflicts

reinbach:
@ntrncic conflicts have been removed

EquiTrack/partners/exports_v2.py
```diff
@@ -60,7 +60,29 @@ class PartnerOrganizationHactCsvRenderer(r.CSVRenderer):
     }
 
 
-class AgreementCvsRenderer(r.CSVRenderer):
+class PartnerStaffMemberCSVRenderer(r.CSVRenderer):
```
ntrncic: I don't think we need a custom one for this model

EquiTrack/partners/migrations/0057_merge.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.10 on 2017-10-25 18:13
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
ntrncic: why do we need this?

EquiTrack/partners/models.py
```diff
@@ -234,35 +234,45 @@ class PartnerOrganization(AdminURLMixin, models.Model):
...
+        verbose_name=_('CSO Type'),
         max_length=50,
         choices=CSO_TYPES,
-        verbose_name='CSO Type',
-        blank=True, null=True
+        blank=True,
+        null=True,
     )
     name = models.CharField(
+        verbose_name=_('Full Name'),
```
ntrncic: Name

EquiTrack/partners/models.py
```diff
@@ -1152,103 +1280,150 @@ class Intervention(TimeStampedModel):
...
+    contingency_pd = models.BooleanField(
+        verbose_name=_("Contingency PD"),
+        default=False,
+    )
 
-    offices = models.ManyToManyField(Office, blank=True, related_name='office_interventions+')
-    population_focus = models.CharField(max_length=130, null=True, blank=True)
+    offices = models.ManyToManyField(
+        Office,
+        verbose_name=_("UNICEF Office"),
```
ntrncic: Office

EquiTrack/partners/models.py
```diff
@@ -1512,17 +1687,33 @@ class InterventionAmendment(TimeStampedModel):
...
+    signed_date = models.DateField(
+        verbose_name=_("Signed Date"),
+        null=True,
+    )
+    amendment_number = models.IntegerField(
+        verbose_name=_("Number"),
+        default=0,
+    )
     signed_amendment = models.FileField(
+        verbose_name=_("Amendment"),
```
ntrncic: Amendment Document

EquiTrack/reports/models.py
```diff
@@ -278,12 +347,17 @@ class LowerResult(TimeStampedModel):
     # Lower result is always an output
 
     # link to intermediary model to intervention and cp ouptut
-    result_link = models.ForeignKey('partners.InterventionResultLink', related_name='ll_results', null=True)
+    result_link = models.ForeignKey(
+        'partners.InterventionResultLink',
+        verbose_name=_("Reference Number"),
```
ntrncic: Hmmm this is an intermediary model. it shouldn't be included in any exports. therefore no verbose name

EquiTrack/reports/models.py
```diff
@@ -358,24 +456,54 @@ class AppliedIndicator(models.Model):
     indicator = models.ForeignKey(IndicatorBlueprint)
 
     # the result this indicator is contributing to.
-    lower_result = models.ForeignKey(LowerResult, related_name='applied_indicators')
+    lower_result = models.ForeignKey(
+        LowerResult,
+        verbose_name=_("Lower Level Result"),
```
ntrncic: PD Result

EquiTrack/partners/migrations/0057_merge.py
reinbach: Divergent migrations happened at 0054, this merges the 2 migration "threads"

reinbach:
Woot!

ntrncic:
Could you please remove any merge migrations. We would like to keep the migrations linear.

PR #940, opened 2017-10-12
eTools not compatible with Postgres 9.6

reinbach:
fixes https://github.com/unicef/etools-issues/issues/569

In order to test against both 9.5 and 9.6 instances in CI, it looks like we will need to upgrade to circleci v2, and make use of workflows (https://circleci.com/docs/2.0/workflows/). I'll leave that to be a separate ticket.

etools-infra has not been updated to use postgresql 9.6. if that should be part of this ticket, let me know.


vkurup:
Woo hoo! I'll finally be able to shut down my extra PG server on my laptop :)

I see a couple warnings when running:
```
WARNINGS:
?: (tenant_schemas.W003) Your default storage engine is not tenant aware.
	HINT: Set settings.DEFAULT_FILE_STORAGE to 'tenant_schemas.storage.TenantFileSystemStorage'
django.conf.settings: (tenant_schemas.W001) You should put 'tenant_schemas' before any django core applications in INSTALLED_APPS.
	HINT: This is necessary to overwrite built-in django management commands with their schema-aware implementations.
```

But I don't know enough to know whether they are important to fix, or not.

If not, then a hearty :+1: from me!

reinbach:
I'm not sure if we want tenant schemas applied to all the django core apps. As we currently have;

`INSTALLED_APPS = SHARED_APPS + TENANT_APPS + ('tenant_schemas',)`

If someone with more knowledge about etools could let me know, I can make the change. Or maybe at the very least change it to;

`INSTALLED_APPS = SHARED_APPS + ('tenant_schemas',) + TENANT_APPS`

but we will continue to have this warning, with the above change.

I'm also not clued in enough with etools to make a call on the `DEFAULT_FILE_STORAGE` setting. If I understand correctly, there is some development/changes around file storage in the dev pipeline?

robertavram:
didn't mean to close it:)

reinbach:
Looks like we can ignore the storage warning. In order to make media files tenant aware we could make use of the tenant storage, but we already make use of `get*_file_path` functions in the partner models to handle this separation between tenants.

vkurup:
FWIW, :+1: from me on this PR.

domdinicola:
it looks good to me as well

PR #942, opened 2017-10-12
add migration command to update Palestine agreeements

csabadenes:


csabadenes:
connects unicef/etools-issues#818

PR #951, opened 2017-10-12
Add test for list & create of InterventionLowerResult API endpoint (prp-refactor)

caktus-philip:


EquiTrack/partners/tests/test_api_interventions.py
```diff
@@ -614,3 +615,145 @@ def test_group_permission_non_staff(self):
         # Now the request should succeed.
         response = self._make_request(user)
         self.assertResponseFundamentals(response)
+
+
+class TestAPIInterventionLowerResultListView(APITenantTestCase):
+    '''Exercise the list view for InterventionLowerResultListCreateView'''
+    def setUp(self):
```
czue: nit picky, but might be marginally more efficient to put this in a `setUpClass` so it only runs once, since it looks like all tests depend on it and you're not mutating data or doing anything special in `tearDown`

EquiTrack/partners/tests/test_api_interventions.py
```diff
@@ -614,3 +615,145 @@ def test_group_permission_non_staff(self):
...
+        _add_user_to_partnership_manager_group(user)
+
+        # Now the request should succeed.
+        response = self._make_request(user)
+        self.assertResponseFundamentals(response)
+
+
+class TestAPIInterventionLowerResultCreateView(APITenantTestCase):
+    '''Exercise the create view for InterventionLowerResultListCreateView'''
+    def setUp(self):
```
czue: same comment here

EquiTrack/partners/tests/test_api_interventions.py
```diff
@@ -614,3 +615,145 @@ def test_group_permission_non_staff(self):
...
+
+    def test_code_read_only(self):
+        '''Ensure lower_result.code can't be written'''
+        user = UserFactory()
+        _add_user_to_partnership_manager_group(user)
+        data = self.data.copy()
+        data['code'] = 'ZZZ'
+
+        response = self.forced_auth_req('post', self.url, user=user, data=data)
+        response_json = self.assertResponseFundamentals(response)
```
czue: not really a concern with this PR but more just a question on my side - is it the case that DRF just silently swallows invalid params on all create / update requests?

caktus-philip: Great question! Some of DRF's behavior surprises me. I expected this request to give a 400 when I first tried it.

The exact answer depends on the definition of "invalid". (I had to look at DRF's serializer code to figure that out. It was educational!) The reason that DRF silently ignores this attempt to set `code` is because that field is defined as read-only on the serializer. If a serializer gets incoming data, it knows it's handling a write request (create/update), so it only concerns itself with writeable fields. Data associated with non-writeable fields is dropped. That's here --
https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L455

That means you can also cram irrelevant fields into the data dict and populate them with lorem ipsum and DRF won't bat an eyelash. However, if you write bad data to a writeable field (e.g. 'lorem ipsum' where it's expecting a PK), then validation will fail, ValidationError will be raised, and you'll get a 400.

caktus-philip:
@czue I think Rob specifically said that he'd like us to approve one another's PRs if we feel OK about reviewing the Python. (I definitely don't feel comfortable reviewing anything that involves business rules.) PRs that have already been approved are much easier for Rob and Nik to review. So, if you feel good about this, please approve.

EquiTrack/partners/tests/test_api_interventions.py
czue: nice, thanks for the comprehensive explanation!

caktus-philip: yw. As I said, this concept was fuzzy for me and it was nice to get it straight. I hope that typing it out will help me remember it.

PR #960, opened 2017-10-13
Add tests for intervention indicators list & create API views

caktus-philip:


EquiTrack/partners/tests/test_api_interventions.py
```diff
@@ -763,3 +765,176 @@ def test_code_read_only(self):
...
+        if expected_keys:
+            for d in response_json:
+                self.assertEqual(sorted(d.keys()), expected_keys)
+
+        actual_ids = sorted([d.get('id') for d in response_json])
+        expected_ids = sorted((self.indicator1.id, self.indicator2.id))
+
+        self.assertEqual(actual_ids, expected_ids)
+
+    def test_no_permission_user_forbidden(self):
```
czue: I'm noticing that a lot of these test cases have a very similar structure and involve very similar tests... e.g. ensuring unauthenticated users get 403s and such. I haven't thought too hard about this, but wondering if it would make sense to create a `MixIn` (akin to some of DRF's `CreateUpdateListView` type of stuff that is something like `APIRequirePermissionTextMixIn` or such and then you would just bootstrap some variables in the test and you wouldn't have to copy/paste all these checks into every subclass? Might reduce the amount of duplication and also provide a canonical way for people to write API tests (as you're introducing by convention with the `assertResponseFundamentals` pattern)

 Haven't thought too hard about this so may be some downsides I'm not thinking of.

EquiTrack/reports/serializers/v2.py
```diff
@@ -136,7 +136,6 @@ def validate(self, attrs):
                 )
 
         elif not attrs.get('cluster_indicator_id'):
-            print "no cluster id"
```
czue: :+1:

PR #963, opened 2017-10-16
interventions-dash api returns only active interventions now

marko911:
connects unicef/etools-issues#322


marko911:
Add to gitignore next time plz @robertavram

PR #964, opened 2017-10-17
add admin level to location gatewaytype

csabadenes:


EquiTrack/partners/views/v2.py
```diff
@@ -169,6 +170,7 @@ def get(self, request):
         intervention_doc_type = choices_to_json_ready(Intervention.INTERVENTION_TYPES)
         intervention_status = choices_to_json_ready(Intervention.INTERVENTION_STATUS)
         intervention_amendment_types = choices_to_json_ready(InterventionAmendment.AMENDMENT_TYPES)
+        location_types = GatewayType.objects.values('id', 'name', 'admin_level').order_by('id')
```
robertavram: is this the format they need on the frontend? everything else in this endpoint is label,value..  just making sure...

csabadenes: I think so, yes. I see that this format does not fit with the others, but the alternative would be to make another endpoint, which would be additional effort for the webserver, but maybe it is ok.. please let me know which variant do you prefer.

PR #972, opened 2017-10-24
Tasks transactional

ntrncic:


EquiTrack/partners/tasks.py
```diff
@@ -48,6 +48,7 @@ def agreement_status_automatic_transition():
         _make_agreement_status_automatic_transitions(country.name)
 
 
+@transaction.atomic
```
robertavram: not the whole function.. /record

EquiTrack/partners/tasks.py
```diff
@@ -97,6 +98,7 @@ def intervention_status_automatic_transition():
         _make_intervention_status_automatic_transitions(country.name)
 
 
+@transaction.atomic
```
robertavram: same here.. only atomic around the record...

PR #736, opened 2017-06-22
Add Flake8 testing to CI

vkurup:
This just adds the ability for flake8 to run on CircleCI (and locally). I expect this build to fail and then I'll update .flake8 to make sure it passes.

I've removed our exclusion of E121 and E126 from the guidelines because apparently there is currently no code violating those rules. Here is the current list of violations (by code type):

```
$ flake8 --statistics --count -qq
47    E111 indentation is not a multiple of four
10    E122 continuation line missing indentation or outdented
39    E127 continuation line over-indented for visual indent
35    E128 continuation line under-indented for visual indent
7     E201 whitespace after '{'
5     E202 whitespace before '}'
1     E203 whitespace before ':'
1     E225 missing whitespace around operator
15    E231 missing whitespace after ':'
9     E261 at least two spaces before inline comment
3     E262 inline comment should start with '# '
31    E265 block comment should start with '# '
42    E266 too many leading '#' for block comment
4     E301 expected 1 blank line, found 0
44    E302 expected 2 blank lines, found 1
20    E303 too many blank lines (3)
5     E305 expected 2 blank lines after class or function definition, found 1
6     E306 expected 1 blank line before a nested definition, found 0
20    E402 module level import not at top of file
557   E501 line too long (154 > 120 characters)
1     E703 statement ends with a semicolon
1     E712 comparison to False should be 'if cond is False:' or 'if not cond:'
1     E731 do not assign a lambda expression, use a def
108   F401 '.celery.app as celery_app' imported but unused
4     F403 'from local_base import *' used; unable to detect undefined names
35    F405 'INSTALLED_APPS' may be undefined, or defined from star imports: local_base
5     F811 redefinition of unused 'WorkplanProject' from line 6
5     F821 undefined name 'PCALocation'
29    F841 local variable 'exp' is assigned to but never used
7     N802 function name should be lowercase
1     N803 argument name should be lowercase
13    N806 variable in function should be lowercase
1     W291 trailing whitespace
12    W292 no newline at end of file
3     W293 blank line contains whitespace
12    W391 blank line at end of file
1139
```

PR #976, opened 2017-10-25
added setup.cfg

domdinicola:
maybe we can include .flake8 content here as well

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,15 @@
+[isort]
+combine_as_imports = true
+default_section = THIRDPARTY
+include_trailing_comma = true
+known_first_party = EquiTrack, assets, attachments, audit, environment, firms, funds, locations, management, monitoring, notification, partners, publics, reports, supplies, t2f, tpm, users, utils, vision, workplan
+multi_line_output = 0
+line_length = 80
+balanced_wrapping = true
+order_by_type = false
+not_skip = __init__.py
+
+[pep8]
+ignore = E501,E401,W391,E128,E261,E731
+max-line-length = 160
+exclude = migrations,.git,docs,diff_match_patch.py
--- a/EquiTrack/setup.cfg
+++ b/EquiTrack/setup.cfg
@@ -2,6 +2,8 @@
 combine_as_imports = true
 default_section = THIRDPARTY
 include_trailing_comma = true
+known_django = django
+sections = FUTURE,STDLIB,DJANGO,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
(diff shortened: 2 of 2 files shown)


caktus-philip:
@domdinicola I think flake8 settings have to be in the EquiTrack directory otherwise they won't be copied into Docker. See #736. @vkurup might be able to comment more.

setup.cfg
```diff
@@ -0,0 +1,15 @@
...
+include_trailing_comma = true
+known_first_party = EquiTrack, assets, attachments, audit, environment, firms, funds, locations, management, monitoring, notification, partners, publics, reports, supplies, t2f, tpm, users, utils, vision, workplan
+multi_line_output = 0
+line_length = 80
+balanced_wrapping = true
+order_by_type = false
+not_skip = __init__.py
+
+[pep8]
+ignore = E501,E401,W391,E128,E261,E731
```
vkurup: Personally, I wouldn't be in favor of ignoring these. Note that we just spent a bit of time cleaning up the entire repo of all flake8 violations, so allowing them back in would feel a little like a step backwards. If we do agree to change this policy, we should document it here: https://github.com/unicef/etools/blob/master/docs/guidelines.rst#flake8-exceptions

setup.cfg
```diff
@@ -0,0 +1,15 @@
...
+max-line-length = 160
```
vkurup: My main argument for keeping it at 120 or lower is that longer lines often go past the right edge of the Github display box, so tend not to get reviewed as well.

vkurup:
Never heard of `isort` before, btw. Sounds interesting!

reinbach:
Just to let everyone know my preference, cause I know you're all dying to know what it is;

The closer we are to `pep8` standard the happier I am. So the less `ignore`s we have the better.
I'm old school, I still prefer line lengths being 80, but I know I'm not going to get any other takers on that one :)

domdinicola:
Looks like isort follows (or can be configured to follow) what Organizing imports section of the guideline says.

setup.cfg
domdinicola: I was not aware of the doc!
Looks really useful in order to be aligned.

caktus-philip:
@domdinicola Thanks for `isort`. Sorting imports statements is something I do begrudgingly if at all. I'd be glad not to have to do it anymore. I think organizing imports in the Python/django/3rd party/projects sections provides more benefit, so if it can do that too, that's great.

domdinicola:
@caktus-philip latest commit should do what you were asking for

PR #982, opened 2017-10-25
partners: improvements 

domdinicola:


EquiTrack/partners/tests/test_models.py
```diff
@@ -722,6 +722,17 @@ def test_total_unicef_cash(self):
         )
         self.assertEqual(int(self.intervention.total_unicef_cash), 100000)
 
+    def test_total_partner_contribution(self):
+        InterventionBudget.objects.create(
```
reinbach: A suggestion would be to change `InterventionBudget.objects.create` to `InterventionBudgetFactory`, in case we should modify `InterventionBudget` to have new not null field, then we only need to update the factory and not all the tests.

domdinicola:
nice idea @reinbach

PR #983, opened 2017-10-26
Update snapshot functionality

reinbach:
Wanting to get feedback to make sure I am on the right path.

I decided to create a new app to handle the functionality, this allows us to remove `django-activity-stream` which is one of the major blockers to upgrade to Django 1.10 (see https://github.com/unicef/etools-issues/issues/246). 

I've put the snapshot functionality in place for a single `create`, and an `update` action to see the effects. 

So if this direction looks good, I'll go ahead and;
- remove `django-activity-stream`
- roll the functionality across the board to the other create/update calls
- apply this functionality in the Django admin section


EquiTrack/partners/tests/test_api_interventions.py
```diff
@@ -247,6 +248,7 @@ def test_fr_details_is_accurate_on_creation(self):
                          float(sum([self.fr_1.intervention_amt, self.fr_2.intervention_amt])))
 
     def test_add_two_valid_frs_on_update_pd(self):
+        self.assertEqual(Activity.objects.all().count(), 0)
```
vkurup: optional nitpick. All of the `.all().count()` calls can be simplified to `.count()`.

vkurup: ... or in this case, even `.exists()`

EquiTrack/partners/tests/test_api_interventions.py
```diff
@@ -263,6 +265,18 @@ def test_add_two_valid_frs_on_update_pd(self):
                          float(sum([self.fr_1.total_amt, self.fr_2.total_amt])))
         self.assertEqual(response['frs_details']['total_intervention_amt'],
                          float(sum([self.fr_1.intervention_amt, self.fr_2.intervention_amt])))
+        self.assertEqual(Activity.objects.all().count(), 1)
+        activity = Activity.objects.all()[0]
```
vkurup: `.all()[0]` can be simplified to `.first()`

EquiTrack/snapshot/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    else:
+        change = ""
+
+    return change
+
+
+def create_snapshot(target, target_before, by_user, data):
+    """If target_before is None, then action is create, otherwise update
+
+    For to many relation fields add them to the target
```
vkurup: s/For/For many/

EquiTrack/EquiTrack/serializers.py
```diff
@@ -10,3 +12,21 @@ def to_representation(self, value):
...
         return json.dumps(data) if isinstance(data, dict) else data
+
+
+class SnapshotModelSerializer(serializers.ModelSerializer):
+    def save(self, **kwargs):
+        data = dict(
+            list(self.validated_data.items()) +
+            list(kwargs.items())
+        )
+        pre_save = self.instance
```
caktus-philip: It seems to me that you'd need to mess around with `deepcopy()` here. `pre_save` is just a reference to `self.instance`, so if `self.instance.stuff` is e.g. a list of `[1, 2, 3]` that becomes `[1, 2, 3, 4]` as a result of the `save()`, then `pre_save.stuff` will also be `[1, 2, 3, 4]` after the save and the change won't be noted. 

Another approach might be to set `pre_save` to a dictified version of `self.instance` which would kill two birds with one stone -- making copies, but doing it more cheaply than `deepcopy()` would because you're not copying all the model baggage.

EquiTrack/snapshot/utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+from django.forms import model_to_dict
+from rest_framework.utils import model_meta
+
+from snapshot.models import Activity
+
+
+def jsonify(data):
+    for key, value in data.items():
+        if type(value) not in [int, float, bool, str, list]:
```
caktus-philip: I prefer `isinstance()` to `type()` here (as in most places), in part because it's a little more generous -- `isinstance()` returns `True` for subclasses, whereas type returns `False`. `isinstance()` also accepts an iterable so you can say this --

```
if not isinstance(value, (int, float, bool, str, list)):
```

Speaking of iterables, shouldn't `tuple` be in the list of fundamental types above? What about `dict`?

EquiTrack/snapshot/utils.py
```diff
@@ -0,0 +1,65 @@
...
+            data[key] = unicode(data[key])
+    return data
+
+
+def set_relation_values(obj, data):
+    obj_dict = jsonify(model_to_dict(obj))
+    instance_info = model_meta.get_field_info(obj.__class__)
```
caktus-philip: I'm not sure I understand this usage of `get_field_info()`. The DRF docstring says 

> Given an instance of a serializer field, return a dictionary metadata about it.

It seems to me you're passing arbitrary class instances (like an instance of `Intervention`) to something that expects something quite specific, no? The current implementation of `get_field_info()` might serve the purpose, but I'd be nervous that some future version of DRF might change behavior in a way that suits its use case but breaks this one. That risk is increased by the fact that the function isn't clearly meant to be public (has a docstring and doesn't have an underscore prefix, but is not mentioned in DRF's documentation).

EquiTrack/snapshot/utils.py
```diff
@@ -0,0 +1,65 @@
...
+        for k, v in data.items():
+            if k in previous_obj_dict and data[k] != previous_obj_dict[k]:
+                change.update({
+                    k: jsonify({
+                        "before": previous_obj_dict[k],
+                        "after": data[k],
+                    })
+                })
+    else:
+        change = ""
```
caktus-philip: Should be `change = {}`, or you could just make that the top line of the function and eliminate line 32 and the `else` clause.

EquiTrack/snapshot/utils.py
reinbach: Was using `get_field_info` to get the `*_to_many` relations. Based on your suggestion, reproduced needed functionality, and no longer rely on DRF.

reinbach:
@caktus-philip the kids are happy, I gave each of them a gold star :)

reinbach:
The functionality we wanted was very different to what was being provided by `django-activity-stream` so I don't think it is worth our effort to port their tests across.

EquiTrack/snapshot/models.py
```diff
@@ -0,0 +1,57 @@
...
+    UPDATE = "update"
+    ACTION_CHOICES = (
+        (CREATE, _("Create")),
+        (UPDATE, _("Update")),
+    )
+
+    target_content_type = models.ForeignKey(
+        ContentType,
+        related_name='activity',
+        on_delete=models.CASCADE,
```
ewheeler: if i'm reading this right, when a target record is deleted then its related activities will also be deleted.

is there a way to capture the deletion activity for posterity? 

i can only imagine dodgy ways that may not actually be possible, like having a `GhostRecord` model with a json field that we stuff with a serialized representation of the deleted record data.. and then use `on_delete=models.SET(entomb_record)` to remap the foreign key to the ghost?

EquiTrack/snapshot/utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+from django.forms import model_to_dict
+
+from snapshot.models import Activity
+
+
+def jsonify(data):
+    """Convert data into a dictionary that can be json encoded"""
+    for key, value in data.items():
+        if not isinstance(value, (int, float, bool, str, list, tuple, dict)):
```
caktus-philip: This list should also include `None` otherwise it won't be properly translated to JSON `null`. 

For completeness, the list should also include `long` and `set` although they're unlikely to be encountered in practice. 

Last but not least, `str` should be `basestring` since `isinstance(u'', str) == False)`, although the only downside of unicode strings slipping through the call to `isinstance()` is that `unicode()` will be called on them which just amounts to an extra no-op function call.

EquiTrack/snapshot/tests/test_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    def test_jsonify(self):
+        intervention = InterventionFactory()
+        j = utils.jsonify(model_to_dict(intervention))
+        self.assertEqual(j["title"], intervention.title)
+
+
+class TestGetToManyFieldNames(TenantTestCase):
+    def test_intervention(self):
+        intervention = InterventionFactory()
+        fields = utils.get_to_many_field_names(intervention.__class__)
```
caktus-philip: `get_to_many_field_names()` docstring (and param name) say it accepts an object, but this is passing a class rather than an object.

EquiTrack/snapshot/utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def create_change_dict(prev_dict, current_dict):
+    """Create a dictionary showing the differences between the
+    initial target and the target after being saved.
+
+    If prev_dict is empty, then change is empty as well
+    """
+    change = {}
+    if prev_dict:
```
caktus-philip: This line is harmless but unnecessary; if `prev_dict` is `{}` then the `for` loop below is a no-op anyway.

EquiTrack/snapshot/models.py
reinbach: The target record would be the `ContentType` record I believe. So only if the `ContentType` record is removed, will these be removed. Not when the target record is removed.

When a record is deleted that is referenced, the `target` field becomes `None`, but the `target_content_type`, and `target_object_id` values remain. I've added a test that is meant to check this.

ntrncic:
please resolve conflicts sorry!

PR #984, opened 2017-10-26
Tweak PartnerOrganizationListAPIView

reinbach:
Move create functionality to the specific create serializer.

Not sure if I'm missing anything, but I thought this change may make the code a little easier to read.

The view does a bunch of things, uses various serializers depending on the method, and having the `create` function there can be confusing as to which serializer you're dealing with. 

Moving the `create` to the serializer in this instance relieves us of having to recreate the serializer class and perform it's various tasks before the `save`.


EquiTrack/partners/views/partner_organization_v2.py
```diff
@@ -124,33 +124,6 @@ def list(self, request, format=None):
 
         return response
 
-    @transaction.atomic
-    def create(self, request, *args, **kwargs):
-        # TODO: on create we should call the insight API with the vendor number and use that information to populate:
```
caktus-philip: My unasked-for $.02 -- I would not lose this TODO unless an issue has been created to track it elsewhere. Otherwise +1 to the PR. Thanks!

reinbach: Thanks, feedback is always welcome.

reinbach:
In that case can we remove `PartnerOrganizationAddView` view as well then?

ntrncic:
@reinbach `PartnerOrganizationAddView` stays as it's a different way of adding a partner.

reinbach:
:ok_hand:

ntrncic:
since this PR touches Partner stuff please rebase branch and PR against PRP refactoring

PR #987, opened 2017-10-26
Load fixtures for shared schema (prp-refactoring)

caktus-philip:
Current behavior of `FastTenantTestCase` only loads fixtures for the tenant schema. Any shared apps (including 3rd party ones `post_office`) referenced in fixtures are not loaded. This PR alters `FastTenantTestCase` to load fixtures for both the shared schema _and_ the tenant schema. 

Essentially, this is a one-line change. I moved the fixture-loading code into a new class method called `_load_fixtures()` and added an extra call to it (before calling `connection.set_tenant(cls.tenant)`).

Also added a test case that shows this is working. If you want to test it yourself, comment out the "new" call to `cls._load_fixtures()` on line 124 of `mixins.py`. Commenting out the line emulates current behavior. Then run `python manage.py test  audit.tests.test_email`. The test will fail. Uncomment line 124, re-run the test, and it will succeed.

Loading shared schema fixtures will become important when we squash migrations. There are 5 migrations that create `EmailTemplate` instances. This change enables us to replace those migrations with a fixture and not break tests.


czue:
:+1:

PR #990, opened 2017-10-27
Make audit models str method python 2 and 3 compatible

caktus-philip:
Ensure all models in audit have Python 3-compatible `str()` methods using Django's `@python_2_unicode_compatible` decorator. Fixed a bug in `AuditorStaffMember`.

Add non-ASCII tests for `str()` and `unicode()` methods for most audit models. 

In `EquiTrack/tests/test_all_models.py`, remove `audit` from the list of modules not tested for Python 3-compatibility since it's now compliant.

### DIFF
--- a/EquiTrack/audit/tests/test_models.py
+++ b/EquiTrack/audit/tests/test_models.py
@@ -1,8 +1,24 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+import sys
+from unittest import skipIf, TestCase
+
 from django.core import mail
 
 from EquiTrack.tests.mixins import FastTenantTestCase
 
-from .factories import AuditPartnerFactory
+from .factories import (
+    AuditPartnerFactory,
+    AuditorStaffMemberFactory,
+    EngagementFactory,
+    PurchaseOrderFactory,
+    RiskBluePrintFactory,
+    RiskCategoryFactory,
+    RiskFactory,
+    SpotCheckFactory,
+    )
 from ..models import AuditorStaffMember, Auditor
 from firms.factories import UserFactory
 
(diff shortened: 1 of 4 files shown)


reinbach:
How come we're merging into `prp-refactoring`, should I be changing my pull requests to merge with that branch as well?

EquiTrack/audit/tests/test_models.py
```diff
@@ -1,8 +1,24 @@
+from __future__ import absolute_import
```
reinbach: Looks like `audit/models.py` is lacking all these `__future__` imports

EquiTrack/audit/tests/test_models.py
```diff
@@ -21,3 +37,93 @@ def test_signal(self):
         self.assertIn(Auditor.name, staff_member.user.groups.values_list('name', flat=True))
 
         self.assertEqual(len(mail.outbox), 1)
+
+
+@skipIf(sys.version_info.major == 3, "This test can be deleted under Python 3")
```
reinbach: Nice!

ntrncic:
this one can be against develop since it doesn't have migrations and doesn't touch partner and report models

EquiTrack/audit/tests/test_models.py
caktus-philip: Thanks @reinbach. Earlier this year we (Vinod & I) wrote recommendations for UNICEF and the recommendation was to add the 4  `from __future__` imports to all *new* files. But since adding them to existing code can can change behavior, we recommend leaving existing code as-is unless adding a `from __future__` declaration solves a specific problem.

reinbach: :+1: thanks for the clarifcation

caktus-philip: But I guess I added all 4 `from __future__` imports to the existing file `test_models.py`, so not following my own recommendation. 

` ¯\_(ツ)_/¯`

caktus-philip:
Replaced by #993

PR #992, opened 2017-10-31
Fix transition to methods

reinbach:
The following methods;
- `transition_to_cancelled`
- `transition_to_terminated`

Will always result in `TransitionNotAllowed` due to the `agreements_illegal_transition` function always returning `False`.
`agreements_illegal_transition` is a condition for both these methods.

The question is whether these methods are in use, and I should tweak them a bit to allow transitions for the mentioned methods above?
If not in use, should I remove them?

### DIFF
--- a/EquiTrack/partners/models.py
+++ b/EquiTrack/partners/models.py
@@ -13,7 +13,7 @@ from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext as _
 from django.utils.functional import cached_property
 
-from django_fsm import FSMField, transition
+from django_fsm import FSMField, transition, RETURN_VALUE
 from smart_selects.db_fields import ChainedForeignKey
 from model_utils.models import (
     TimeFramedModel,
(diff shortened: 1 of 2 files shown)


robertavram:
transition to cancelled is a bad function name.. but if you look at the transition itself.. source [suspended, signed] target [draft]... that should be illegal.. so we need to keep it in.

the transition to terminated needs to be illegal from status draft... we can keep that function in...

EquiTrack/partners/models.py
```diff
@@ -946,28 +946,28 @@ def update_related_interventions(self, oldself, **kwargs):
...
                 source=[DRAFT],
-                target=[SIGNED],
+                target=SIGNED,
                 conditions=[agreement_transition_to_signed_valid])
     def transition_to_signed(self):
         pass
 
     @transition(field=status,
                 source=[SIGNED],
-                target=[ENDED],
```
robertavram: is there a reason we're changing this? I'd like to keep it as a list if not... just for consistency

reinbach: Actually it errors out if that is in a list like that. `target` can be a list by using `RETURN_VALUE(ENDED,)` but you're expected to return the value then in the method.

reinbach:
The issue with the `transition_to_cancelled` and `transition_to_terminated` methods at the moment is that they will always raise `TransitionNotAllowed` when you call them. No matter the status value of the object.

If the status does not match those in the `source` then you get a `TransitionNotAllowed`, and then if it does match the `source`, the `condition` returns `False` and so you get the `TransitionNotAllowed` again.

PR #997, opened 2017-11-02
Update InterventionAdmendment compute_reference_number

reinbach:
Return integer as `number` field is type IntegerField


EquiTrack/partners/models.py
```diff
@@ -1774,11 +1774,12 @@ class InterventionAmendment(TimeStampedModel):
     tracker = FieldTracker()
 
     def compute_reference_number(self):
+        qs = self.intervention.amendments
         if self.signed_date:
-            return '{0:02d}'.format(self.intervention.amendments.filter(signed_date__isnull=False).count() + 1)
+            seq = qs.filter(signed_date__isnull=False).count() + 1
```
robertavram: this is an old req... we don't need to save any amendment unless is signed... it used to be that we had to add one to generate the document... now we cannot save un-singed amendments.

reinbach: So `signed_document` value is expected (it is no longer nullable) when saving the amendement record initially, but `signed_date` value can/is set in a later transaction, possibly? Does the `amendment_number` need to then change/update if/when the `signed_date` value is set in a later transaction?

robertavram: so.. no.. the amendment number is not changing anymore.. also the signed_date is required by validation... so I believe the else can now be removed.

reinbach: OK, I've updated the method, by removing the `else` statement

PR #998, opened 2017-11-02
Feature flip implementation

vkurup:
Connects unicef/etools-issues#599

Here is my WIP for feature-flipping. I need feedback on the approach/design so please review mercilessly!

We use the [django-waffle](https://github.com/jsocol/django-waffle) package and I created a new app called `features` to modify it slightly to allow segmentation of users by 'country'.

The UI for adding or changing a flag is via the Waffle admin. The important thing is to give it a name and assign it countries via the multiselect box at the bottom. There are multiple options on that page to assign specific users, groups, percentages, etc. All of that comes from Waffle. Which leads to my first question:

Should we allow admins to feature flip ONLY by country? We could start that way, simplifying both the UI to create flags and to check them. It wouldn't be hard to later incorporate more of django-waffles segmentation options (users, groups, percentages) or add our own (user/country combo, etc).
We could choose not to show any of those other fields, and only show countries, but I wanted to show you the possibilities. Alternatively, we could hide this from the admin and create our own UI for tag management.

Segmenting options:
1. Only segment by country
2. Allow segmentation by country plus all of waffle's options

UI options:
1. Use waffle admin, showing all possibilities
2. Use waffle admin, showing only countries.
3. Create our own admin, showing only countries.

Django-waffle also has the concept of switches (always on or off) and samples which are not
dependent on any info in the request. I haven't exposed those via the API, but you can see them in the admin. Flags can be used to accomplish the same thing as switches and samples, with perhaps a tiny performance benefit to the latter.

For the API, I went with Cory and Greg's advice to have a read-only endpoint that only returns the active
flags for a request.

To test this PR:
1. go to the bottom of the admin page and under Waffle, click 'Add Flag'
2. Create a flag named 'can_view_foo' and assign it a few countries (way at the bottom of the form). Make sure that you uncheck the superuser box otherwise the flag will always be on.
3. Go back to the dashboard (/dash/) and make sure you're in the workspace of a country that you added
to the flag.
4. Load /api/v2/features/flags/ -> should return JSON `{active_flags: ['can_view_foo']}`
5. Go back to the dashboard and choose a different country
6. Load /api/v2/features/flags/ -> should return JSON `{active_flags: []}`

Before I call this ready, I will add:
- [ ] Tests
- [ ] Docs for admins on how to add flags
- [ ] Docs for devs on how to use flags, switches, and samples

What else?


reinbach:
Nice, I like how simple you have made the wrapper.

I have some inline comments, but overall I think this is really good!

EquiTrack/features/models.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import unicode_literals
```
reinbach: Maybe add the rest of the `__future__` imports, also to the rest of the files?

EquiTrack/features/views.py
```diff
@@ -0,0 +1,9 @@
+from django.http import JsonResponse
+from features.models import TenantFlag
+
+
+def get_flags(request):
```
reinbach: Any reason not to use DRF? My preference would be for all API endpoints to be consistently implemented, otherwise there is a bit of contextual switch required between different views. And I'm really slow with that.

It would also be a good idea to add handling for requests missing the `tenant` attribute on the request (non authenticated requests?)

EquiTrack/features/admin.py
```diff
@@ -0,0 +1,19 @@
...
+
+from features.models import TenantFlag
+
+
+# add the TenantFlag as an Inline on the Flag Admin
+class TenantFlagInline(admin.StackedInline):
+    model = TenantFlag
+
+
+class TenantFlagAdmin(FlagAdmin):
```
reinbach: Could we change the admin form field used for the Country selector to something similar to Users Countries Available field. Once the list of countries gets long, it will be hard to see which countries are selected at a glance.

EquiTrack/features/models.py
vkurup: oops, yes!

EquiTrack/features/views.py
vkurup: No good reason... just that I'm still not so facile with DRF, whereas whipping up a quick django view is easy :) But you're absolutely right on all points. Will fix.

EquiTrack/features/admin.py
vkurup: I actually switched to the `raw_ids_field` but that widget also isn't the best. You either have to enter actual ID values, or click the 'magnifying glass' which shows a popup allowing you to select 1 country, which closes the popup. So selecting all the countries would take a lot of clicks.

But I agree that the multiselect is horrible too... I'll see if I can find a better widget.

reinbach: :+1:

vkurup:
image

Thanks for the great feedback @reinbach! I found a much better widget, and reworking the code into DRF made me realize that I was short-circuiting the Flag-checking logic by only checking for tenant (rather than all of the other characteristics that django-waffle lets you check). We might decide that that is a better approach for us, but I wasn't meaning to do it yet :) Now the code does the right thing, but please continue to critique my usage of DRF since I'm not sure this is the best way to do it.

I'll hold off on proceeding with tests and documentation until I get a nod of approval from @robertavram, @ntrncic, or @ewheeler

EquiTrack/features/models.py
```diff
@@ -0,0 +1,35 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from django.db import models
+from waffle.models import Flag
+
+from users.models import Country
+
+
+class TenantFlag(models.Model):
```
reinbach: Maybe add a `__str__` method? As `TenantFlag object` is showing up on the admin page

reinbach:
All looks good to me. I like the widget you implemented

czue:
A second option to adding the new `features` app could be to combine it with the `environment` app - which was kind of meant to represent environment-specific adminy data. but no strong feelings.

vkurup:
closed in favor of #1083

PR #999, opened 2017-11-02
removed time from due date on emails action points and trips

marko911:


marko911:
connects unicef/etools-issues#59

EquiTrack/t2f/serializers/mailing.py
```diff
@@ -19,8 +19,8 @@ class TravelMailSerializer(serializers.Serializer):
     estimated_travel_cost = serializers.DecimalField(max_digits=18, decimal_places=2, required=False)
     supervisor = serializers.CharField(source='supervisor.get_full_name')
     traveler = serializers.CharField(source='traveler.get_full_name')
-    start_date = serializers.DateTimeField(format='%m/%d/%Y %H:%M')
-    end_date = serializers.DateTimeField(format='%m/%d/%Y %H:%M')
+    start_date = serializers.DateTimeField(format='%m/%d/%Y')
```
reinbach: I have no idea who receives this mailings, but if there are recipients who are non US, then this format could be pretty confusing at times? Maybe better to change it to `format='%Y-%m-%d'` or something similar?

Please excuse my ignorance :)  Do we have a standard date/time format that is used throughout the system?

marko911: Not sure, this fix addresses an issue with the TIME discrepancy, not the actual format of the date itself so we just decided to remove the time portion. @reinbach

reinbach: You're right, sorry not the place to be raising that point.

ewheeler: @reinbach good point, though. we should pick a display format to use consistently and add it to https://github.com/unicef/etools/blob/master/docs/guidelines.rst

i created an issue here: https://github.com/unicef/etools-issues/issues/889

PR #1011, opened 2017-11-03
Intervention appliedindicator/cluster changes

csabadenes:
- add cluster name to applied indicator
 - add response plan name to applied indicator


csabadenes:
Connects unicef/etools-issues#892

PR #1012, opened 2017-11-03
add formal names for countries

ewheeler:
new jsonfield on publics.Country
new management command to load from csv


ewheeler:
not only did i add this field to the wrong model, but i used the jsonfield library that we are removing...

EquiTrack/publics/management/commands/add_formal_names.py
```diff
@@ -0,0 +1,69 @@
...
+                'UNTERM French Formal': 'fr',
+                'UNTERM Russian Formal': 'ru',
+                'UNTERM Spanish Formal': 'es'}
+
+        # columns we want values from
+        to_pluck = ['ISO3166-1-Alpha-3'] + keys.keys()
+
+        # https://raw.githubusercontent.com/datasets/country-codes/master/data/country-codes.csv
+        # put the csv in the same directory as manage.py
+        reader = csv.DictReader(open('country-codes.csv'))
```
robertavram: can we do a requests.get for the file? it would be cool.. that way we don't have to go download the file in prod by hand

ewheeler: @robertavram good call. actually might be best to me to just run this locally and commit fixtures for these

PR #1013, opened 2017-11-06
Partner pca templates new pdf service

csabadenes:
Move PDF generation to a separate component


csabadenes:
Relates to:
https://github.com/unicef/etools-infra/pull/39
https://github.com/unicef/etools-nginx-proxy/pull/6
https://github.com/unicef/etools-pdf-service

This PR attempts to move the PCA PDF generation functionality to a new docker machine/service. 
The reason for this move is that one of the templates(arabic) requires RTL text support. To achieve this, i had installed some extra libraries, which were considered too heavy to be put on the actual Web server, 
so they were moved to a new machine/Service instead.

The current backend API for the PCA PDF generation should stay unchanged, the backend will make a background request to the PDF service, sending the required context, and the PDF machine should return the generated PDF document.

The authentication is done by sending the HTTP Basic Auth header with every request. The user/pass should be configured as an ENV variable on every server(local/dev/demo/staging/prod).

Another note: i had to add a new and restore few of the recently cleaned up serializers to "partners/serializers/v1.py".

EquiTrack/partners/serializers/v1.py
```diff
@@ -10,3 +15,31 @@ class FileTypeSerializer(serializers.ModelSerializer):
     class Meta:
         model = FileType
         fields = '__all__'
+
+
+class PartnerStaffMemberSerializer(serializers.ModelSerializer):
```
ntrncic: can we use any of the serializers in `partner_organization_v2`.
Maybe we should move the views and serializers to v2 files.

EquiTrack/EquiTrack/utils.py
```diff
@@ -52,6 +53,24 @@ def get_data_from_insight(endpoint, data={}):
     return True, result
 
 
+def load_internal_pdf_template(endpoint='', data={}):
+    api_url = 'http://{}/{}/'.format(
+        settings.PDF_API_URL,
+        'api/pdf/' + endpoint
+    )
+
+    response = requests.post(
```
ntrncic: why a post request?

EquiTrack/EquiTrack/settings/base.py
```diff
@@ -540,6 +540,10 @@ def str2bool(value):
 VISION_USER = os.getenv('VISION_USER', 'invalid_vision_user')
 VISION_PASSWORD = os.getenv('VISION_PASSWORD', 'invalid_vision_password')
 
+PDF_API_URL = os.getenv('PDF_API_URL', '')
+PDF_API_USER = os.getenv('PDF_API_USER', '')
```
ntrncic: I think we are working on JWT authentication. maybe we can send a token to the container. 
@robertavram what do you think?

csabadenes: I can look at the current JWT auth implementation and try to port it to this service.

EquiTrack/EquiTrack/utils.py
csabadenes: POST request is required to send the context data(not possible with GET)

EquiTrack/partners/serializers/v1.py
csabadenes: Sure, i can refactor it to v2.

EquiTrack/EquiTrack/settings/base.py
csabadenes: I have also asked Rob about the auth, i think the conclusion was that JWT auth does not really worth to be implemented for this scenario. 
Also, JWT has multiple steps(fetching a token from the server), which would slow down the request, unless we will use a pre-generated token(not sure how?) which would be basically the same as the current auth, as far as i can tell.

EquiTrack/partners/serializers/v1.py
csabadenes: I have moved the serializers to v2 for both partners and reports, hopefully it is compatible with the direction of the latest related changes.

ntrncic:
@csabadenes please resolve conflicts. sorry

domdinicola:
please rebase the PR

domdinicola:
@csabadenes can you resolve conflicts

PR #1017, opened 2017-11-07
clean apps

domdinicola:


EquiTrack/monitoring/views.py
```diff
@@ -1,6 +1,7 @@
 from django.http import HttpResponse
```
vkurup: This is a relatively small file, and looking over it, I think it would be safe to add the `from __future__ import unicode_literals`. I think it would be needed for the string literals in this file to avoid errors.

EquiTrack/users/tasks.py
```diff
@@ -2,18 +2,21 @@
...
 from django.db.models import Q
 
+import requests
+
 from EquiTrack.celery import app
 from users.models import Country, Section, User, UserProfile
 from vision.models import VisionSyncLog
 from vision.vision_data_synchronizer import VisionException
 
+logger = logging.getLogger()
```
vkurup: logging.getLogger(__name__)

EquiTrack/users/tasks.py
```diff
@@ -86,7 +89,7 @@ def _set_special_attr(self, obj, attr, cleaned_value):
                 if not obj.country == new_country:
                     obj.country = self._get_country(cleaned_value)
                     obj.countries_available.add(obj.country)
-                    print("Country Updated for {}".format(obj))
+                    logger.info("Country Updated for {}".format(obj))
```
vkurup: These and other string literals should probably be unicode literals.

PR #1019, opened 2017-11-07
introduce pip tools to handle requirements

domdinicola:
handle requirements


vkurup:
A few drive-by comments:
- I think this should be against the prp-refactoring branch. I just removed a bunch of requirements in that branch, and we'll have to redo all that work if we make this change in develop
- Is there a waffle card for this change? Just curious about the discussion behind it, i.e., what problem does it solve? I'm not familiar with this approach so this feels like it moves the project away from "Django community standards", but I'd be open to being convinced.
- This is still labelled 'WIP', so maybe it's still coming, but I think I'd want some documentation on how a developer would use these changes.

domdinicola:
Thanks for your comment:

-  I was thinking to keep this as wip until the prp-refactoring branch would have been merged (the requirements you removed are listed at the end of base.in file). After this we will just need to run 'make requirements'

- No waffle card for this yet, mine is just a proposal. The problem to solve is about frozen dependencies. When the version of a dependency is not frozen, could be that my version is . Pip tools ensures that all requirements are pinned, so we are sure that the virtualenv of the developers are perfectly aligned.

- is still WIP because I wanted to discuss this with all of you before.

https://github.com/jazzband/pip-tools

Please give feedback, I'd like to know which is your opinion about this.

domdinicola:
in this way we can use the latest djangosaml2 version and override only what we need
@caktus-philip this should also help the upgrade to Django 1.10 :)

caktus-philip:
@domdinicola I'm +0 on this. I like the idea of more stable virtualenvs. However, I also agree with Vinod's criticisms, particularly the question of "what problem does it solve?" 

It would be nice if this would help with the Django 1.10 path, but there's nothing in the current structure of eTools' `requirements.txt` that prevents upgrading to `djangosaml2` for Django 1.10 compatibility. What's holding that up is that UNICEF has decided that they're going to replace SAML with another authentication method that will make `djangosaml2` irrelevant so no one wants to do the work to replace the current package.

domdinicola:
Ok, great we want to replace djangosaml2.
About the current structure of requirement, we are moving in prod packages like coverage, flake8 and mock.. do we need that?
pip-tools main advantage is that given the first level requirements, it takes care of the dependencies.
Do you have any suggestions in order to build more stable environment?

vkurup:
Unfortunately, we do need packages like coverage, flake8, and mock in prod because the current UNICEF workflow is to run flake8 and unit tests on the 'production' docker images on CircleCI before pushing them to dockerhub to be deployed. So the production requirements need to have those packages.

I wholeheartedly agree with pinning all requirements, including dependencies. In most projects that I've worked on, we just enforce that as a 'guideline'. Any time a new requirement is added, the developer is expected to pin the dependencies as well as the new requirement. In order to facilitate this, we often indent the 'dependencies', like this: https://github.com/SmartElect/SmartElect/blob/develop/requirements/base.txt

Once we've done that, I haven't had any problems. Upgrading top-level packages which have a dependency that is pinned by a different package takes a little bit of thought, but not enough to make me seek out a different solution :)

So again, I'm open to switching to this, but if I had to vote I'd prefer to spend energies elsewhere IMHO.

czue:
I just wanted to throw in a very quick drive-by :+1: to this change. :stuck_out_tongue: 

In other projects I've found requirements/dependency management to be only improved by the introduction of pip-tools. My favorite aspect is the ability to properly manage the removal of libraries. You can often get into a state where you've got a bunch of stuff in your requirements files that's only there because some other library at one point depended on it, whereas with this approach the moment all dependencies are removed the library will also get removed. I've also found it results in much shorter, more concise requirements.in files.

dpoirier:
I don't have experience with pip-tools, but pipenv (http://pipenv.readthedocs.io/en/latest/) appears to be the official recommendation for managing requirements now. If that solves the same problems, maybe that's a better direction to go in, long-term?

domdinicola:
@dpoirier pipenv seems really interesting! Let's see if we can put this in in the future, maybe when we move to python3.

domdinicola:
@robertavram @ntrncic check that though pip tools dependency related to python have been updated. (we need to test the authentication asap in my opinion)

PR #1020, opened 2017-11-08
Coverage tests for partners app

reinbach:
Attempting to increase the test coverage on partners app.


EquiTrack/EquiTrack/factories.py
```diff
@@ -136,10 +136,11 @@ class PartnerStaffFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = partner_models.PartnerStaffMember
 
+    partner = factory.SubFactory('EquiTrack.factories.PartnerFactory')
     title = 'Jedi Master'
     first_name = 'Mace'
     last_name = 'Windu'
-    email = factory.Sequence(lambda n: "mace{}@theforce.org".format(n))
+    email = factory.Sequence(lambda n: "mace{}@example.com".format(n))
```
robertavram: 👍 I don't know why not everybody uses example.com ...

EquiTrack/partners/validation/interventions.py
```diff
@@ -36,7 +36,7 @@ def transition_to_closed(i):
         'earliest_start_date': None,
         'latest_end_date': None
     }
-    for fr in i.frs.filter():
+    for fr in i.frs.all():
```
robertavram: @reinbach 
there used to be an error here... using .all() is cached on a previous version of the object... .filter() renews the query.. otherwise we can use i.total_frs for this function... recently @ntrncic an I made a fix in the mixin that should have fixed this.. we can probably go back to i.total_frs... there are no tests that cover that efficiently.. (check comments on line 29 & 30) 
In order to test:
1 we manually changed an Active PD to Draft (from admin). 
2 we remove the FRs associated to the PD. (from admin)
3 from frontend PMP.. we add FRs and hit save.... the status should change to active on first save but it doesn't.
4 hit save a second time without changing anything... now the status moves to Active...

I'm 99% sure that this was fixed in a recent PR that touched EquiTrack.mixins (you can do a blame to see the change). but it's worth a check.

EquiTrack/partners/validation/interventions.py
```diff
@@ -74,7 +74,7 @@ def transition_to_closed(i):
     return True
 
 
-def transtion_to_signed(i):
+def transition_to_signed(i):
```
robertavram: 👍

EquiTrack/partners/validation/interventions.py
caktus-philip: I think there are some places in the code that use `.filter()` instead of `.all()` to bust cache created by `prefetch_related()`. I don't know if this is one of those places, but it's worth reviewing the code that creates `i`. (Is that an `Intervention`? The choice of variable name & lack of a docstring makes it hard to tell out of content. Not your doing, I realize!)

EquiTrack/partners/views/partner_organization_v2.py
```diff
@@ -206,7 +206,7 @@ def list(self, request, format=None):
         return response
 
 
-class PartnerStaffMemberListAPIVIew(ExportModelMixin, ListCreateAPIView):
+class PartnerStaffMemberListAPIVIew(ExportModelMixin, ListAPIView):
```
caktus-philip: Good catch!

EquiTrack/partners/validation/interventions.py
reinbach: I'm going to revert for now, till we can come up with a unit test to validate

EquiTrack/partners/tests/test_validation_interventions.py
```diff
@@ -0,0 +1,668 @@
...
+        self.expected = {
+            'total_frs_amt': 0,
+            'total_outstanding_amt': 0,
+            'total_intervention_amt': 0,
+            'total_actual_amt': 0,
+            'earliest_start_date': None,
+            'latest_end_date': None
+        }
+
+    def assertFundamentals(self, data):
```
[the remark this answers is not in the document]
reinbach: Good idea, done

PR #1027, opened 2017-11-08
clean utils

domdinicola:


EquiTrack/utils/common/views.py
```diff
@@ -1,8 +1,8 @@
-import six
-
 from django.core.exceptions import ValidationError as CoreValidationError
 from django.db import ProgrammingError
 from django.http import Http404
+
+import six
```
robertavram: can we import six from django?

PR #1033, opened 2017-11-08
add formal country names

ewheeler:
new and improved version of https://github.com/unicef/etools/pull/1012

now adds new field to the correct model!
bonus: doesn't use the soon-to-be-removed jsonfield dependency


ewheeler:
labelling WIP until I add fixtures or download the csv per https://github.com/unicef/etools/pull/1012#pullrequestreview-75202513

ewheeler:
hmm. fixtures probably aren't the best approach bc this model has a couple fields that are not semi-static like the others

not suggesting that we undertake it now, but I feel like we can model this a little better. `users.Country` isn't so much a country.. and isn't strictly a business area..  but seems more like what could be described conceptually as a distinct 'unicef programmatic scope' or 'unicef management node', typically the combination of 

1. a physical unicef office(s)/presence that probably has a regular country programme framing its activities (usually with a document describing objectives for a fixed time period that is approved by UNICEF executive board?) and headed by someone with a JD that reads like a JD for a 
 Representative
2. a primary cost center (business area) with unique ID to find various data that etools syncs/needs from VISION
3. a primary government partner (usually of a nation state) that we work on behalf of (usually with a signed Basic Cooperation Agreement document)
4. a particular, bounded geography (usually a country or set of countries you can point to on a globe) where this government partner effectively projects power .. or geography that contains the locations of a set of programmatic activities/results

we might be benefit from modeling this in a slightly different way that could be closer to our conceptual reality, clearer in terminology, and also accommodate the many exceptions:

1. sometimes we don't have a substantial physical office preference in the same geography/nation as the associated programmatic activities (e.g., Somalia) and sometimes we don't call top-level manager a Representative (e.g., Palestine)
2. sometimes there is not a regular cost center (like Palestine) or there are several due to emergencies (e.g., Syria cross border)
3. sometimes the primary partner is not a fully-recognized nation-state (e.g., Palestine)
4. sometimes we bundle nation states into a single cost center/office (e.g., Pacific islands)

i'm not certain if there is a single, organizational concept/source-of-truth that we can reference that would give us a set of these things
- might be BCAs, but those probably wouldn't account for special CPs/emergency CPs
- 'duty stations' are probably irrelevant to this, but might be a related child concept (same for DSA areas)

ewheeler:
Here is a possible approach:

Four IDs encoded as two lists of tuples delimited by hyphens (tuples of two IDs without any delimiter-- TODO confirm that unicef region codes are _always_ two digits)

If a single `UNICEF Jurisdiction` includes more than one of any ID, then these are all included and tuple pairs are separated by underscores (encoded tuples are ordered lexicographically.. note that business areas should always be treated as strings to preserve leading zeros and can include letters.. e.g., syria cross border is 234R.. which is the same as MENA's business area ?!)

{VISION region code+VISION business area}-{ISO3166-1 Alpha 3+UN M49}

For example, 

Uganda country office would be
`634380-UGA800`

UNICEF Pacific (https://www.unicef.org/pacificislands/) would be
`601430-COK184_FJI242_FSM583_KIR296_MHL584_NIU570_NRU520_PLW585_SLB090_TKL772_TON776_TUV798_VUT548_WSM882`

UNICEF occupied Palestinian territory would be
`597050-PSE275`

UNICEF Syria (and Syria cross border) would be
`594140_59234R-SYR760`

PR #1038, opened 2017-11-09
Ssfa reference number

ntrncic:


EquiTrack/partners/models.py
```diff
@@ -1666,21 +1666,24 @@ def transition_to_terminated(self):
...
     def reference_number(self):
-        number = '{agreement}/{type}{year}{id}'.format(
-            agreement=self.agreement.base_number,
-            code=connection.tenant.country_short_code or '',
-            type=self.document_type,
-            year=self.year,
-            id=self.id
-        )
+        if self.document_type == self.SSFA:
+            number = '{agreement}'.format(
```
vkurup: this could be simplified to `return self.agreement.base_number` (and the `else` block could just return its result as well)

EquiTrack/partners/models.py
```diff
@@ -1666,21 +1666,24 @@ def transition_to_terminated(self):
...
+                agreement=self.agreement.base_number,
+            )
+        else:
+            number = '{agreement}/{type}{year}{id}'.format(
+                agreement=self.agreement.base_number,
+                code=connection.tenant.country_short_code or '',
```
vkurup: `code` is not used in the string. Is it needed?

EquiTrack/partners/models.py
```diff
@@ -1691,6 +1694,10 @@ def update_ssfa_properties(self):
                 self.agreement.start = self.start
                 self.agreement.end = self.end
 
+            if self.agreement.agreement_number != self.number:
```
vkurup: It would make me happy to see a comment here explaining this new logic. A test would also be great, but I don't know if this code path is tested already, nor how difficult that would be.

EquiTrack/partners/models.py
ntrncic: yup. i thought something looked wrong when i wrote it. must have been sleeping

EquiTrack/partners/models.py
ntrncic: removed it

EquiTrack/partners/models.py
ntrncic: not needed anymore. took it out

vkurup:
@ntrncic have you pushed up the changes you mentioned yet? I don't see them on Github.

PR #1044, opened 2017-11-14
Convert t2f models to Python 3-compatible str() methods

caktus-philip:
Convert all models in `t2f` to Python 3-compatible `__str()__` methods using Django's `@python_2_unicode_compatible decorator` while retaining compatibility with Python 2.

Only 3 models in `t2f` have `__str__()` methods, so this is a small change.

### DIFF
--- a/EquiTrack/t2f/tests/test_models.py
+++ b/EquiTrack/t2f/tests/test_models.py
@@ -1,0 +1,49 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import sys
+from unittest import skipIf
+
+
+from EquiTrack.tests.mixins import FastTenantTestCase
+from t2f.tests.factories import (
+    InvoiceFactory,
+    ItineraryItemFactory,
+    TravelFactory,
+    )
+
+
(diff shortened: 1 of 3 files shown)


EquiTrack/t2f/tests/test_models.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@skipIf(sys.version_info.major == 3, "This test can be deleted under Python 3")
+class TestStrUnicode(FastTenantTestCase):
+    '''Ensure calling str() on model instances returns UTF8-encoded text and unicode() returns unicode.'''
+    def test_travel(self):
+        instance = TravelFactory()
+        # This model's __str__() method contains only formatted integers, so it's not possible to challenge it
+        # with non-ASCII text. As long as str() and unicode() succeed, that's all the testing we can do.
+        str(instance)
```
reinbach: Maybe pass in `reference_number` to `TravelFactory()`  and ensure that that value is part of the `str`/`unicode` result?

caktus-philip: @reinbach Good suggestion. I thought that `reference_number` was always constructed programmatically (see `make_travel_reference_number()`) but I was wrong. I updated the test.

PR #1045, opened 2017-11-14
Replace PartnerOrganization hardcoded cash xfer level number with a constant

caktus-philip:
Replace hardcoded `500,000.00` with new constant `PartnerOrganization.CT_CP_AUDIT_TRIGGER_LEVEL`. This makes it easier to find, use, and change this value. 

It also reduces the likelihood of comparing this `Decimal` value to a `float` which is risky in Python. The results of such comparisons depend on the behavior of the underlying C library, and may be unexpected. For instance --

```
$ python2
Python 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 12:54:16)
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from decimal import Decimal
>>> Decimal('500000.00') == 500000.00   # This works as expected
True
>>> Decimal('1.1') == 1.1   # Surprise!
False
>>> Decimal('1.1') < 1.1    # Surprise!
True
```

There are other similarly problematic hardcoded values in `partners/models.py` but I will address them in another PR.

Other fixes in this PR - 
 - fixed many (but not all) cases where 'assessment' was spelled as 'assesment'
 - removed unused param from `PartnerOrganization.audit_needed()`


EquiTrack/partners/tests/test_models.py
```diff
@@ -344,7 +335,6 @@ def test_audit_done_zero(self):
 
     def test_hact_min_requirements_ct_equals_0(self):
         self.partner_organization.total_ct_cy = 0
-        self.partner_organization.save()
```
vkurup: Nice to see these `.save()` calls go away, if not needed!

domdinicola: :+1:

EquiTrack/vision/tasks.py
```diff
@@ -197,8 +197,7 @@ def update_partners(country_name=None):
         for partner in partners:
             try:
                 PartnerOrganization.micro_assessment_needed(partner)
-                if partner.total_ct_cp > 500000:
-                    PartnerOrganization.audit_needed(partner)
+                PartnerOrganization.audit_needed(partner)
```
ntrncic: would like to have this check here or in the audit_needed class method. We do not want to save the Partners unless they have a total_ct_cp > 500000.

caktus-philip: @ntrncic The check for `total_ct_cp > 500k` was in `audit_needed()` prior to this PR, and it's still there. This PR changes the 500k to a constant (`CT_CP_AUDIT_TRIGGER_LEVEL`) but it's still in `audit_needed()` so the logic of that method hasn't changed.

ntrncic: When the task runs it will call `audit_needed` and every `PartnerOrganization` will get saved with `0` or `1` depeding on the threshold. We want to only call `audit_needed` and update the `PartnerOrganization` if it's over the threshold. Please leave the if statement in there.

caktus-philip: @ntrncic Understood. It's just not very DRY to have the same check coded in two places, so I'm trying to avoid that. If you don't want the partner to be saved when `total_ct_cp` < the threshold, then IMO it would be better to code  `audit_needed()` this way --

```
def audit_needed(cls, partner):
    if partner.total_ct_cp > cls.CT_CP_AUDIT_TRIGGER_LEVEL:
        ...do stuff...
```

What do you think about that?

ntrncic: @caktus-philip we still want the partner to be saved on initial run to `1` or `0`. then when running the task, it would update if its a `1`. maybe an `if` `else` would work in `audit_needed`?

caktus-philip: @ntrncic I've recoded `audit_needed()` to only save the partner if the threshold has been exceeded. This is less robust than the previous version because it assumes that `total_ct_cp` only increases and never decreases. Perhaps this is true, but it's not stated anywhere so I didn't want to assume. 

When a partner instance is created, `hact_values` is initialized to `hact_default` which sets `audits_mr` to 0, so that value is set even before `audit_needed()` has run. When `audit_needed()` is called during `save()` or from the vision task, `audits_mr` will get updated to 1 if necessary.

PR #1050, opened 2017-11-15
cleaned locations

domdinicola:
removed old command
flake8
isort


EquiTrack/locations/tasks.py
```diff
@@ -7,6 +7,7 @@
 from carto.sql import SQLClient
 
 from EquiTrack.celery import app
+
 from .models import Location
```
reinbach: Change to absolute import

EquiTrack/locations/tests/test_models.py
```diff
@@ -1,16 +1,9 @@
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
-from __future__ import unicode_literals
+from __future__ import absolute_import, division, print_function, unicode_literals
```
reinbach: I believe the preference is to have these on separate lines

domdinicola: Do you think would be useful to handle 'Multi line output modes' with 'Vertical Hanging Indent' mode (option 3)? (check https://pypi.python.org/pypi/isort)

reinbach: I guess there is no mandated way that these are imported, I've always seen them on separate lines and I've alwats done that as well.

domdinicola: To me is fine, I've been cleaning those in all the code :/

caktus-philip: I agree with Greg. Obviously both are valid syntax and I don't have a strong personal preference for either version. However, I think it's good practice to respect the established style on a project even if that style is not my personal preference.

domdinicola: done!

PR #1051, opened 2017-11-15
improved celery

domdinicola:


EquiTrack/locations/admin.py
```diff
@@ -50,9 +50,8 @@ class CartoDBTableAdmin(admin.ModelAdmin):
     actions = ('import_sites',)
 
     def import_sites(self, request, queryset):
-
         for table in queryset:
-            update_sites_from_cartodb.delay(table)
+            update_sites_from_cartodb.delay(table.pk)
```
vkurup: :+1:

EquiTrack/locations/tasks.py
```diff
@@ -85,7 +86,9 @@ def create_location(pcode, carto_table, parent, parent_instance,
 
 
 @app.task
-def update_sites_from_cartodb(carto_table):
+def update_sites_from_cartodb(carto_table_pk):
+
+    carto_table = CartoDBTable.objects.get(pk=carto_table_pk)
```
reinbach: If `CartoDBTable` does not exist, do we want to handle it more gracefully?
Granted it looks like it is being called from a queryset loop, so probably not a big deal.

I'm actually not sure what our team's standard procedure is for handling exceptions.

domdinicola: I think you're right @reinbach, although is called by a queryset loop, maybe in the future update_sites_from_cartodb can be called from somewhere else

PR #1052, opened 2017-11-16
Api changes for the PMP dropdowns and users v3 endpoint

csabadenes:
Connects unicef/etools-issues#921


EquiTrack/users/tests/test_views.py
```diff
@@ -260,3 +261,7 @@ def test_minimal_verbosity(self):
                                         data={'verbosity': 'minimal'}, user=self.unicef_superuser)
         response_json = json.loads(response.rendered_content)
         self.assertEqual(len(response_json), 2)
+
+    def test_retrieve_user_countries(self):
+        response = self.forced_auth_req('get', '/api/v3/users/country/', user=self.unicef_user)
```
caktus-philip: It'd be better to use `reverse()` than hardcoding the URL.

EquiTrack/users/tests/test_views.py
```diff
@@ -260,3 +261,7 @@ def test_minimal_verbosity(self):
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
caktus-philip: This would be a stronger test if you actually verify that what is returned is what you expect.

EquiTrack/users/serializers_v3.py
```diff
@@ -43,6 +43,27 @@ class Meta:
         )
 
 
+class MinimalCountryDetailSerializer(serializers.ModelSerializer):
+    local_currency = serializers.CharField(source='local_currency.name')
```
caktus-philip: Since the other two `local_currency` fields are read-only, should this one also be read-only?

EquiTrack/users/tests/test_views.py
csabadenes: I have made some changes to the tests, thx for the feedback.

EquiTrack/users/serializers_v3.py
```diff
@@ -43,6 +43,27 @@ class Meta:
         )
 
 
+class MinimalCountryDetailSerializer(serializers.ModelSerializer):
```
ntrncic: change name to `CountryDetailSerializer`

EquiTrack/users/urls_v3.py
```diff
@@ -7,7 +7,7 @@
     # api
     url(r'^profile/$', MyProfileAPIView.as_view(http_method_names=['get', 'patch']), name="myprofile-detail"),
     url(r'^changecountry/$', ChangeUserCountryView.as_view(http_method_names=['post'])),
-    url(r'^country/$', CountryView.as_view(http_method_names=['get']), name="country-detail"),
+    url(r'^country/$', CountryView.as_view(http_method_names=['get']), name="v3-user-country-details"),
```
ntrncic: please remove the `v3` in name

EquiTrack/users/views_v3.py
```diff
@@ -99,7 +103,7 @@ class CountryView(ListAPIView):
     Country is determined by the currently logged in user.
     """
     model = Country
-    serializer_class = CountrySerializer
```
ntrncic: is the Country Serializer used anymore? If it's not, it can be removed.

csabadenes: CountrySerializer it is still used in ProfileRetrieveUpdateSerializer / Profile view.

PR #1064, opened 2017-11-21
Create a mixin which return an empty queryset when in public tenant for tenant models

domdinicola:
We get a lot of errors in Sentry..
do you like this idea?

### DIFF
--- a/EquiTrack/EquiTrack/views.py
+++ b/EquiTrack/EquiTrack/views.py
@@ -1,4 +1,5 @@
 from django.views.generic import TemplateView
+from django.db import connection
 
 
 class MainView(TemplateView):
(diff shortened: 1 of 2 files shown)


czue:
Nice idea! My quick thoughts:

Pros: simple, elegant, quiet solution.
Possible Downside: when a tenant schema queries the public tenant could that sometimes be masking a bigger problem? If so, this might prevent those errors from reaching us.

reinbach:
I think it is a good idea to remove these errors in Sentry, as no action is being taken on this error I take it.

I agree with @czue, it would be a good to know why/how are requests being made on these endpoints for a public schema, is there an underlying issue?

caktus-philip:
Agree with Greg and Cory. I wouldn't want to silence the errors without first being sure that they're always 100% harmless. The mixin would need a big docstring that explains (a) what it does, (b) why it's needed, (c) when to use it, and (d) why it's not harmful.

domdinicola:
I added a doc string and a warning log. Mine was just an idea, let's keep it on hold for a while.

robertavram:
Hey @domdinicola I think this is not needed... There shouldn't be anyone in global schema making requests on those views... if they are.. let it explode... I think maybe a better approach would be to prevent users from being in global schema in any of the front end apps...

closing for now..

PR #1075, opened 2017-11-28
Increase test coverage in EquiTrack app

reinbach:
Ended up refactoring the parses file, attempted to make it clearer and to remove the use of `exec` statements.

Docstrings and tests were added in the initial commits, so if the refactoring is no good, we just need to rollback the last commit. 

I think this needs a fair amount of more testing with real use cases, to ensure nothing is missing.

We may want to change how we handle multipart data, but changing it would require tweaks to the frontend and backend, so think that needs more of a discussion around that. For now I just tackled the initial notes raised by @caktus-philip and @vkurup

@caktus-philip has done some nice research around potential improvements to handling the posting of data, which I have included below;

Vinod and I flagged parsers.py early on as trouble. I'd love to see it go away. Here's my initial notes on the subject -- 
https://docs.google.com/document/d/1xmimX7aNh8zcRUSSrmsCHaJTxGPQoVlqLblQVn6__70/edit#heading=h.308mi3o99f9h

This is potentially interesting GUI stuff where the request gets created and sent —
pmp/app-elements/agreements/agreement-data/agreement-item-data.html (_triggerAgreementRequest)
pmp/bower_components/etools-ajax/etools-ajax-request-behavior.html (_prepareConfigOptions)
pmp/bower_components/etools-ajax/etools-ajax-data-behavior.html (_prepareMultiPartFormData)

I haven't actually seen one of these requests 'in the field', so to speak, so my understanding so far is all theoretical. It'd be really helpful to see what data is actually sent across the wire. Figuring out how to make that happen is where I paused on this problem.

Here's some writing on the web that might be useful for coding a solution -- 
https://stackoverflow.com/questions/3938569/how-do-i-upload-a-file-with-metadata-using-a-rest-web-service
https://philsturgeon.uk/api/2016/01/04/http-rest-api-file-uploads/
https://shazwazza.com/post/uploading-files-and-json-data-in-the-same-request-with-angular-js/

Especially ”Just because you're not wrapping the entire request body in JSON, doesn't meant it's not RESTful”) — https://stackoverflow.com/a/13076550


EquiTrack/EquiTrack/tests/test_parsers.py
```diff
@@ -0,0 +1,125 @@
...
+        self.assertEqual(parsers.int_or_str(1), 1)
+
+    def test_str_to_int(self):
+        self.assertEqual(parsers.int_or_str("1"), 1)
+
+    def test_str(self):
+        self.assertEqual(parsers.int_or_str("one"), "one")
+
+
+class TestCreateListsFromKeys(TestCase):
```
caktus-philip: It'd be good to add a test that exercises the sorting that happens in `create_lists_from_keys()`.

EquiTrack/EquiTrack/tests/test_parsers.py
```diff
@@ -0,0 +1,125 @@
...
+        """If only one element in list then return str of that element"""
+        self.assertEqual(parsers.create_key([1]), "1")
+
+    def test_multi_element(self):
+        """If multiple elements in list then return dict str of path"""
+        res = parsers.create_key(["d", "k", "2"])
+        self.assertEqual(res, "d[k][2]")
+
+
+class TestBuildDict(TestCase):
```
caktus-philip: `build_dict()` is the most complicated function in this code so I'd like to see it tested more thoroughly.

EquiTrack/EquiTrack/tests/test_parsers.py
```diff
@@ -0,0 +1,125 @@
...
+    def test_str_list_key(self):
+        """Check that string keys results in valid dictionary"""
+        data = {
+            "[d _obj key-2]": "val-2",
+            "[d][_obj][key-2]": "val-2",
+        }
+        res = parsers.parse_multipart_data(data)
+        self.assertEqual(res[""], {
+            "d": {"key-2": "val-2"},
+        })
```
caktus-philip: Maybe if I understood `parse_multipart_data()` better, this would be obvious, but is `''` guaranteed to be the only key in the result? Maybe that should be an assertion?

reinbach:
Looks like circleci had issues on the last check, if someone who has the permissions, rebuild the last check ([#6119](https://circleci.com/actions/retry/github/unicef/etools/6119), that would be great

ntrncic:
@reinbach here is the error i get
```File "/usr/local/lib/python2.7/site-packages/django/utils/datastructures.py" in __getitem__
  85.             raise MultiValueDictKeyError(repr(key))

Exception Type: MultiValueDictKeyError at /api/v2/interventions/422/
Exception Value: "u'attachments'"```

PR #1086, opened 2017-12-04
more cleaning

domdinicola:


EquiTrack/vision/vision_data_synchronizer.py
```diff
@@ -131,6 +132,7 @@ def sync(self):
             log.successful = True
         except Exception as e:
             log.exception_message = e.message
-            raise VisionException(message=e.message), None, sys.exc_info()[2]
+            six.reraise(VisionException(message=e.message), None, sys.exc_info()[2])
```
caktus-philip: This isn't necessary. `six` allows you to write code that's compatible with Python 2 and 3 simultaneously, but that's not a goal for eTools. When the time comes to convert eTools to Python 3, `2to3` will convert the `raise` line to this Python 3-compatible version --
```
raise VisionException(message=e.message).with_traceback(sys.exc_info()[2])
```

domdinicola: I agree with you, I used six in this case because flake8 is starting notifying these category of statement: E901 SyntaxError: invalid syntax

ntrncic:
@caktus-philip we have updated the code with your recommendation

PR #1087, opened 2017-12-05
User model refactoring - do not reference auth.User directly

csabadenes:
Connects unicef/etools-issues#648


EquiTrack/partners/serializers/partner_organization_v2.py
```diff
@@ -28,6 +28,7 @@ def validate(self, data):
         data = super(PartnerStaffMemberCreateSerializer, self).validate(data)
         email = data.get('email', "")
         active = data.get('active', "")
+        User = get_user_model()
```
vkurup: The User model will be set at server-start, so we shouldn't need to call it for every call to `validate`. Generally we do this just at the top of each file where `User` is needed:

```
from django.contrib.auth import get_user_model
User = get_user_model()
```

@ntrncic @robertavram I know this is already merged, but I'd be happy to create a PR to update the excellent work already done by @csabadenes here. There's also still one reference to the User model in `users.models`

PR #1089, opened 2017-12-05
remove paintstore

domdinicola:
part 2 would remove reference of paintstore from migrations and the proper requirement


EquiTrack/locations/models.py
```diff
@@ -9,11 +9,10 @@
 from django.utils.encoding import python_2_unicode_compatible
 
 from mptt.models import MPTTModel, TreeForeignKey
-from paintstore.fields import ColorPickerField
 
 logger = logging.getLogger(__name__)
 
-
+# TODO remove when remove the library
 def get_random_color():
```
reinbach: I believe you can remove this function completely

domdinicola: not yet.. it's used in location 0001 migration

EquiTrack/locations/migrations/0001_initial.py
```diff
@@ -32,7 +31,7 @@ class Migration(migrations.Migration):
                 ('name_col', models.CharField(default=b'name', max_length=254)),
                 ('pcode_col', models.CharField(default=b'pcode', max_length=254)),
                 ('parent_code_col', models.CharField(blank=True, max_length=254, null=True)),
-                ('color', paintstore.fields.ColorPickerField(blank=True, default=locations.models.get_random_color, max_length=7, null=True)),
+                ('color', models.CharField(blank=True, default='#ffffff', max_length=7, null=True)),
```
ntrncic: we should leave the default `get_random_color`

EquiTrack/locations/models.py
```diff
@@ -9,19 +8,10 @@
 from django.utils.encoding import python_2_unicode_compatible
 
 from mptt.models import MPTTModel, TreeForeignKey
-from paintstore.fields import ColorPickerField
 
 logger = logging.getLogger(__name__)
 
 
-def get_random_color():
```
ntrncic: leave this in

domdinicola: Why? Is not used elsewhere.

domdinicola:
I don't think we don't need a migration to change it to charfield. I edit the first migration to use a charfield from the beginning (the ColorPickerField extends the charfield) so we can remove all references to paintstore in eTools (otherwise I would have to keep the requirement in or o would get an import error)

PR #1101, opened 2017-12-20
Convert all models in users to Python 3-compatible str() methods

caktus-philip:
Convert all models in `users` to Python 3-compatible `__str()__` methods using Django's `@python_2_unicode_compatible decorator` while retaining compatibility with Python 2.


caktus-philip:
@domdinicola grazie!

Issue #1071, opened 2017-11-24
Organize users views


caktus-philip:
Classes in the `EquiTrack/users` tree be easier to understand if they were reorganized and de-duped. Currently the following files exist in `users` --

```
urls.py
urls_v2.py
urls_v3.py
serializers.py
serializers_v3.py
views.py
views_v3.py
```

The file names imply that there are 3 independent API versions, but they're actually not independent at all. URLs defined in 4 different URL files (the 3 listed above, plus `EquiTrack/urls.py` point to `users/views.py`. In addition, some of the implementations in `views_v3.py` are duplicates of implementations in `views.py`.

View tests are all lumped into one file (`test_views.py`) which makes it hard to figure out what is and is not tested. 

Last but not least, it will be messy to decommission versions of this API because it's not clear where one ends and another begins.

Some starting suggestions on how to reorganize -- 
 - Create `users/views/views_v2.py`, add stubs that subclass V1 views
 - Create `users/views/views_v3.ChangeUserCountryView`, have it subclass V1 views of the same name
 - Change `users/views/views_v3.MyProfileAPIView` to subclass V1 view of the same name.
 - Change `users/views/views_v3.CountryView` to subclass V1 view of the same name.
 - Rename `users.views.UsersView` to `UsersThatAreStaffView`
 - Break tests up into `test_views_v1.py`, `test_views_v2.py`, and `test_views_v3.py`

Other, related improvements -- 
 - Address TODO in `urls.py` (see related #1070)
 - Improve test coverage for views
 - Add explicit tests of serializers

PR #1105, opened 2017-12-28
Reorganize users views into separate version files

reinbach:
Including the tests

See #1071


EquiTrack/users/tests/test_views_v3.py
```diff
@@ -3,11 +3,14 @@
...
 from django.core.urlresolvers import reverse
 from rest_framework import status
 
-from EquiTrack.factories import UserFactory
+from EquiTrack.factories import GroupFactory, UserFactory
 from EquiTrack.tests.mixins import APITenantTestCase
+from users.serializers_v3 import AP_ALLOWED_COUNTRIES
 
 
 class TestCountryView(APITenantTestCase):
```
caktus-philip: IMHO this test belongs in the tests file that corresponds to the views file in which implementation resides (v1 or v2 if you decide to move it there).

EquiTrack/users/views_v2.py
```diff
@@ -0,0 +1,49 @@
+from users import views as v1
+
+
+class UserAuthAPIView(v1.UserAuthAPIView):
+    """Stub for UserAuthAPIView"""
+
```
caktus-philip: I didn't know you could define a class like this! I thought you always needed at least a `pass` statement. Thanks for the education.

caktus-philip:
Not sure why CircleCI failed. All tests pass for me and flake8 runs clean.

reinbach:
The failing test is due to the end date used in t2f tests, end of year. It has been fixed in other branches.

EquiTrack/users/views.py
```diff
@@ -73,14 +73,14 @@ def post(self, request, format=None):
             return Response(status=status.HTTP_204_NO_CONTENT)
 
 
-class UsersView(ListAPIView):
+class StaffUsersView(ListAPIView):
```
domdinicola: what's going on here?

reinbach: Gave the class a more accurate name, as this view returns only staff users.

domdinicola: mmm it seems you changed also the permissions no?

reinbach: Yes, as I think we wanted to limit the list of staff users to superusers and staff and not any user logged in

PR #1107, opened 2017-12-28
increase size of city, country and phone_number in PartnerOrganization

domdinicola:
I noticed error when saving PartnerOrganization when synchronizing


domdinicola:
@vkurup to be honest I was thinking the same... until I found out this in the logs several times

[worker-vision-staging-1]2017-12-21T14:41:28.474419709Z DataError: value too long for type character varying(32)

PR #1115, opened 2018-01-02
change current year default behavior for HactHistory

domdinicola:
otherwise we will have a migration every year
same problem in InterventionPlannedVisits, but is in prp-refactoring branch


EquiTrack/EquiTrack/utils.py
```diff
@@ -185,6 +185,10 @@ def process_file():
     return response
 
 
+def get_current_year():
```
ntrncic: Lets move this method into the migration itself.

vkurup: Just a note that I'd prefer for this method to be here because while it is required in the migration, it is also in `EquiTrack.hact.models`, and I don't like importing migration code from 'production' code. (I also think that there are other places where we get the current year, in slightly different ways, which could be improved by using this method, but that's not a high priority)

domdinicola: :+1:  especially if we squash migrations we risk to loose this

PR #1117, opened 2018-01-03
Update djangosaml2 requirement to 0.16.11

reinbach:
Being able to upgrade Django 1.10 requires us to either replace djangosaml2 or update it.

If we want to update it, I [forked](https://github.com/unicef/djangosaml2) the maintained version of djangosaml2 which is Django 1.10 compatible and applied the tweaks that had been made to our current version of djangosaml2.

I'm not able to fully test this implementation, but it is available in this branch if we want to try it out. It may provide us with a step closer to being able to upgrade to Django 1.10


domdinicola:
I'd love to update this, so we can move upgrading Django.
In the future most probably we will use openid connect, but not sure when.

reinbach:
Did you apply the tweaks that were made to the previous djangosaml2 version by Rob?

domdinicola:
yes Greg, I created a tag on this branch https://github.com/domdinicola/djangosaml2/tree/feature/etools

reinbach:
So the point of failure for us is at https://github.com/rohe/pysaml2/blob/master/src/saml2/response.py#L839

We do not appear to have any `.extension_elements` attribute on our assertion, so the assertions are not decrypted and assigned. Resulting in the error `'NoneType' object has no attribute 'authn_statement'` at

```
/usr/local/lib/python2.7/site-packages/saml2/response.py in session_info: line 1088;

       authn_statement = self.assertion.authn_statement[0]
```

EquiTrack/EquiTrack/settings/base.py
```diff
@@ -401,7 +401,8 @@ def str2bool(value):
     'givenName': ('first_name',),
     'surname': ('last_name',),
 }
-SAML_DJANGO_USER_MAIN_ATTRIBUTE = 'email'
+SAML_DJANGO_USER_MAIN_ATTRIBUTE = 'username'
```
robertavram: why are we changing this to username?

reinbach: We were forcing the main attribute to be username in the code, as an enhancement to the djangosaml2 code base, See djangosaml2/backends.py file ~line 163;

```
if django_user_main_attribute != 'username':
    user_query_args['username'] = main_attribute
```

I thought if we could set this setting to username we would not need to have this extra code change.

reinbach:
I think I'm going to have to cancel this pull request and create a new one

EquiTrack/EquiTrack/settings/base.py
domdinicola: :+1:

reinbach:
This pull request has been replaced by https://github.com/unicef/etools/pull/1129

EquiTrack/EquiTrack/settings/base.py
robertavram: oh... I forgot why we put it in there.. but if memory serves me well, there was a problem when we were automatically creating users that did not exist in our DB.

We wanted the attribute that stores as the username and email to come from the email claim. However since only one key was used to get_or_create the user it ended up that after the first user was created `{'email':'johnsmith@example.com', 'username':''}` all subsequent users were raising an error since username needed to be unique...

I completely forgot which code did that... but that's our intention.. to allow users that have never been set up in our db to be automatically created in our db the moment they log in. All this while making sure that the email attribute is what we store as username and as email for the user. I think that's also why we opted for querying by email.

```
if django_user_main_attribute != 'username':
    user_query_args['username'] = main_attribute
```
i think here user_query_args dict does not have the 'username' key, just email.. and this gives user_query_args two keys to "get_or_create" by.. however.. I have not looked at the code and I might be wrong.. let me know if so :)

reinbach: Mmmm... I think my changes are not right then. I've gone ahead and removed those changes.

PR #1123, opened 2018-01-11
Issue of adding extra required model fields for validation

csabadenes:
Connects unicef/etools-issues#1101

@robertavram, this is just an idea regarding the permission issues we discussed on hangouts.


robertavram:
great minds think alike :) 👍 

fixed slightly differently through #1127

PR #1124, opened 2018-01-12
Add filtering on funds reservations for grants and donors

reinbach:
[ch2083]

Made use of the grant number for relation the between grants and fundsreservationitems. Did not make changes to the vision sync.

The request current requires `values` to be provided, I left that requirement in place. Should that be relaxed, so long as either `values`, `grants`, and/or `donors` filter is provided instead?

See https://github.com/unicef/etools-issues/issues/1080


EquiTrack/funds/views.py
```diff
@@ -17,6 +17,43 @@ class FRsView(APIView):
     """
     permission_classes = (permissions.IsAdminUser,)
 
+    def filter_by_donors(self, qs, donors):
```
vkurup: I think it would be slightly clearer if we renamed `donors` to `donor_pks` (same for `grants` -> `grant_pks` in the next method)

EquiTrack/funds/views.py
```diff
@@ -17,6 +17,43 @@ class FRsView(APIView):
...
+            "name",
+            flat=True
+        ).filter(donor__pk__in=donors)
+        qs = self.filter_by_grants(qs, None, list(grant_numbers))
+        return qs
+
+    def filter_by_grants(self, qs, grants, grant_numbers=[]):
+        """Filter queryset by grant ids provided
+
+        `name` field in Grants table matches `grant_number` in
```
vkurup: :+1: for this comment!

EquiTrack/funds/tests/test_views.py
```diff
@@ -8,15 +8,24 @@
...
+    FundsReservationHeaderFactory,
+    FundsReservationItemFactory,
+    GrantFactory,
+    InterventionFactory,
+    UserFactory,
+)
 
 
 class TestFRHeaderView(APITenantTestCase):
-    fixtures = ['initial_data.json']
```
vkurup: Yay, for fixtures going away! :)

robertavram:
Hmm.. good thought but we need to check if that's the right data to filter by.

Donors and Grants are populated in our DB via an API that I believe doesn't include all available donors and grants. (I need to double check)

However, I think when we synchronize the FR LI we should create / relate to records in the DB rather than stitching together by donor/grant name when we filter.

There is another model for Grants in the publics app (public schema). Not sure yet if we should relate to that. @ntrncic and I would need to dig a bit to find the source of truth for donors and grants.

robertavram:
holding for now

ntrncic:
Donors and grants are not updated in our DB anymore. We have to write new vision synchronizers for Donors and Grants
I agree we should connect the `FundsReservationItem` and `FundsCommitmentItem` in the DB then  by just a number/name. 
I wouldn't use anything in the publics schema.

ntrncic:
I would merge this until we figure out what to do with Donors and grants

reinbach:
@ntrncic in this PR `Grants` are being filtered using the `name` field on grant and then via `FundsReservationItem` through to `FundsReservationHeader`

PR #1125, opened 2018-01-12
Python 3 support

dpoirier:
* Maintain python 2.7 support for now
* Add "tox" config so we can easily run tests with different
  Python versions
* Update some requirements
* Modify some code that would only work on Python 2.7
* Use "six" as needed to make code work with both Python 2 and 3.


dpoirier:
Being replaced by separate PRs against prp-refactoring-python3.

PR #1127, opened 2018-01-15
Fix section validator

robertavram:
fixes: unicef/etools-issues#1101 
fixes: unicef/etools-issues#1043
fixes: unicef/etools-issues#1074
More tests needed.


robertavram:
@csabadenes can you branch out of this one and add some tests for transitioning to signed when sections are present vs when sections are not present?

PR #1152, opened 2018-01-17
Prep funds app for Python3

dpoirier:


EquiTrack/funds/urls.py
```diff
@@ -1,8 +1,8 @@
-from __future__ import unicode_literals
+from __future__ import absolute_import, division, print_function, unicode_literals
 
 from django.conf.urls import url
 
-from views import (
```
domdinicola: I would specify the module (funds.views)

dpoirier:
@domdinicola I've updated this.

PR #1159, opened 2018-01-17
Prepare reports app for Python 3

dpoirier:


EquiTrack/reports/migrations/0016_remove_indicators.py
```diff
@@ -12,7 +12,7 @@ def remove_all_indicators(apps, schema_editor):
    # use the save method to mark invalid fields
    deleted = IndicatorBlueprint.objects.all().delete()
    if deleted[0]:
-       print "Deleted Indicators: {}".format(deleted[0])
+       print(("Deleted Indicators: {}".format(deleted[0])))
```
domdinicola: we can remove a couple of ()

dpoirier: Updated.

PR #1161, opened 2018-01-17
Prepare t2f app for Python 3

dpoirier:


EquiTrack/t2f/serializers/travel.py
```diff
@@ -451,7 +453,7 @@ def update(self, instance, validated_data, model_serializer=None):
         return instance
 
     def update_object(self, obj, data):
-        m2m_fields = {k: data.pop(k, []) for k in data.keys()
+        m2m_fields = {k: data.pop(k, []) for k in data
```
reinbach: Maybe you want to work with copy of keys here as well?

dpoirier: Updated.

PR #1177, opened 2018-01-24
Django 1.10

dpoirier:
* Django 1.10 adds some constraint checking at the end of tests.
  Some of our tests fail at that point. Disable it for now.

* Django object managers are defined on the model classes but
  not instances. Django 1.10 seems to enforce that harder. Fix
  a couple of places that breaks.

* Also rename our FastTenantTestCase to EToolsTenantTestCase, because
  the Django tenant schemas package also has a FastTenantTestCase
  class that is different from ours and it was confusing (to me and
  my IDE, anyway).


EquiTrack/EquiTrack/tests/mixins.py
```diff
@@ -86,10 +86,29 @@ def assertIntParamRegexes(self, names_and_paths, namespace):
                     raise AssertionError(fail_msg)
 
 
-class FastTenantTestCase(TenantTestCase):
+# Not really a mixin - should this be in mixins.py?
```
domdinicola: good point, cases.py file as in the base class?

EquiTrack/partners/tests/test_admin.py
```diff
@@ -5,7 +5,7 @@
 
 from django.contrib.admin.sites import AdminSite
 
-from EquiTrack.tests.mixins import FastTenantTestCase as TenantTestCase
+from EquiTrack.tests.mixins import EToolsTenantTestCase as TenantTestCase
```
domdinicola: we can replace EToolsTenantTestCase in all the file instead of 'import as'

EquiTrack/partners/tests/test_intervention_tags.py
```diff
@@ -8,7 +8,7 @@
     InterventionFactory,
     PartnerFactory,
 )
-from EquiTrack.tests.mixins import FastTenantTestCase as TenantTestCase
+from EquiTrack.tests.mixins import EToolsTenantTestCase as TenantTestCase
```
domdinicola: we can replace EToolsTenantTestCase in all the file instead of 'import as'

EquiTrack/partners/tests/test_models.py
```diff
@@ -37,7 +37,7 @@
     TravelActivityFactory,
     UserFactory,
 )
-from EquiTrack.tests.mixins import FastTenantTestCase as TenantTestCase
+from EquiTrack.tests.mixins import EToolsTenantTestCase as TenantTestCase
```
domdinicola: we can replace EToolsTenantTestCase in all the file instead of 'import as'

EquiTrack/partners/tests/test_validation_agreements.py
```diff
@@ -8,7 +8,7 @@
     PartnerStaffFactory,
     UserFactory,
 )
-from EquiTrack.tests.mixins import FastTenantTestCase as TenantTestCase
+from EquiTrack.tests.mixins import EToolsTenantTestCase as TenantTestCase
```
domdinicola: we can replace EToolsTenantTestCase in all the file instead of 'import as'

EquiTrack/snapshot/tests/test_models.py
```diff
@@ -8,7 +8,7 @@
     InterventionFactory,
     UserFactory,
 )
-from EquiTrack.tests.mixins import FastTenantTestCase as TenantTestCase
+from EquiTrack.tests.mixins import EToolsTenantTestCase as TenantTestCase
```
domdinicola: we can replace EToolsTenantTestCase in all the file instead of 'import as'

EquiTrack/snapshot/tests/test_utils.py
```diff
@@ -3,7 +3,7 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from EquiTrack.tests.mixins import FastTenantTestCase as TenantTestCase
+from EquiTrack.tests.mixins import EToolsTenantTestCase as TenantTestCase
```
domdinicola: we can replace EToolsTenantTestCase in all the file instead of 'import as'

EquiTrack/users/tests/test_models.py
```diff
@@ -11,7 +11,7 @@
     SectionFactory,
     UserFactory,
 )
-from EquiTrack.tests.mixins import FastTenantTestCase as TenantTestCase
+from EquiTrack.tests.mixins import EToolsTenantTestCase as TenantTestCase
```
domdinicola: we can replace EToolsTenantTestCase in all the file instead of 'import as'

domdinicola:
Looks good to me! I just added few changes.
Can you have a look to users.tasks line 132 (create_or_update_user)

dpoirier:
Is there something at user.tasks line 132 that'll break with Django 1.10?  I'm not seeing what it is.

domdinicola:
in Django 1.10 username max_lenght changes from 30 to 150
https://docs.djangoproject.com/en/1.10/ref/contrib/auth/

I think we want to save the new username due this new constraint (please check as well line 105 in _set_attribute)

dpoirier:
I've updated this. Should I remove the changes-requested label or is that up to you @domdinicola ?

dpoirier:
FYI, with these changes, tests all pass with Django 1.11.9 as well, so we could jump straight to an LTS version of Django if we wanted to. (https://www.djangoproject.com/download/#supported-versions)

ntrncic:
thanks for this. I'm going to put it on hold until PRP refactoring is merged.

EquiTrack/users/tasks.py
```diff
@@ -129,8 +129,9 @@ def create_or_update_user(self, ad_user):
                 return
 
         # TODO: MODIFY THIS TO USER THE GUID ON THE PROFILE INSTEAD OF EMAIL on the USer
-        user, created = User.objects.get_or_create(email=ad_user['internetaddress'],
-                                                   username=ad_user['internetaddress'][:30])
+        user, created = User.objects.get_or_create(
+            email=ad_user['internetaddress'],
+            username=ad_user['internetaddress'][:User._meta.get_field('username').max_length])
```
domdinicola: this will explode since are both unique, maybe we will need to run a one time script to update all the username before

dpoirier: good point. there might be users previously created whose username was truncated to 30 chars, and this would try to create a new user, but with an existing email address.

We could change the query to look for ```Q(username=inetaddress[:30])|Q(username=inetaddress[:150])``` (or whatever the actual lengths are)?

domdinicola: mmm does it make sense?
it's like saying if a<3 or a <5 .... it can be replaced by if a<5 
I would leave:

email=ad_user['internetaddress'],
username=ad_user['internetaddress']

and before we go in production we run something

for user in User.objects.all():
    if user.username!=user.email:
        user.username=user.email
        user.save()

so we override the 'old' username

dpoirier: We can do this in a migration:

    User.objects.exclude(username=F('email')).update(username=F('email'))

Do we have any conventions on whether we lowercase the email when copying to the username or anything like that?  And maybe our queries should be case-insensitive?

domdinicola: I suppose are stored in azure in lowercase

dpoirier: It sound like you're saying you think it's probable that azure stores this field in lowercase. Is there somewhere I could verify that? 

Also, I don't know what azure is, or how it's related to this project or this problem.

domdinicola: I would just create a data migration to update the username fields.
@robertavram  @ntrncic what do you think?

dpoirier:
Tests finally passing again.

EquiTrack/users/tasks.py
dpoirier: I've added that migration.

robertavram:
Congrats!

PR #1041, opened 2017-11-13
migrate to Celery 4

domdinicola:

PR #1178, opened 2018-01-24
Retry vision sync tasks on failure

dpoirier:
* Use Celery's retry feature to retry vision sync tasks that fail, at
  least a few times.

* Remove some unused code from vision/tasks.py.

For https://github.com/unicef/etools-issues/issues/1115

### DIFF
--- a/EquiTrack/vision/tasks.py
+++ b/EquiTrack/vision/tasks.py
@@ -1,5 +1,4 @@
-import datetime
-import time
+from __future__ import absolute_import, division, print_function, unicode_literals
 
 from django.db import connection
 from django.utils import timezone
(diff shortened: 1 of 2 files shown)


EquiTrack/vision/tasks.py
```diff
@@ -130,93 +87,10 @@ def sync_handler(country_name, handler):
...
-    countries = Country.objects.filter(vision_sync_enabled=True)
-    if country_name is not None:
-        countries = countries.filter(name=country_name)
-    for country in countries:
-        connection.set_tenant(country)
-        logger.info(u'Updating '.format(country.name))
-        partners = PartnerOrganization.objects.all()
-        for partner in partners:
+            # This might be worth retrying.
             try:
```
domdinicola: do you think we can do something at decorator level like this http://docs.celeryproject.org/en/latest/userguide/tasks.html#automatic-retry-for-known-exceptions

dpoirier: We're using Celery 3.1 and autoretry_for wasn't added until Celery 4.0.    I agree the decorator approach is a lot clearer, but we'd have to upgrade Celery first.

domdinicola:
what if we do that against https://github.com/unicef/etools/pull/1041?
what's your idea about this?

dpoirier:
Making that improvement against #1041 sounds like a good idea. I added a comment there that the retry code can be simplified along with that work.

dpoirier:
This PR replaced by https://github.com/unicef/etools/pull/1180

PR #1102, opened 2017-12-20
Factory cleanup

reinbach:
Move factories into relevant apps, and a few more tweaks based on ticket 644

See https://github.com/unicef/etools-issues/issues/644

PR #1179, opened 2018-01-25
Merge base test classes

reinbach:
FastTenantTestCase and APITenantTestCase merged

See https://github.com/unicef/etools-issues/issues/643

### DIFF
--- a/EquiTrack/EquiTrack/tests/mixins.py
+++ b/EquiTrack/EquiTrack/tests/mixins.py
@@ -3,7 +3,7 @@ from django.core.management import call_command
 from django.core.urlresolvers import resolve, reverse, NoReverseMatch
 from django.db import connection
 from rest_framework.test import APIClient, force_authenticate, APIRequestFactory
-from tenant_schemas.test.cases import TenantTestCase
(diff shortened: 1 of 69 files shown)


vkurup:
Heh, this is the work that I've been mentioning is stashed in my local repo waiting for #1102 to get through. :)

I also just looked at #1177 and it looks like @dpoirier  has made a similar change there (but against prp-refactoring). Wondering if we should just go with that? Thoughts?

reinbach:
Ha, oh dear. I didn't even look around to see if this ticket had been worked on, just saw it in "On deck" and it was assigned to me. Not sure which repo is best to go with. Either way there is going to be a bunch of merging to deal with.

I'm fine with which ever direction we want to go with.

vkurup:
Yes, having multiple 'assignees' on a ticket is definitely a confusing situation... I'm also fine with any PR. The code in both look fine to me.

reinbach:
Only thing I would mention is that in Dan's changes we still have the 2 TestCases in place, I think we want to get those merged. So if you does that, then we can get rid of this pull request.

ntrncic:
I agree we should use Dan's PR...

PR #1183, opened 2018-01-29
Run tests under both python 2 and 3 using tox

dpoirier:
This runs all the unit tests, and flake8, under both Python 2 and 3.  To run locally, change to EquiTrack and run "tox". CircleCI will also run all the tests when changes are pushed.

This changes the Dockerfile that builds the docker image that CircleCI uses to test.

I have *temporarily* built the image from the updated docker file and pushed it to docker hub as dpoirier/dan-etools, and modified .circleci/config.yml in this PR to use that image, so if circleci tests pass on this pull request, that's good.

I've also created and pushed a temporary branch to make sure the current develop branch will also pass tests when using the new image in circleci - see https://circleci.com/gh/unicef/etools/6686 for that successful test.

I think, before merging this, we'll want to change the .circleci/config.yml back to running tests using unicef/etools:test-base, but for that to work, we'll need to use this Dockerfile to update the image under that name on Docker hub. I'll need advice/help/approval to do that.


.circleci/config.yml
```diff
@@ -17,18 +19,20 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: deps1-{{ .Branch }}-{{ checksum "EquiTrack/requirements/base.txt" }}
+          key: deps1-{{ .Branch }}-{{ checksum "EquiTrack/requirements/base.txt" }}-{{ checksum "EquiTrack/requirements/local.txt" }}-{{ checksum ".circleci/config.yml" }}
```
domdinicola: why do we need to change this to local?

dpoirier: Ah - I was assuming that the test dependencies were in local.txt and that's what we were installing requirements from before running tests, but it turns out that neither of those is true - all the test packages are in base.txt and we install from production.txt before running tests.  I'll update this.

domdinicola: I think would be better to move test dependencies in another requirement file.
Maybe we can think about that later.

EquiTrack/EquiTrack/settings/local.py
```diff
@@ -16,6 +16,7 @@
 
 
 # change config to remove CSRF verification in localhost in order to enable testing from postman.
+REST_FRAMEWORK = REST_FRAMEWORK  # noqa: F405 (defined in wildcard import from base)
```
domdinicola: why do we need this?

EquiTrack/requirements/base.txt
```diff
@@ -34,15 +34,20 @@ djangorestframework-recursive==0.1.1
 # Other Python packages
 psycopg2==2.6.2
 celery==3.1.25
-GDAL==1.10.0           # DO NOT USE GDAL <= 2.1.0
+GDAL==1.10.0           # DO NOT USE GDAL <= 2.1.0  (DAN: I DO NOT UNDERSTAND THIS LINE - we appear to be using GDAL <= 2.1.0 here)
```
domdinicola: check if this can be upgraded

EquiTrack/EquiTrack/settings/local.py
dpoirier: Not strictly needed, but same as line 81, just shuts up flake8.

EquiTrack/requirements/base.txt
dpoirier: Can we do this in a separate ticket? GDAL 1.10.0 appears to work okay with Python 3 (assuming the unit tests are covering that).

robertavram:
@dpoirier 
I think you're right with this:
`I think, before merging this, we'll want to change the .circleci/config.yml back to running tests using unicef/etools:test-base, but for that to work, we'll need to use this Dockerfile to update the image under that name on Docker hub. I'll need advice/help/approval to do that.`

let's point it to unicef/etools:test-base and we'll push the updated image there...

.circleci/config.yml
```diff
@@ -17,18 +19,20 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: deps1-{{ .Branch }}-{{ checksum "EquiTrack/requirements/base.txt" }}
+          key: deps1-{{ .Branch }}-{{ checksum "EquiTrack/requirements/base.txt" }}-{{ checksum ".circleci/config.yml" }}
       - run:
+          name: create main virtualenv, install tox, generate test envs
           command: |
-            virtualenv env1
+            test -d env1 || python3 -m venv env1
```
domdinicola: is this to test if a venv exists and in case create a new one?

EquiTrack/EquiTrack/settings/production.py
```diff
@@ -2,6 +2,7 @@
 from cryptography.hazmat.primitives import serialization
 from cryptography.x509 import load_pem_x509_certificate
 from cryptography.hazmat.backends import default_backend
+import os
```
robertavram: instead of using these to trick flake8, let's either ignore these from flake config, or the lines that are using it.

EquiTrack/EquiTrack/tests/test_validation_mixins.py
```diff
@@ -17,9 +17,6 @@ def test_basic_validation_error(self):
         e = BasicValidationError()
         self.assertEqual(six.text_type(e), '')
 
-        e = BasicValidationError('hello world')
-        self.assertEqual(six.binary_type(e), b'hello world')
```
robertavram: do we not need this anymore?

EquiTrack/requirements/base.txt
```diff
@@ -42,15 +42,27 @@ sqlparse==0.2.2  # for django debug toolbar to not break admin
 azure-storage==0.20.2
 requests==2.11.1
 xlrd==1.0.0
-xhtml2pdf==0.1b3
+
+#xhtml2pdf==0.0.6  <--- Python 2 only
```
robertavram: not ideal...

EquiTrack/requirements/base.txt
```diff
@@ -42,15 +42,27 @@ sqlparse==0.2.2  # for django debug toolbar to not break admin
...
+# This is a work-in-progress:
+git+https://github.com/xhtml2pdf/xhtml2pdf@d2ca1d3b054a6e3faf89ce13116da5cacf5e1cf2#egg_info=xhtml2pdf
+# TODO: use a released version once Python 3 is supported that way.
+# TODO: or switch from xhtml2pdf to weasyprint, which seems to be the better-maintained replacement
```
domdinicola: here it says that python 3 is supported...
https://pypi.python.org/pypi/xhtml2pdf/0.1b1

EquiTrack/requirements/base.txt
```diff
@@ -42,15 +42,27 @@ sqlparse==0.2.2  # for django debug toolbar to not break admin
...
+# http://weasyprint.readthedocs.io/en/latest/index.html
```
domdinicola: please remove this

robertavram: @domdinicola wanna look into these options here?

EquiTrack/tox.ini
```diff
@@ -0,0 +1,33 @@
+[tox]
+envlist = {flake8-py2,flake8-py3,python2.7,python3.4,python3.6}
```
domdinicola: I would leave in the envlist only flake8-py2,flake8-py3 (execute tests 5 times looks too much)

.circleci/config.yml
dpoirier: right, just me trying to be cute I guess. Probably better if I write this as "if [ ! -d env1 ] ; then python3 -m venv env1; fi"

EquiTrack/EquiTrack/tests/test_validation_mixins.py
dpoirier: I don't think we do need this anymore. It was "str(e)" in Python 2, but there's no analog in Python 3 to convert an object to a byte string.  six.binary_type(e) doesn't really make sense.

EquiTrack/tox.ini
dpoirier: The flake8-* envs only run flake8, not the unit tests.

Maybe just run the unit tests on Python 2.7 and whatever Python 3.x we expect to be used in production when this is deployed?

dpoirier:
Pushed several changes:

    * Change name of docker image back to what it was.
    * Rewrite a line in the circleci config to create a Python 3
      virtualenv in a simpler way.
    * Remove some non-functional code that was just added to quiet flake8.

EquiTrack/requirements/base.txt
dpoirier: Yes, obviously I tried the latest pypi version before falling back to an unreleased commit. And I included a TODO with the information about weasyprint replacing the dead xhtml2pdf project, but I don't think this ticket is the right place to rewrite the PDF code to use a different library.

So I'm not clear on what this comment means. Are you requesting a specific change, or is this okay the way it is for this PR?

dpoirier:
CirclCI tests now fail because the current docker image (unicef/etools:test-base) does not provide Python 3.  Can someone update the image based on the updated Dockerfile in this ticket?

EquiTrack/runtests.sh
```diff
@@ -1,20 +1,24 @@
...
+# Ensure there are no missing migrations. If there are, show what they are.
+python manage.py makemigrations --dry-run >mig.out
+grep 'No changes detected' mig.out || (cat mig.out; rm mig.out; echo 'There are changes which require migrations.';  exit 1)
+rm mig.out
 
 # Run unittests and coverage report
 coverage erase
-coverage run manage.py test --noinput --keepdb --settings="$TEST_SETTINGS" "$@"
-coverage report --include "$@/*.py"
+time coverage run manage.py test --noinput --keepdb "$@"
```
reinbach: Curious about the use of `time`, are we using the timed output later on, or just have it available in console output?

EquiTrack/runtests.sh
```diff
@@ -1,20 +1,24 @@
...
+coverage report -m
```
reinbach: Do we need the `-m` option? Especially without a specified module?

EquiTrack/runtests.sh
dpoirier: I just like to see how long tests took, especially when they take a while and I'm likely to have switched to doing something else why they ran.

EquiTrack/runtests.sh
dpoirier: The -m option shows which lines are missing coverage. If we were doing html reports, we could just look at the source displayed there, but with the text output I don't know how else to know where we still need coverage.

reinbach: 👍

dpoirier:
I've updated to XHTML2PDF 0.2.1, an official release that supports Python 3.

Is there anything else blocking this pull request? I'd like to wrap it up and get it merged (at least to prp-refactoring-python3).

PR #1186, opened 2018-02-02
Ensure --keepdb option works

reinbach:
Fix tests so we can make use of --keepdb option


EquiTrack/monitoring/tests/test_checks.py
```diff
@@ -7,24 +7,29 @@
...
 
 class TestCheckDB(TestCase):
     def test_no_users(self):
-        check = check_db()
-        self.assertFalse(check.success)
-        self.assertEqual(
-            check.message,
-            "{}:OK No users found in postgres".format(
-                settings.DATABASES["default"]["NAME"]
+        # if --keepdb flag used Users probably exist in db
```
domdinicola: :100:

PR #1191, opened 2018-02-07
azure integration

domdinicola:


EquiTrack/azure_graph_API/client.py
```diff
@@ -0,0 +1,67 @@
...
+            'resource': 'https://graph.microsoft.com/'
+        }
+        response = requests.post(path, post_dict)
+        if response.status_code == 200:
+            jresponse = response.json()
+            token = jresponse['access_token']
+            cache.set(AZURE_GRAPH_API_TOKEN, token)
+            logger.info('Token retrieved')
+        else:
+            logger.error('Error during token generation')
```
reinbach: Very minor, on success logging says `retrieved` but on error logging says `generation`, maybe have them both say the `retrieved`?

EquiTrack/azure_graph_API/tests/test_client.py
```diff
@@ -0,0 +1,76 @@
...
+
+from azure.common import AzureHttpError
+from azure_graph_API.client import azure_sync_user, azure_sync_users, get_token
+from mock import Mock, patch
+
+
+class TestClient(TestCase):
+
+    def test_get_token_ok(self):
+        mock_obj = type(str("PostMock"), (object,), {
```
reinbach: In future, for testing/mocking responses to to requests you could use this handy little library https://github.com/getsentry/responses

EquiTrack/azure_graph_API/client.py
```diff
@@ -0,0 +1,67 @@
...
+
+from django.conf import settings
+from django.core.cache import cache
+
+import requests
+from azure.common import AzureHttpError
+from azure_graph_API.utils import handle_record, handle_records
+
+logger = logging.getLogger(__name__)
+AZURE_GRAPH_API_TOKEN = 'azure_graph_api_token'
```
reinbach: Later on your cache key constants have a suffix `_CACHE_KEY` on the constant, maybe do the same here for consistancy. To me, it is clearer as well, that this is the cache key and not the actual token.

EquiTrack/azure_graph_API/client.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def azure_sync_users(url):
+    access_token = get_token()
+    jresponse = {}
+    while url:
+        headers = {'Authorization': 'Bearer {}'.format(access_token)}
+        response = requests.get(url, headers=headers)
+        jresponse = response.json()
+        if response.status_code == 200:
```
reinbach: I'd suggest wrapping this `if/else` section and reuse it in both `azure_sync_user` and `azure_sync_users`, makes it a little DRYier.
The `url = jresponse.get('@odata.nextLink', None)` statement can happen afterwards. Or maybe there is no `@odata.nextLink` in a single user sync response?

EquiTrack/azure_graph_API/client.py
```diff
@@ -0,0 +1,67 @@
...
+        }
+        response = requests.post(path, post_dict)
+        if response.status_code == 200:
+            jresponse = response.json()
+            token = jresponse['access_token']
+            cache.set(AZURE_GRAPH_API_TOKEN, token)
+            logger.info('Token retrieved')
+        else:
+            logger.error('Error during token generation')
+            raise AzureHttpError('Error during token generation {}'.format(response.status_code), response.status_code)
```
reinbach: The exception is already being passed the `status_code`, maybe pass the logger.error message into the `format()`
This also applies to the other `AzureHttpError` calls.

EquiTrack/azure_graph_API/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+
+logger = get_task_logger(__name__)
+
+AZURE_GRAPH_API_USER_CACHE_KEY = 'azure_graph_api_user_cache_key'
+
+
+@app.task
+def sync_user(username):
+    logger.info('Azure Sync User started for {}'.format(username))
+    url = 'https://graph.microsoft.com/{}/users/{}'.format(
```
reinbach: The first part of the url is consistant (`https://graph.microsoft.com/{}/.format(settings.AZURE_API_VERSION)`) across all these tasks, maybe move this to the settings file. If it changes, then one place to update

EquiTrack/users/tasks.py
```diff
@@ -102,10 +104,11 @@ def _set_attribute(self, obj, attr, value):
         """
         # clean the value
         field = obj._meta.get_field(attr)
-        if field.get_internal_type() == "CharField" and len(value) > field.max_length:
+        if type(value) == list and len(value) == 1:
```
reinbach: I'm curious why, if it is a list and is greater than one we use the list, otherwise use the first element in the list?

EquiTrack/users/tasks.py
```diff
@@ -122,59 +125,61 @@ def _set_attribute(self, obj, attr, value):
...
+            # user.is_staff = user.is_staff or True
+            # user.is_active = user.is_active or True
+
+            if created:
+                user.groups.add(self.groups['UNICEF User'])
+                logger.info('Group added to user {}'.format(user))
+
+            # most attributes are direct maps.
+            for attr, attr_val in ad_user.iteritems():
+                mapped_attribute = self.ATTR_MAP.get(attr, 'unusable_attr')
```
reinbach: 👍

EquiTrack/users/tasks.py
```diff
@@ -22,6 +22,8 @@
 
 class UserMapper(object):
 
+    key_attribute = 'internetaddress'
```
reinbach: This is a constant, maybe make it uppercase

EquiTrack/users/tasks.py
```diff
@@ -230,6 +235,29 @@ def map_users(self):
                     user.profile.save()
 
 
+class AzureUserMapper(UserMapper):
+    key_attribute = 'userPrincipalName'
```
reinbach: This is a constant, maybe make it uppercase

EquiTrack/azure_graph_API/tests/test_tasks.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from django.test import TestCase
+
+from azure_graph_API.tasks import sync_all_users, sync_delta_users, sync_user
+from mock import Mock, patch
+
+
+class TestTasks(TestCase):
```
reinbach: Maybe add some `asserts` that the mock objects are/were called with correct parameters in these tests?

EquiTrack/azure_graph_API/client.py
domdinicola: +1

EquiTrack/azure_graph_API/client.py
domdinicola: sure

EquiTrack/azure_graph_API/client.py
domdinicola: @odata.nextLink is not in the single user response... I'd like to find a way to make it dry as well..
they call different functions (handle_record and handle_records)

EquiTrack/azure_graph_API/client.py
domdinicola: I did like this, because when the exception is raised I couldn't see the status code..
how would you write it?

EquiTrack/users/tasks.py
domdinicola: ;)

EquiTrack/users/tasks.py
domdinicola: mmm that was my thought.
value = ', '.join([str(val) for val in value])
does this look better to you?

EquiTrack/azure_graph_API/client.py
reinbach: You may be able to have a single `azure_sync_users` and tweak the `handle_records` utils function to have an `else` that passes the `jresponse` parameter to the `handle_record` function.

EquiTrack/azure_graph_API/client.py
reinbach: How were you viewing the AzureHttpError exception, I assumed that the passing of the status code as a paramter to that error/exception would have shown it, but I guess not. Yeah if it was more helpful to add the status code to the message, that's fine then.

```
msg = "'Error during token generation ({})".format(response.status_code)
logger.error(msg)
raise AzureHttpError(msg, response.status_code)
```

EquiTrack/azure_graph_API/tasks.py
domdinicola: ;)

EquiTrack/azure_graph_API/tests/test_client.py
domdinicola: wow amazing!!!!!! I'm going to move this in

EquiTrack/azure_graph_API/client.py
domdinicola: :+1:

EquiTrack/users/tasks.py
```diff
@@ -102,10 +104,11 @@ def _set_attribute(self, obj, attr, value):
         """
         # clean the value
         field = obj._meta.get_field(attr)
-        if field.get_internal_type() == "CharField" and len(value) > field.max_length:
+        if type(value) == list:
+            value = '- '.join([str(val) for val in value])
```
vkurup: This should be `u'- '`, if we expect any non-ascii characters in `value` (or even if we don't expect it...  just to be safe) :)

EquiTrack/users/tasks.py
```diff
@@ -102,10 +104,11 @@ def _set_attribute(self, obj, attr, value):
...
+        if field.get_internal_type() == "CharField" and value and len(value) > field.max_length:
             cleaned_value = value[:field.max_length]
-            logging.warn('The attribute "%s" was trimmed from "%s" to "%s"' %
-                         (attr, value, cleaned_value))
+            logging.warn('The attribute "%s" was trimmed from "%s" to "%s"' % (attr, value, cleaned_value))
```
vkurup: This wasn't part of your change, but I'd make this `u'The attribute...` and also switch from string interpolation to using logging arguments. (Also this should be `logger`, not `logging`.. Both will work, but `logger` will use the logger specific to this module that was set up in line 20)

EquiTrack/users/tasks.py
```diff
@@ -122,59 +125,62 @@ def _set_attribute(self, obj, attr, value):
...
     @transaction.atomic
     def create_or_update_user(self, ad_user):
-        logger.debug(ad_user.get('sn'), ad_user.get('givenName'))
+        logger.debug(ad_user.get(self.KEY_ATTRIBUTE))
         for field in self.REQUIRED_USER_FIELDS:
             if not ad_user.get(field, False):
                 logger.info("User doesn't have the required fields {}".format(ad_user))
                 return
 
         # TODO: MODIFY THIS TO USER THE GUID ON THE PROFILE INSTEAD OF EMAIL on the USer
```
vkurup: Can we remove the TODO above, since I can't figure out what it means? :)

EquiTrack/users/tasks.py
```diff
@@ -122,59 +125,62 @@ def _set_attribute(self, obj, attr, value):
...
-                profile.save()
+            user, created = User.objects.get_or_create(email=ad_user[self.KEY_ATTRIBUTE],
+                                                       username=ad_user[self.KEY_ATTRIBUTE][:30])
+            if created:
+                user.set_unusable_password()
+
+            try:
+                profile = user.profile
+            except ObjectDoesNotExist:
+                logger.warning('No profile for user {}'.format(user))
```
vkurup: This should be unicode, and should use logging args rather than interpolation. (same for other logging calls in this file)

EquiTrack/users/tasks.py
```diff
@@ -122,59 +125,62 @@ def _set_attribute(self, obj, attr, value):
...
+                    profile_modified = profile_modified or modified
+            try:
+                if user_modified:
+                    logger.debug('saving modified user')
+                    user.save()
+                if profile_modified:
+                    logger.debug('saving profile for: {}'.format(user))
+                    profile.save()
+            except IntegrityError as e:
+                logging.error('Integrity error on user: {} - exception {}'.format(user.email, e))
```
vkurup: First s/logging/logger. If we use `.exception`, instead of `.error`, then Python will automatically include the traceback in the message, so we won't need to include it explicitly.

EquiTrack/users/tasks.py
```diff
@@ -230,6 +236,30 @@ def map_users(self):
...
+    ATTR_MAP = {
+        'userPrincipalName': 'username',
+        'mail': 'email',
+        'givenName': 'first_name',
+        'surname': 'last_name',
+        'id': 'guid',
+        'businessPhones': 'phone_number',
+        'extension_f4805b4021f643d0aa596e1367d432f1_extensionAttribute1': 'country',
+        'extension_f4805b4021f643d0aa596e1367d432f1_extensionAttribute2': 'staff_id',
+        # 'unicefSectionCode': 'section_code',
```
vkurup: Is this needed here?

EquiTrack/users/tasks.py
```diff
@@ -275,7 +305,6 @@ def map_users():
 def sync_users_local(n=20):
     user_sync = UserMapper()
     with open('/code/etools.dat') as csvfile:
-        # with open('/Users/Rob/Downloads/users.dat') as csvfile:
```
vkurup: Thank you! :)

EquiTrack/users/templates/admin/users/user/change_form.html
```diff
@@ -0,0 +1,6 @@
+{% extends "admin/change_form.html" %}{% load i18n %}
+{% block object-tools-items %}
+    {{ block.super }}
+    <li><a href="{% url 'admin:users_sync_user' original.id %}" class=''>{% trans "Sync User" %}</a>
```
vkurup: Let's close the `</li>` tag.

domdinicola: ;)

domdinicola:
@reinbach @vkurup can you verify if the pr satisfy your comments ?

vkurup:
Since you've rebased the branch, instead of adding commits, it's hard to see what you've changed and what you've left the same. I see that many of my comments were not addressed or discussed, so I'm wondering if you just didn't receive them from Github's notifications? I'll review the PR from the beginning again.

EquiTrack/azure_graph_api/client.py
```diff
@@ -0,0 +1,74 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import logging
+import requests
+
+from django.conf import settings
+from django.core.cache import cache
+
+from azure.common import AzureHttpError
```
vkurup: I believe `azure.common` is a third-party import ,so it should be up there with the `django` ones.

EquiTrack/azure_graph_api/client.py
```diff
@@ -0,0 +1,74 @@
...
+from azure_graph_api.utils import handle_records
+
+logger = logging.getLogger(__name__)
+AZURE_GRAPH_API_TOKEN_CACHE_KEY = 'azure_graph_api_token_cache_key'
+
+
+def get_token():
+    """
+    generate or retrieve token for azure integration
+    """
```
vkurup: I'm quite unfamiliar with all things Azure :) It would be nice if we included a link to the Azure docs which you referenced while writing this code, so that I (or someone else) would find it easier to change the code if we need to make changes.

EquiTrack/azure_graph_api/client.py
```diff
@@ -0,0 +1,74 @@
...
+    logger.info('Request for token started')
+
+    token = cache.get(AZURE_GRAPH_API_TOKEN_CACHE_KEY)
```
vkurup: I'm still wondering how caching works in this use case. I understand how the cache gets populated, but I don't see where the cache gets invalidated. My brief reading of the azure docs indicate that the token expires after 1 hour by default. Won't we have a problem once that hour is up? We will keep trying the same expired access token until the cache expires.

EquiTrack/azure_graph_api/client.py
```diff
@@ -0,0 +1,74 @@
...
+    if not token:
+        logger.info('Request generating new token')
```
vkurup: grammar: 'Request new token to be generated'

EquiTrack/azure_graph_api/client.py
```diff
@@ -0,0 +1,74 @@
...
+            'resource': settings.AZURE_GRAPH_API_BASE_URL
+        }
+        response = requests.post(path, post_dict)
+        if response.status_code == 200:
+            jresponse = response.json()
+            token = jresponse['access_token']
+            cache.set(AZURE_GRAPH_API_TOKEN_CACHE_KEY, token)
+            logger.info('Token retrieved')
+        else:
+            logger.error('Error during token retrieved')
```
vkurup: s/retrieved/retrieval

EquiTrack/azure_graph_api/client.py
```diff
@@ -0,0 +1,74 @@
...
+            raise AzureHttpError('Error during token retrieved {}'.format(response.status_code), response.status_code)
```
vkurup: It seems like there is more useful information in the response, which might help us debug things: https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols-oauth-client-creds#error-response-1 (Maybe it's not needed for the moment, but putting a link in as a comment could help a future developer add useful things to the error message later)

EquiTrack/azure_graph_api/client.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    headers = {'Authorization': 'Bearer {}'.format(access_token)}
+    response = requests.get(url, headers=headers)
+    jresponse = response.json()
+    if response.status_code == 200:
+        logger.info('Azure: Information retrieved')
+        handle_records(jresponse)
+        url = jresponse.get('@odata.nextLink', None)
+    else:
+        logger.error('Error during syncronization process')
```
vkurup: s/syncronization/synchronization

EquiTrack/azure_graph_api/tasks.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from django.conf import settings
+from django.core.cache import cache
+
+from azure_graph_api.client import azure_sync_users
+from celery.utils.log import get_task_logger
```
vkurup: celery is a third party import so should be up with Django.

EquiTrack/azure_graph_api/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+from EquiTrack.celery import app
+
+logger = get_task_logger(__name__)
+
+AZURE_GRAPH_API_USER_CACHE_KEY = 'azure_graph_api_user_cache_key'
+
+
+@app.task
+def sync_user(username):
+    logger.info('Azure Sync User started for {}'.format(username))
```
vkurup: This is a general comment for all of the `logger` calls throughout this PR. It's better to do:
```
logger.info('blah %s', username)
```
rather than either of the following:
```
logger.info('blah %s' % username)
logger.info('blah {}.format(username))
```
The first version lets the logging library do the string interpolation, and the logging library is smart enough to only do it if the message actually needs to be logged (based on the loglevel and other settings). This is a performance improvement, which admittedly doesn't matter in most cases, but does add up and is significant in some cases. In addition, logging in this way allows third-party tools (like Sentry as an example) to group messages better since it can group by the message, without the placeholder variance causing a problem.

There's a TODO issue to fix this project-wide: https://github.com/unicef/etools-issues/issues/649

EquiTrack/azure_graph_api/tests/test_client.py
```diff
@@ -0,0 +1,67 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from django.conf import settings
+from django.test import TestCase
+
+import responses
+from azure.common import AzureHttpError
+from azure_graph_api.client import azure_sync_users, get_token
```
vkurup: This import should be separated from the others since it is a project import.

EquiTrack/azure_graph_api/tests/test_client.py
```diff
@@ -0,0 +1,67 @@
...
+    def test_get_token_ok(self):
+        responses.add(
+            responses.POST, settings.AZURE_TOKEN_URL,
+            json={'access_token': 't0k3n'}, status=200
+        )
+        token = get_token()
+        assert token == 't0k3n'
+
+    @responses.activate
+    def test_get_token_ko(self):
```
vkurup: what does `_ko` mean here? Would something like `test_get_token_bad_response` be accurate?

EquiTrack/azure_graph_api/tests/test_utils.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from django.contrib.auth import get_user_model
+
+from azure_graph_api.utils import handle_record, handle_records
```
vkurup: This import should be with the project imports down below

EquiTrack/azure_graph_api/utils.py
```diff
@@ -0,0 +1,66 @@
...
+def handle_records(jresponse):
+    if 'value' in jresponse:
+        for record in jresponse['value']:
+            handle_record(record)
+    else:
+        handle_record(jresponse)
+
+
+def handle_record(record):
+    logger.debug('Azure: Information retrieved'.format(record.get('userPrincipalName', None)))
```
vkurup: See my previous comment about improving all of the logger statements. In addition, this message doesn't have a placeholder (`{}`), so `userPrincipalName` will not get logged.

EquiTrack/azure_graph_api/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    logger.debug('Department: {}'.format(record.get('department', None)))
+
+    logger.debug('Country Code [Business Area Code]: {}'.format(
+        record.get('extension_f4805b4021f643d0aa596e1367d432f1_extensionAttribute1', '-')))
+    logger.debug('Country: {}'.format(record.get('country', '-')))
+    logger.debug('Index Number: {}'.format(
+        record.get('extension_f4805b4021f643d0aa596e1367d432f1_extensionAttribute2', '-')))
+    logger.debug('Nationality: {}'.format(record.get(
+        'extension_f4805b4021f643d0aa596e1367d432f1_extensionAttribute3', '-')))
+    # logger.debug('4: {}'.format(record.get(
```
vkurup: Are the commented lines here still needed or can we remove them?

EquiTrack/users/templates/admin/users/country/change_form.html
```diff
@@ -1,10 +1,10 @@
...
-    <li><a href="{% url 'admin:users_country_fund_commitment' original.id %}" class=''>{% trans "Sync Fund Commitment" %}</a>
-    <li><a href="{% url 'admin:users_country_fund_reservation' original.id %}" class=''>{% trans "Sync Fund Reservation" %}</a>
-    <li><a href="{% url 'admin:users_country_partners' original.id %}" class=''>{% trans "Sync Partners" %}</a>
-    <li><a href="{% url 'admin:users_country_programme' original.id %}" class=''>{% trans "Sync Programme" %}</a>
-    <li><a href="{% url 'admin:users_country_ram' original.id %}" class=''>{% trans "Sync RAM" %}</a>
+    <li><a href="{% url 'admin:users_country_fund_commitment' original.id %}" class=''>{% trans "Sync Fund Commitment" %}</a></li>
+    <li><a href="{% url 'admin:users_country_fund_reservation' original.id %}" class=''>{% trans "Sync Fund Reservation" %}</a></li>
+    <li><a href="{% url 'admin:users_country_partners' original.id %}" class=''>{% trans "Sync Partners" %}</a></li>
+    <li><a href="{% url 'admin:users_country_programme' original.id %}" class=''>{% trans "Sync Programme" %}</a></li>
+    <li><a href="{% url 'admin:users_country_ram' original.id %}" class=''>{% trans "Sync RAM" %}</a></li>
```
vkurup: Thank you :)

EquiTrack/vision/adapters/partner.py
```diff
@@ -194,7 +194,8 @@ def _partner_save(processed, partner):
                 processed += 1
 
             except Exception as exp:
-                logger.exception(u'Exception occurred during Partner Sync: {}'.format(exp.message))
+                logger.exception(u'Exception occurred during Partner Sync {}: {}'.format(
+                    partner['VENDOR_CODE'], exp.message))
```
vkurup: I don't think we need `exp` or `exp.message` at all here. `logger.exception` includes the traceback of the exception in the log entry, and the last line of the of the traceback will show the `exp.message` value. Unless it's important to have that message in the first line of the log message AND the last line of the log message?

EquiTrack/azure_graph_api/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+    url = '{}/{}/users/{}'.format(
+        settings.AZURE_GRAPH_API_BASE_URL,
+        settings.AZURE_GRAPH_API_VERSION,
```
reinbach: Looks like the azure url is always based on both `settings.AZURE_GRAPH_API_BASE_URL`, and `settings.AZURE_GRAPH_API_VERSION` you could make it a little DRYer and have a setting value that combines these values together and use that value throughout this file.

EquiTrack/azure_graph_api/tests/test_client.py
```diff
@@ -0,0 +1,67 @@
...
+            settings.AZURE_GRAPH_API_PAGE_SIZE
+        )
+        responses.add(
+            responses.GET, url, status=200,
+            json={'@odata.deltaLink': 'delta'},
+        )
+        delta = azure_sync_users(url)
+        self.assertEquals(delta, 'delta')
+        self.assertEqual(token.call_count, 1)
+        self.assertEqual(token.call_args[0], ())
```
reinbach: Interesting, I have not seen the mock args checked this way. I learnt something to day.
FYI, I generally write this statement as;

`token.assert_called_with()`

EquiTrack/azure_graph_api/tests/test_client.py
```diff
@@ -0,0 +1,67 @@
...
+        token = get_token()
+        assert token == 't0k3n'
+
+    @responses.activate
+    def test_get_token_ko(self):
+        responses.add(
+            responses.POST, settings.AZURE_TOKEN_URL,
+            json={'access_token': 't0k3n'}, status=300
+        )
+        with self.assertRaises(AzureHttpError):
```
reinbach: If you know the message that will be raised as well, I think it is a good idea to test against that, to make sure you're testing the actual exception you expect;

`with self.assertRaisesRegexp(AzureHttpError, "part of message you expect"):`

this applies to other `self.assertRaises`

EquiTrack/azure_graph_api/client.py
domdinicola: sure

EquiTrack/azure_graph_api/tests/test_utils.py
domdinicola: correct!

EquiTrack/azure_graph_api/utils.py
domdinicola: I don't know which is the mapping for those... I will remove the comment and just log Attribute #number

EquiTrack/azure_graph_api/tests/test_client.py
domdinicola: :)

EquiTrack/users/tasks.py
domdinicola: :)

EquiTrack/azure_graph_api/client.py
domdinicola: good idea

EquiTrack/azure_graph_api/client.py
domdinicola: i will set the timeout as 3600

EquiTrack/azure_graph_api/client.py
domdinicola: yes!

EquiTrack/azure_graph_api/client.py
domdinicola: sure

EquiTrack/azure_graph_api/client.py
domdinicola: yes!

EquiTrack/azure_graph_api/tasks.py
domdinicola: correct

EquiTrack/azure_graph_api/tasks.py
domdinicola: didn't know that, nice :+1:

EquiTrack/azure_graph_api/utils.py
domdinicola: ops!

EquiTrack/vision/adapters/partner.py
domdinicola: fine :+1:

EquiTrack/azure_graph_api/client.py
```diff
@@ -0,0 +1,75 @@
...
+            'grant_type': 'client_credentials',
+            'client_id': settings.AZURE_CLIENT_ID,
+            'client_secret': settings.AZURE_CLIENT_SECRET,
+            'resource': settings.AZURE_GRAPH_API_BASE_URL
+        }
+        response = requests.post(path, post_dict)
+        if response.status_code == 200:
+            jresponse = response.json()
+            token = jresponse['access_token']
+            cache.set(AZURE_GRAPH_API_TOKEN_CACHE_KEY, token, 3600)
```
vkurup: Could we add a comment here explaining why we've chosen 3600? Maybe something like: 'Cache token for 3600 seconds, which matches the default Azure token expiration'

EquiTrack/EquiTrack/settings/base.py
```diff
@@ -560,6 +560,12 @@ def str2bool(value):
     'EMAIL_FOR_USER_RESPONSIBLE_FOR_INVESTIGATION_ESCALATIONS', 'integrity1@unicef.org'
 )
 
+AZURE_CLIENT_ID = os.getenv('AZURE_CLIENT_ID', 'invalid_azure_client_id')
+AZURE_CLIENT_SECRET = os.getenv('AZURE_CLIENT_SECRET', 'invalid_azure_client_secret')
+AZURE_TOKEN_URL = 'https://login.microsoftonline.com/unicef.org/oauth2/token'
+AZURE_GRAPH_API_BASE_URL = 'https://graph.microsoft.com'
```
dpoirier: If this URL is never going to change (it looks like a fixed URL), does it need to be a setting? Could it just be a constant somewhere instead?

EquiTrack/users/templates/admin/users/country/change_form.html
```diff
@@ -1,10 +1,10 @@
 {% extends "admin/change_form.html" %}{% load i18n %}
 {% block object-tools-items %}
     {{ block.super }}
-    <li><a href="{% url 'admin:users_country_fund_commitment' original.id %}" class=''>{% trans "Sync Fund Commitment" %}</a>
-    <li><a href="{% url 'admin:users_country_fund_reservation' original.id %}" class=''>{% trans "Sync Fund Reservation" %}</a>
-    <li><a href="{% url 'admin:users_country_partners' original.id %}" class=''>{% trans "Sync Partners" %}</a>
-    <li><a href="{% url 'admin:users_country_programme' original.id %}" class=''>{% trans "Sync Programme" %}</a>
-    <li><a href="{% url 'admin:users_country_ram' original.id %}" class=''>{% trans "Sync RAM" %}</a>
+    <li><a href="{% url 'admin:users_country_fund_commitment' original.id %}" class=''>{% trans "Sync Fund Commitment" %}</a></li>
```
dpoirier: Any particular reason for these empty class attributes?

EquiTrack/vision/adapters/partner.py
```diff
@@ -178,8 +178,8 @@ def _partner_save(processed, partner):
 
                 processed += 1
 
-            except Exception as exp:
-                logger.exception(u'Exception occurred during Partner Sync: {}'.format(exp.message))
+            except Exception:
+                logger.exception(u'Exception occurred during Partner Sync %s', partner['VENDOR_CODE'])
```
dpoirier: Thank you!

EquiTrack/users/templates/admin/users/country/change_form.html
domdinicola: I'll clean that

EquiTrack/EquiTrack/settings/base.py
domdinicola: I think is better to keep a settings for now, maybe we can change it at some point

EquiTrack/vision/adapters/partner.py
domdinicola: :)

PR #1192, opened 2018-02-12
Intervention amendments related - reenable & rewrite skipped tests

csabadenes:


ntrncic:
logic has cganged., these test are no longer valid.

PR #1194, opened 2018-02-15
Auditor portal updates

pitsevich:
Auditor portal updates:
- identity management
- ability to switch workspaces for auditors
- multiple relatively minor change requests changes


roman-karpovich:
@ntrncic yes, we still need it. custom model was removed and we now use prp-like scheme with custom view because drfpaswordless is unusable out of box with session authorization.

PR #1201, opened 2018-02-20
pmp changes for hact

domdinicola:


domdinicola:
@csabadenes can you help here with some tests on serializers?

PR #1203, opened 2018-02-20
Clean up EquiTrack.mixins

vkurup:
Connects unicef/etools-issues#650

Probably easiest to review commit-by-commit.


EquiTrack/partners/models.py
```diff
@@ -699,6 +699,10 @@ def audits_completed(cls, partner, update_one=False):
         partner.hact_values['audits']['completed'] = completed_audit
         partner.save()
 
+    def get_admin_url(self):
```
domdinicola: :100:

ntrncic:
@vkurup please update tests as factory cleanup PR has been merged

EquiTrack/EquiTrack/tests/test_middleware.py
```diff
@@ -0,0 +1,89 @@
...
+        "Superuser in inactive workspace is allowed to pass."
+        superuser = UserFactory(is_superuser=True)
+        self.request.user = superuser
+        self.assertEquals(EToolsTenantMiddleware().process_request(self.request), None)
+
+    @override_settings(PUBLIC_SCHEMA_URLCONF='foo')
+    def test_public_schema_urlconf(self):
+        """
+        This just tests a code path that was copied from the django-tenant-schemas middleware when we
+        copy/pasted. eTools does not use it.
```
dpoirier: If we're not using this code path, should we just get rid of it?

EquiTrack/EquiTrack/middleware.py
```diff
@@ -12,32 +18,43 @@
 
 logger = logging.getLogger(__name__)
 
+ANONYMOUS_ALLOWED_URL_FRAGMENTS = [
```
dpoirier: We could use this in the test too.

EquiTrack/EquiTrack/middleware.py
```diff
@@ -12,32 +18,43 @@
...
         if request.user.is_anonymous():
             # check if user is trying to reach an authentication endpoint
-            if any(x in request.path for x in [
-                u'api',
-                u'login',
-                u'saml',
-                u'accounts',
-                u'monitoring',
-            ]):
+            if any(fragment in request.path
```
dpoirier: I guess it's always been like this, but can we be sure no other possible URL paths will ever include any of these substrings anywhere?  Would it be safer to check that the request.path starts with e.g. "api/" or "login/"?

EquiTrack/EquiTrack/tests/test_middleware.py
vkurup: Good question. I don't think we're using it, but OTOH, it is a code path which is a documented feature of django-tenant-schemas: http://django-tenant-schemas.readthedocs.io/en/latest/install.html#PUBLIC_SCHEMA_URLCONF

So, if we were to remove it, and later someone expected it to work based on the library's documentation, they might be surprised to find it broken. I think a better long term solution would be to refactor our middleware so that it extends the library's middleware (rather than copy/pasting it), but I didn't see an easy way to do that, so punted on that for now. I could create a Clubhouse issue to clean this up?

EquiTrack/EquiTrack/middleware.py
vkurup: Good point!

EquiTrack/EquiTrack/middleware.py
vkurup: Yes, this is definitely a risk that we need to fix, but I wasn't comfortable making that change in this PR. For the most part, I think changing this to only match the first part of the URL makes sense, but there is at least one URL path that looks like /users/api/ that we'd also have to catch. I'll create a Clubhouse issue for this as well.

vkurup:
I believe the test failures on CircleCI are related to tests which are relying on an inconsistent ordering, which @dpoirier has fixed in #1266.

EquiTrack/partners/mixins.py
```diff
@@ -10,3 +13,35 @@ def formfield_for_foreignkey(self, db_field, request=None, **kwargs):
         return super(HiddenPartnerMixin, self).formfield_for_foreignkey(
             db_field, request, **kwargs
         )
+
+
+class CountryUsersAdminMixin(object):
```
ntrncic: do we need this anymore?

vkurup: It is used by the InterventionAdmin and the AgreementAdmin. Can I remove it from those places? I don't know offhand if it would break functionality or not...

ntrncic: can you check if it breaks anything?

vkurup: It doesn't break anything if we remove it, but it does allow you to choose non-staff users and non-country users when creating or editing Agreements or Interventions via the admin.

So, as an example, you could create an Agreement with a `signed_by` user who was not a staff user, and was not associated with the same country of the Agreement you were creating. If we don't care about validating that kind of thing via the admin, then we could remove it.

@ntrncic ^

PR #1210, opened 2018-02-22
Remove commented code

vkurup:
Connects unicef/etools-issues#639

So I went through and removed 'commented code' that I thought could be removed (using something like `grep '^[ ]*#` across the project, and looking for matches that looked like code).

For almost all of these, however, I'm not 100% sure that there isn't some 'issue' that needs to be captured as a TODO when we delete this, so let me know if there is, and we can create new issues to track what needs to be done.


ntrncic:
please label the branch

EquiTrack/t2f/serializers/travel.py
```diff
@@ -537,15 +537,12 @@ class Meta:
 
     def get_status(self, obj):
```
domdinicola: do we need a serializer method here?

vkurup: Hmm... good question. I only focused on removing commented code in this PR. But you're right... I think I could change the serializer field classes to be `ReadOnlyField()` and then remove the methods.

EquiTrack/t2f/serializers/travel.py
```diff
@@ -527,23 +527,14 @@ class Meta:
 class TravelActivityByPartnerSerializer(serializers.ModelSerializer):
     locations = serializers.SlugRelatedField(slug_field='name', many=True, read_only=True)
     primary_traveler = serializers.CharField(source='primary_traveler.get_full_name')
-    reference_number = serializers.SerializerMethodField()
-    status = serializers.SerializerMethodField()
-    trip_id = serializers.SerializerMethodField()
+    reference_number = serializers.ReadOnlyField()
```
domdinicola: :100:

EquiTrack/partners/urls_v2.py
```diff
@@ -176,8 +163,6 @@
         view=InterventionReportingPeriodDetailView.as_view(http_method_names=['get', 'patch', 'delete']),
         name='intervention-reporting-periods-detail'),
 
-    # TODO: figure this out
```
ntrncic: leave this in please

EquiTrack/partners/serializers/prp_v1.py
```diff
@@ -88,7 +88,6 @@ class Meta:
         model = Disaggregation
```
ntrncic: leave this entire file untouched as it's still in progress by @robertavram

EquiTrack/partners/permissions.py
```diff
@@ -98,17 +98,9 @@ class InterventionPermissions(PMPPermissions):
     def __init__(self, **kwargs):
         '''
         :param kwargs: user, instance, permission_structure
-        if 'inbound_check' key, is sent in, that means that instance now contains all of the fields available in the
```
ntrncic: leave this code in please

EquiTrack/partners/urls_v2.py
vkurup: OK, but a more expressive comment might let someone actually help figure it out :D

ntrncic:
@vkurup please resolve conflicts

PR #1211, opened 2018-02-26
MOU in past status transition changes

csabadenes:
Relates to https://app.clubhouse.io/unicefetools/story/1556/mou-cannot-change-status-of-an-mou-with-end-date-in-the-past-to-signed-alm-305


EquiTrack/partners/validation/agreements.py
```diff
@@ -31,7 +31,7 @@ def agreement_transition_to_signed_valid(agreement):
     if not agreement.start or agreement.start > today:
         raise TransitionError(['Agreement cannot transition to signed until '
                                'the start date is less than or equal to today'])
-    if not agreement.end or agreement.end < today:
+    if not agreement.end or (agreement.agreement_type != agreement.MOU and agreement.end < today):
```
robertavram: This should not have anything to do with MOU, transition to signed should be available for all agreements that meet the signed criteria. (read through the comments of the issue raised). we can discard all logic specific to MOU and remove the end date validation to be greater than today.

instead of the original: `if not agreement.end or agreement.end < today:` we can just check if end date is populated `if not agreement.end:`.

The `agreement_transition_to_ended_valid` function will check the rest...
```
AUTO_TRANSITIONS = {
        DRAFT: [SIGNED],
        SIGNED: [ENDED],
    }
```
is supposed to try transitioning to ended immediately after...

Things to check and test...
make sure an agreement of any kind will transition all the way to ended if it's in status Draft and all data is saved that would fulfill all the criteria to ended.

csabadenes: @robertavram, i might have misunderstood the last comment from https://github.com/unicef/etools-issues/issues/840, i'll ask you again tomorrow:)
I will do the change, and i have tested this quite extensively, i think it works as expected in the current form. 
(There also was another issue, not having 'authorized_officers' was blocking the auto transition of MOU's)

PR #1231, opened 2018-03-06
Add an extra validation to intervention amendments

csabadenes:
Do not allow adding new  amendments while intervention is already in amendent.


EquiTrack/partners/serializers/interventions_v2.py
```diff
@@ -54,6 +55,17 @@ class Meta:
         model = InterventionAmendment
         fields = "__all__"
 
+    def validate(self, data):
+        data = super(InterventionAmendmentCUSerializer, self).validate(data)
+
+        if 'signed_date' in data and data['signed_date'] > date.today():
```
csabadenes: Question: is this validation OK? (I have added an extra check to the backend for a frontend-only validation rule)

robertavram: 👍

PR #1235, opened 2018-03-06
DO NOT MERGE: Prp refactoring auth poc2 - AD integration

chachra:


reinbach:
The `AZURE_B2C_CLIENT_ID` and `AZURE_B2C_CLIENT_SECRET` environment vars needs to be set in the backend docker instance. I passed them in the devup fab func;

eg:
```
def devup(quick=False, DB_PORT="51322"):
    key = '...'
    secret = '...'
    with shell_env(DB_PORT=DB_PORT, AZURE_B2C_CLIENT_ID=key, AZURE_B2C_CLIENT_SECRET=secret):
        ....
```

PR #1237, opened 2018-03-07
added outstanding_findings to hact json

domdinicola:


EquiTrack/partners/models.py
```diff
@@ -550,19 +550,8 @@ def hact_min_requirements(self):
...
-        ff = audits.filter(financial_findings__isnull=False).aggregate(
-            total=Coalesce(Sum('financial_findings'), 0))['total']
-        ar = audits.filter(amount_refunded__isnull=False).aggregate(
-            total=Coalesce(Sum('amount_refunded'), 0))['total']
-        asdp = audits.filter(additional_supporting_documentation_provided__isnull=False).aggregate(
-            total=Coalesce(Sum('additional_supporting_documentation_provided'), 0))['total']
-        wor = audits.filter(write_off_required__isnull=False).aggregate(
-            total=Coalesce(Sum('write_off_required'), 0))['total']
-        return ff - ar - asdp - wor
+        hact = json.loads(self.hact_values) if isinstance(self.hact_values, str) else self.hact_values
```
vkurup: Looks like self.hact_values is a JSONField, so I don't understand why we need to do this (rather than just using `hact = self.hact_values`)

domdinicola: i think is related to the comment line 461: # JSONFIELD has an issue where it keeps escaping characters

vkurup: Ahh, I see. Looks like Rob added that comment over a year ago. Might be worth creating a separate issue to see if we can remove that comment (and associated workarounds) now that we're using a more modern Django. 

Definitely not a blocker for this PR.

domdinicola: I agree with you, maybe the json field at the time was still outside Django.
An alternative approach would be to create a property to centralize the retrieve of json field, instead of do if that always.

EquiTrack/EquiTrack/settings/base.py
```diff
@@ -309,7 +309,7 @@ def str2bool(value):
 CELERY_BROKER_URL = os.environ.get('REDIS_URL', 'redis://localhost:6379/0')
 CELERY_BROKER_VISIBILITY_VAR = os.environ.get('CELERY_VISIBILITY_TIMEOUT', 1800)  # in seconds
 CELERY_BROKER_TRANSPORT_OPTIONS = {'visibility_timeout': int(CELERY_BROKER_VISIBILITY_VAR)}
-CELERY_RESULT_BACKEND = 'django_celery_results.backends.database:DatabaseBackend'
+CELERY_RESULT_BACKEND = 'django-db'
```
robertavram: try this: 'db+' + os.getenv('DATABASE_URL', 'invalid_url')

EquiTrack/partners/models.py
robertavram: 👍

robertavram: I think in order to retrieve it, it's not a problem.. but on save (even admin) it used to get transformed in a string... we can check sometime

robertavram: let's do this on the serializer itself...
or tell Marko to use the jsonfield

PR #1239, opened 2018-03-08
Merge base test cases

reinbach:
Merged EToolsTenantTestCase and APITenantTestCase, no need to determine which one to use for new tests


EquiTrack/EquiTrack/tests/cases.py
```diff
@@ -97,3 +101,38 @@ def assertKeysIn(self, keys, container, msg=None, exact=False):
...
+
+        :param method: the HTTP method 'get'/'post'
+        :type method: str
+        :param url: the relative url to the base domain
+        :type url: st
+        :param user: optional user if not authenticated as the current user
+        :type user: django.contrib.auth.models.User
+        :param data: any data that should be passed to the API view
+        :type data: dict
+        """
```
vkurup: I know this docstring is just copy/pasted, but can we make a couple improvements?
1. s/st/str on line 112
2. document that if `user` is not provided, then `self.user` will be used.
3. document that if `view` is provided, then the view function will be called directly, rather than resolving the URL.

I don't think we're publishing these docstrings anywhere, so I'd be fine with removing the "javadoc"-style items, if that makes it easier.

reinbach: :+1:

vkurup: Thank you! :shipit:

ntrncic:
please resolve conflicts

PR #1243, opened 2018-03-08
 Allow termination of same CP duplicate agreements older than 2015 july

csabadenes:
Connects to https://app.clubhouse.io/unicefetools/story/4428/unable-to-terminate-the-pca-amongst-two-pcas-whose-start-date-is-before-july-2015-alm-508

I'm not 100% sure this is the right solution. The agreement validator wants to prevent creating duplicate PCA agreements per CP, but allows to create unlimited nr. of agreements if all of them are before 2015 July.

### DIFF
--- a/EquiTrack/partners/validation/agreements.py
+++ b/EquiTrack/partners/validation/agreements.py
@@ -119,6 +119,7 @@ def ssfa_static(agreement):
 
 def one_pca_per_cp_per_partner(agreement):
     if agreement.agreement_type == agreement.PCA:
+        #Allow the termination of duplicated agreements started before 2015 July
         if agreement.status == agreement.TERMINATED and agreement.start <= date(2015, 7, 1):
             return True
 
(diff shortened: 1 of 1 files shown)


EquiTrack/partners/validation/agreements.py
```diff
@@ -119,6 +119,10 @@ def ssfa_static(agreement):
 
 def one_pca_per_cp_per_partner(agreement):
     if agreement.agreement_type == agreement.PCA:
+        #Allow the termination of duplicated agreements started before 2015 July
```
robertavram: hmmm... i feel like this is a validation that needs to run only if a new agreement is created... one_pca_per_cp_per_partner basic validation should only check when trying to add a draft one...
I don't think the original function was working as needed.

We need to check if the desire is that `we don;t allow creation of draft PCAs in a CP with a Partner that already has a PCA that's in the same CP`    OR   `we allow creation of PCAs in draft and don't allow them to transition to signed unless they're the only PCA / CP / Partner`

either way, this basic validation gets executed on any change of any data in any PCA, therefore will prevent anything from changing in any PCA if a different PCA exists for the same Partner , same CP. This approach i believe it to be flawed.  Good find here @csabadenes !

robertavram:
addressing this in another PR:
https://github.com/unicef/etools/pull/1271

PR #1246, opened 2018-03-08
Revert "CI: less aggressive cache-busting"

robertavram:
Reverts unicef/etools#1240

Throwing errors when testing on other branches randomly... weird

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,28 +10,28 @@ jobs:
           DATABASE_URL: "postgis://postgres:postgres@localhost:5432/circle_test"
       - image: circleci/postgres:9.5-alpine-postgis
         environment:
-          POSTGRES_USER: postgres
+          POSTGRES_USER: postgres 
           PGUSER: postgres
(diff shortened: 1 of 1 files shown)


vkurup:
Hey @robertavram, I'm trying to look into this. I've looked at every failed build since #1240 was pushed, and I've found three that I think are random: [7227](https://circleci.com/gh/unicef/etools/7227), [7234](https://circleci.com/gh/unicef/etools/7234), and [7249](https://circleci.com/gh/unicef/etools/7249).

Of those, only 7234 is using the 'less aggressive cache-busting config'. Both 7227 and 7249 are actually using the old config.

So, I think this might be unrelated. But it IS strange. The PR incidentally changed whitespace on a few lines, and those lines happened to be postgres environment variables for the postgis docker image. Maybe that caused the postgis docker image to have to rebuild, and it didn't build fast enough to be accepting connections?

Let me know if I missed any other failed builds.

Possible next steps:
1. Try again, without the whitespace changes
2. Give up for now :)
3. Something else?

PR #1252, opened 2018-03-12
Cleanup utils

reinbach:
The utils app had a number of "sub" apps and was an amalgation of a number of classes/functions.
We also have a EquiTrack.utils section.

I attempted to separate the various parts of the utils app into specific apps (eg: rest_extra, permissions, and generics) and to move the smaller parts that did not fit into a separate app to the EquiTrack.utils section.

Hopefully this will make it easier to create libraries from these new apps, as well as to be able to break out other apps into libraries in an easier fashion (eg: attachments relies on generics)

I may have placed various classes/functions in the wrong app/section, let me know what should be fixed, or whether this is going in the wrong direction.


EquiTrack/EquiTrack/utils/__init__.py
```diff
@@ -0,0 +1,127 @@
...
+
+def get_environment():
+    return settings.ENVIRONMENT
+
+
+def get_current_site():
+    return Site.objects.get_current()
+
+
+class every_country:
```
dpoirier: Class names should be CamelCase.

dpoirier:
Aside: I can't help noticing how few of the utility methods have docstrings to say what they're actually for.

EquiTrack/rest_extra/fields.py
```diff
@@ -4,7 +4,7 @@
 from rest_framework.utils import model_meta
 from rest_framework_recursive.fields import RecursiveField
 
-from utils.writable_serializers.serializers import WritableListSerializer
+from rest_extra.serializers import WritableListSerializer
 
 
 class builtin_field:
```
dpoirier: Could we CamelCase this while we're here?

EquiTrack/EquiTrack/settings/base.py
```diff
@@ -181,10 +181,9 @@ def str2bool(value):
     'audit.purchase_order',
     'EquiTrack',
     'tpm.tpmpartners',
-    'utils.common',
-    'utils.mail',
```
dpoirier: Still need 'utils.mail' for the 'load_emails' management command.

reinbach: I decided on merging EquiTrack.utils.mail more into EquiTrack app instead, seemed like a better idea to me, rather than having sub apps etc.

domdinicola:

EquiTrack/EquiTrack/utils/scripts.py
```diff
@@ -31,15 +30,15 @@ def printtf(*args):
         f.write('\n')
 
 
-def set_country(name):
+def set_country_by_name(name):
```
saxix: If I have understood correctly, this function is called many times, but should have the save value for each request, why do not decorate with `lru_cache(1)` ?

EquiTrack/EquiTrack/validation_mixins.py
```diff
@@ -13,7 +13,7 @@
 from rest_framework.exceptions import ValidationError
 
 from EquiTrack.parsers import parse_multipart_data
-from utils.common.utils import get_all_field_names
+from EquiTrack.utils import get_all_field_names
```
saxix: I'm not sure we should put generic "utilities" functions under specific application namespace. In this case `get_all_field_names` seems more appropriate in a generic utilities library. I expect to have into Equitrack only application specific utilities, and any utility not related to etools and that can be used everywhere in a dedicated "decoupled" package

EquiTrack/partners/models.py
```diff
@@ -23,7 +23,7 @@
 from dateutil.relativedelta import relativedelta
 
 from EquiTrack.fields import CurrencyField, QuarterField
-from EquiTrack.utils import import_permissions, get_quarter, get_current_year
+from EquiTrack.utils import get_current_year
```
saxix: same as above. If we want start creating common reusable libraries, we should group generic utilities in a package that is not coupled with any specific application.

EquiTrack/EquiTrack/validation_mixins.py
reinbach: I agree that it would be good to get some of these out into a separate package that can be used by various apps. 

This is the first step in merging a number of "utils" sections, and especially moving utils out from the same level as the apps, as a number of the apps have their own utils file and there was starting to be a clash with absolute and relative imports. The next step would be to start creating libraries of similar code.

I consider EquiTrack as the actual project base and has functionality to be used across all apps, eg: base urls, wsgi, base test cases and mixins.

saxix: make sense, let's start with this one.

domdinicola:
@reinbach I don't think we will merge this...
actually we should start thinking about moving some utilities to https://github.com/unicef/unicef-djangolib
mainly code we think can be reused in other projects

reinbach:
I agree that we want to move code out into packages that can be reused. 
This was meant to be a step in grouping code into parts that could be packaged. ie `rest_extra`, `permissions` etc instead of having a catch all like `utils`.

I'd much prefer to have specific packages that have a specific purpose, instead of a catch all.

PR #1255, opened 2018-03-12
Use b-prefix only when appropriate

dpoirier:
Writing text strings as b'text' is equivalent to 'text'
in Python 2, but in Python 3 has a very different meaning.
Only use b'' when we really intend a byte string.

(Django's migration generator seems to have been by far
the worst offender here, strangely enough.)


ntrncic:
pleasew resolve conflicts. sorry

PR #1256, opened 2018-03-12
Add link to Country from vision log

domdinicola:


EquiTrack/vision/templates/admin/vision/vision_log/change_form.html
```diff
@@ -0,0 +1,6 @@
+{% extends "admin/change_form.html" %}{% load i18n %}
+{% block object-tools-items %}
+    {{ block.super }}
+    <li><a href="{% url 'admin:users_country_change' original.country.id %}" class=''>{% trans "Go to Country" %}</a></li>
```
reinbach: Curious why we have `class=''`, I see that we have that in the extended template as well

domdinicola: @reinbach removed and enhanced

EquiTrack/users/templates/admin/users/country/change_form.html
```diff
@@ -1,10 +1,17 @@
...
+
+    <li><a target='_blank' href="https://devapis.unicef.org/BIService/BIWebService.svc/GetPartnerDetailsInfo_xml/{{original.business_area_code}}" style="background-color:#96c525">{% trans "API Partner" %}</a></li>
+    <li><a target='_blank' href="https://devapis.unicef.org/BIService/BIWebService.svc/GetProgrammeStructureList_xml/{{original.business_area_code}}" style="background-color:#96c525">{% trans "API Programme" %}</a></li>
+    <li><a target='_blank' href="https://devapis.unicef.org/BIService/BIWebService.svc/GetRAMInfo_xml/{{original.business_area_code}}" style="background-color:#96c525">{% trans "API RAM" %}</a></li>
+    <li><a target='_blank' href="https://devapis.unicef.org/BIService/BIWebService.svc/GetFundsReservationInfo_xml/{{original.business_area_code}}" style="background-color:#96c525">{% trans "API Fund Reservation" %}</a></li>
+    <li><a target='_blank' href="https://devapis.unicef.org/BIService/BIWebService.svc/GetFundsCommitmentInfo_xml/{{original.business_area_code}}" style="background-color:#96c525">{% trans "API Fund Commitment" %}</a></li>
```
dpoirier: Since we're setting all these backgrounds to one of two colors, are there CSS classes we could use instead of inline styles? (Maybe create new classes if need be, but hopefully we're setting these to colors that are already in use elsewhere, and so there are already classes for them.)

ntrncic: @dpoirier thanks for the heads up. we'll only use this here.

PR #1258, opened 2018-03-12
Exception.message is gone in Python3

dpoirier:
We can use Exception.args[0] to get the first argument
to the exception's constructor instead.


dpoirier:
Hold off on merging this. I rethought this overnight and I don't think it's going to be safe to assume .args has at least one member.  (e.g. someone could write "raise ValueError" in which case .args will be empty.)

dpoirier:
@reinbach I've updated this with a better fix.

domdinicola:
@dpoirier please resolve conflicts

domdinicola:
@dpoirier conflicts...

EquiTrack/vision/vision_data_synchronizer.py
```diff
@@ -121,9 +120,9 @@ def sync(self):
             totals = self._save_records(converted_records)
 
         except Exception as e:
-            logger.info('sync caught {} with message "{}"'.format(type(e).__name__, e.message))
-            log.exception_message = e.message
-            raise VisionException(message=e.message), None, sys.exc_info()[2]
+            logger.info('sync', exc_info=True)
+            log.exception_message = force_text(e)
+            raise VisionException(force_text(e)), None, sys.exc_info()[2]
```
domdinicola: is this py3 compatible?

dpoirier: Maybe not, I'll change it to six.reraise.

EquiTrack/vision/adapters/manual.py
```diff
@@ -116,9 +116,9 @@ def _process_record(self, json_item):
                 model.objects.update_or_create(
                     defaults=defaults, **kwargs
                 )
-        except Exception as exp:
-            logger.warning("Exception message: {}".format(exp.message))
-            logger.warning("Exception type: {}".format(type(exp)))
+        except Exception:
+            logger.warning('Exception processing record in MultiModelDataSynchronizer',
```
ntrncic: Don't understand what `MultiModelDataSynchronizer` means. Can we reword this exception

EquiTrack/vision/adapters/partner.py
```diff
@@ -188,8 +188,9 @@ def _partner_save(self, partner):
 
             processed = 1
 
-        except Exception as exp:
-            logger.exception(u'Exception occurred during Partner Sync: {}'.format(exp.message))
+        except Exception:
+            logger.exception(u'Exception occurred during Partner Sync')
```
ntrncic: does the exception message get printed out?

EquiTrack/vision/adapters/manual.py
dpoirier: What should it say? I don't know this code, I was just updating the exception code and trying to make the log message a little more useful.

EquiTrack/vision/adapters/partner.py
dpoirier: It gets logged, just like before.

EquiTrack/vision/adapters/manual.py
domdinicola: MultiModelDataSynchronizer doesn't make too much. Can we use the self.__class__ or remove it

dpoirier: MultiModelDataSynchronizer removed

PR #1263, opened 2018-03-13
Init Result Type command

domdinicola:


robertavram:
management command would be much better

EquiTrack/reports/tests/test_commands.py
```diff
@@ -0,0 +1,14 @@
...
+from EquiTrack.tests.cases import EToolsTenantTestCase
+from reports.models import ResultType
+
+
+class TestResultTypeCommand(EToolsTenantTestCase):
+
+    def test_command(self):
+        self.assertEqual(ResultType.objects.count(), 0)
+        call_command('init-result-type')
+        self.assertEqual(ResultType.objects.count(), 3)
```
dpoirier: Maybe also test that the 'global' country doesn't get its result types initialized, and that if we pass in a schema, only that country gets initialized?

PR #1265, opened 2018-03-13
Interventions in amendment should not be suspendable

csabadenes:


csabadenes:
Connects to https://app.clubhouse.io/unicefetools/story/4467/issue-in-amendment-flag

EquiTrack/partners/validation/interventions.py
```diff
@@ -290,6 +290,8 @@ def state_signed_valid(self, intervention, user=None):
     def state_suspended_valid(self, intervention, user=None):
         self.check_required_fields(intervention)
         self.check_rigid_fields(intervention, related=True)
+        if intervention.in_amendment is True:
```
robertavram: can you add this validations to the transitions rather than to stateful validation? @csabadenes

csabadenes: @robertavram sure.

csabadenes:
@robertavram i have made the requested changes

EquiTrack/partners/tests/test_validation_interventions.py
```diff
@@ -340,6 +342,34 @@ def test_valid(self):
         self.assertTrue(transition_to_active(intervention))
 
 
+class TestTransitionToSuspended(EToolsTenantTestCase):
+    def test_intervention_in_amendment(self):
+        """Interventions in amendment cannot be suspended"""
+        suspendable_statuses = [Intervention.SIGNED, Intervention.ACTIVE]
```
dpoirier: It seems like a better name for this variable would be "unsuspendable_statuses", wouldn't it?  These seem to be the statuses for which trying to suspend the intervention results in an error.

EquiTrack/partners/tests/test_validation_interventions.py
```diff
@@ -340,6 +342,34 @@ def test_valid(self):
...
+
+        for suspendable_status in suspendable_statuses:
+            intervention = InterventionFactory(
+                status=suspendable_status,
+                in_amendment=True,
+            )
+            with self.assertRaises(TransitionError):
+                transition_to_suspended(intervention)
```
dpoirier: Can we also test that trying to suspend interventions in other statuses still works?

EquiTrack/partners/tests/test_validation_interventions.py
```diff
@@ -340,6 +342,34 @@ def test_valid(self):
...
+        """Interventions in amendment cannot be terminated"""
+        terminable_statuses = [Intervention.SIGNED, Intervention.ACTIVE]
+
+        for terminable_status in terminable_statuses:
+            intervention = InterventionFactory(
+                status=terminable_status,
+                in_amendment=True,
+            )
+            with self.assertRaises(TransitionError):
+                transition_to_terminated(intervention)
```
dpoirier: Same comments as before about naming the list of statuses that result in errors, and testing that other status don't result in errors.

EquiTrack/partners/tests/test_validation_interventions.py
csabadenes: @dpoirier Afaik these are the two statuses where suspend is possible. It's just not possible while the intervention is in amendment mode.

EquiTrack/partners/tests/test_validation_interventions.py
csabadenes: @dpoirier Yes, probably i should have done that. i will update the PR.

EquiTrack/partners/tests/test_validation_interventions.py
csabadenes: @dpoirier i believe i have addressed your comments somewhat.. but for now, i have added a skip to the test that checks transition from other statuses than ACTIVE and SIGNED. The backend checks aren't very strict right now, and allow transition to SUSPENDED and TERMINATED from other statuses(afaik, transition to SUSPENDED and TERMINATED should be possible only manually, from ACTIVE and SIGNED).
@robertavram @ntrncic, the transition to SUSPENDED and TERMINATED from other statuses  than ACTIVE and SIGNED shouldn't be disabled?

EquiTrack/partners/tests/test_validation_interventions.py
dpoirier: I'm confused. The test appears to exist to prove that suspend is *not* possible in these two statuses, right?

EquiTrack/partners/tests/test_validation_interventions.py
dpoirier: I'm not familiar with this application, but I would think in general testing that the application doesn't allow business rules to be violated would be good.

EquiTrack/partners/tests/test_validation_interventions.py
csabadenes: The test exists to see if an intervention *is* suspendable in these two statuses, and *is not* suspendable in the same two statuses when in_amendment is True

EquiTrack/partners/tests/test_validation_interventions.py
csabadenes: As far as i understand the business rules, suspending the interventions in other statuses should not work, but currently it is allowed in the backend, so it is not testable..
T have asked for clarification in a comment below.

EquiTrack/partners/tests/test_validation_interventions.py
dpoirier: But the test asserts that any attempt to suspend from those statuses results in a TransitionError, not that it works.

csabadenes: It should result in a TransitionError only after we set the in_amendment flag to True. Hmm are you checking the updated code?

ntrncic: @csabadenes yes this is good. it should result in error only in amendment

EquiTrack/partners/tests/test_validation_interventions.py
ntrncic: @csabadenes  yes only from Active and Signed should transition to suspended and terminated. if the other statuses are allowed we should address it in another PR

PR #1266, opened 2018-03-13
Small Python3 changes

dpoirier:
Some small changes for Python3 that didn't seem worth separate pull
requests.

* The syntax for MetaClasses changed. Use six.add_metaclass().

* Avoid bare exceptions. (Not sure if this is Python 3 specific, but
  flake was complaining.)

* Files and responses contain binary data, not text.

* In Python 3, we cannot compare (less/greater) a None value to
  anything. Make sure we have an int (0) instead.

* The long integer type is gone. six.integer_types contains
  the valid integer types for the current python version.

* In Python 2 you can call .next() on an iterator, while in
  Python 3, you call the built-in next() and pass in an
  iterator.  For compatibility, use six.next(iterator).

* Relative imports v. absolute imports.


EquiTrack/users/urls_v3.py
```diff
@@ -1,6 +1,7 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
 from django.conf.urls import url
 
-from views_v3 import (
+from .views_v3 import (
```
domdinicola: would you mind to use absolute import here

EquiTrack/users/serializers_v3.py
```diff
@@ -1,8 +1,8 @@
-from __future__ import unicode_literals
+from __future__ import absolute_import, division, print_function, unicode_literals
 
 from django.contrib.auth import get_user_model
 from rest_framework import serializers
-from serializers import GroupSerializer, SimpleCountrySerializer
+from .serializers import GroupSerializer, SimpleCountrySerializer
```
domdinicola: would you mind to use absolute import here

dpoirier:
@domdinicola updated

EquiTrack/partners/views/partner_organization_v2.py
```diff
@@ -392,7 +392,7 @@ def delete(self, request, *args, **kwargs):
                                   "against this partner. The Partner record cannot be deleted")
         elif TravelActivity.objects.filter(partner=partner).count() > 0:
             raise ValidationError("This partner has trips associated to it")
-        elif partner.total_ct_cp > 0:
+        elif (partner.total_ct_cp or 0) > 0:
```
ntrncic: why do we need `or 0` everywhere?

reinbach: I believe, in case the parameter is `None` and in py3 a comparison like this (`None > 0`) results in a `TypeError`

dpoirier: Exactly. Do you think we need a comment?

dpoirier:
@ntrncic conflicts resolved

PR #1273, opened 2018-03-19
API Dropdown sorted alphabetically by value: 1880

domdinicola:


EquiTrack/EquiTrack/mixins.py
```diff
@@ -273,6 +274,28 @@ def get_renderer_context(self):
...
+                value = self.request.query_params.get(param_filter)
+                if query_filter.endswith(('__in')):
+                    value = value.split(',')
+                queries.append(Q(**{query_filter: value}))
+        return queries
+
+    def search_params(self, filters, param_name='search'):
+        search_term = self.request.query_params.get(param_name)
+        search_query = Q()
+        if param_name in self.request.query_params:
```
robertavram: why are you iterating over all the queryparams?

EquiTrack/EquiTrack/mixins.py
```diff
@@ -273,6 +274,28 @@ def get_renderer_context(self):
...
+            for filter in filters:
```
robertavram: the term "filter" is reserved, it would be good to chose something else

EquiTrack/EquiTrack/mixins.py
```diff
@@ -273,6 +274,28 @@ def get_renderer_context(self):
...
+                q = Q(**{filter: search_term})
```
robertavram: why not just use equal here, rather than unpacking a dict with 1 key

EquiTrack/EquiTrack/mixins.py
```diff
@@ -273,6 +274,28 @@ def get_renderer_context(self):
...
+                search_query = search_query | q
```
robertavram: not sure if there are any implications of doing it this way vs creating a list and "or-ing" them one by one .. I imagine the result on 3 params will be:
`(((Q() | Q(filter1=cond1)) | Q(filter2=cond2)) | Q(filter3=cond3))` rather than 
`Q() | Q(filter1=cond1) | Q(filter2=cond2) | Q(filter3=cond3)`

EquiTrack/EquiTrack/mixins.py
```diff
@@ -273,6 +274,28 @@ def get_renderer_context(self):
...
 
+class QueryStringFilterMixin(object):
+
+    def filter_params(self, filters):
+        queries = []
+        for param_filter, query_filter in filters:
+            if param_filter in self.request.query_params:
+                value = self.request.query_params.get(param_filter)
+                if query_filter.endswith(('__in')):
+                    value = value.split(',')
```
robertavram: 👍

EquiTrack/EquiTrack/mixins.py
robertavram: got it... having a brain lapse :)

PR #1274, opened 2018-03-20
str and unicode for Python 3 compatibility

dpoirier:
In Python 2, unicode(thing) is a handy idiom to make a text string
from an object, while str(thing) makes a bytestring.

In Python 3, str(thing) makes a text string, and the unicode keyword
doesn't even exist.

So to write code that works the same in Python 2 and 3, we cannot use
either str() or unicode().

Where we're currently using unicode(thing) in Python 2, we can use
six.text_type(thing) and continue getting a text string in either
version.

Where we're using str(thing) in Python 2, we might think we could use
six.binary_type(thing) - but unlike text strings, there's no built-in
method on objects in Python 3 to convert them to bytestrings.  We
probably meant unicode(thing) in most of these places anyway.  If not,
we might need to write six.text_type(thing).encode('utf-8').

A bunch of tests that were checking that both str(thing) and
unicode(thing) aren't really useful anymore, but we can still check
that six.text_type(thing) returns the text string we expect for each
model.


ntrncic:
please resolve conflict

domdinicola:
I resolved

PR #1278, opened 2018-03-21
Attachments endpoint caching

reinbach:
Attachment endpoint data is cached by tenant

Caching is cleared when attachment is saved for the tenant

`prefetch_related()` did not have any impact on the performance of the endpoint.

[ch4173]


robertavram:
It's a good option for now... but we'll have to figure out a way to populate the cache.. it's not really awesome for the first person that accesses this every day to have to wait 5 min :)

Maybe we can create a DB view to help with this from a performance standpoint... prefetch related would not work on generic relationships as far as i know...
anyone any other ideas? @vkurup @dpoirier @reinbach @ntrncic @ewheeler @domdinicola

reinbach:
The response data is what is cached, we could move that to a function that can be called from the view as well as when the copying of attachment data is completed.

Agree this is not ideal, but may mimize the initial load issue for the user, as it will be re-cached after copy is complete.

Cache is being cleared when attachment data is saved, so there is a period of time, when the copying of attachments is happening and a user makes a request prior to the cache being reset

domdinicola:
Here it's saying that it should work with generic foreign kes as well!!!
https://docs.djangoproject.com/en/1.10/ref/models/querysets/#prefetch-related

reinbach:
I didn't see any performance improvement when testing `prefetch_related()` locally. Maybe I was missing something?

domdinicola:
https://docs.djangoproject.com/en/1.10/ref/models/querysets/#prefetch-related

While prefetch_related supports prefetching GenericForeignKey relationships, the number of queries will depend on the data. Since a GenericForeignKey can reference data in multiple tables, one query per table referenced is needed, rather than one query for all the items. There could be additional queries on the ContentType table if the relevant rows have not already been fetched.

I missed this statement: prefetching objects referenced by a GenericForeignKey is only supported if the query is restricted to one ContentType.

reinbach:
I played around with denomarlizing the data instead, this removes the need for caching.

See https://github.com/unicef/etools/pull/1281

PR #1279, opened 2018-03-21
Locations gis api

csabadenes:
Add few new API endpoints for the locations GIS investigation


csabadenes:
Related to https://app.clubhouse.io/unicefetools/story/4758/locations-gis-api

EquiTrack/locations/serializers.py
```diff
@@ -75,6 +76,39 @@ class Meta:
...
+            'id',
+            'name',
+            'p_code',
+            'gateway_id',
+            'level',
+        )
+
+
+class GisLocationGeoDetailSerializer(serializers.ModelSerializer):
+    id = serializers.CharField(read_only=True)
```
domdinicola: do we need this?

EquiTrack/locations/views.py
```diff
@@ -108,3 +118,84 @@ def get_queryset(self):
             qs = qs.filter(name__istartswith=self.q)
 
         return qs
+
+
+class GisLocationsInUseViewset(ListAPIView):
+    model = Location
+    serializer_class = GisLocationListSerializer
+    permission_classes = (permissions.IsAdminUser,)
```
robertavram: isSuperUser would be better for now

EquiTrack/locations/views.py
```diff
@@ -108,3 +118,84 @@ def get_queryset(self):
             qs = qs.filter(name__istartswith=self.q)
 
         return qs
+
+
+class GisLocationsInUseViewset(ListAPIView):
```
robertavram: can we move all of these api endpoints to the "management" app, and add the urls there as well...
the endpoints should be under: `/api/management/`

EquiTrack/locations/serializers.py
csabadenes: Not 100% sure, but i think yes, they have new fields (this exports the "geom" field in the format requested by the GIS team). I made new ones to avoid breaking old functionality..

EquiTrack/locations/views.py
csabadenes: Yeah, i will change it to superuser. I was not sure about the access level the GIS users have.

EquiTrack/locations/views.py
csabadenes: Sure, i will move them asap.

csabadenes:
@robertavram i think i made the requested changes

PR #1281, opened 2018-03-22
Attachment data denormalized

reinbach:
Attachment data denormalized on save, this allows attachments endpoint to not require caching and helps with performance.

[ch4173]


EquiTrack/attachments/utils.py
```diff
@@ -0,0 +1,102 @@
...
+            "vendor_number": vendor_number,
+            "pd_ssfa_number": pd_ssfa_number,
+            "file_type": file_type,
+            "file_link": attachment.file_link,
+            "uploaded_by": uploaded_by,
+            "created": attachment.created.strftime("%d %b %Y"),
+        }
+    )
+
+    if not created:
```
dpoirier: Would update_or_create be helpful here?  https://docs.djangoproject.com/en/stable/ref/models/querysets/#update-or-create

reinbach: Yes that would be much more helpful :)

PR #1289, opened 2018-03-23
Run tests on both python2 and python3

dpoirier:
[ch1939] https://app.clubhouse.io/unicefetools/story/1939/python3


EquiTrack/vision/client.py
```diff
@@ -106,7 +104,7 @@ def main():
         logger.info(client.get_business_areas())
 
     except Exception as exp:
-        logger.exception(six.text_type(exp))
+        logger.exception(str(exp))
```
domdinicola: six.string_types ?

EquiTrack/users/tests/factories.py
```diff
@@ -61,28 +62,19 @@ class Meta:
     )
 
 
+@factory.django.mute_signals(signals.pre_save, signals.post_save)
```
domdinicola: :100:

EquiTrack/t2f/tests/test_travel_details.py
```diff
@@ -2,10 +2,10 @@
 
 import json
 from datetime import datetime
-from StringIO import StringIO
 
 from django.core.urlresolvers import reverse
 from django.utils import six
+from six import StringIO
```
domdinicola: please import from django.utils.six

EquiTrack/t2f/tests/test_action_points.py
```diff
@@ -3,10 +3,11 @@
 import csv
 import json
 from datetime import datetime, timedelta
-from StringIO import StringIO
+from io import StringIO
```
domdinicola: shouldn't we import this from six?

EquiTrack/reports/migrations/0015_auto_20180129_1921.py
```diff
@@ -13,11 +13,12 @@
 
 def myprint(*args):
     print(*args)
-    file_name = 'migration_reports_0015.txt'
-    args_list = [six.text_type(arg) for arg in args]
-    with open(file_name, 'ab') as f:
-        f.write(', '.join(args_list))
-        f.write('\n')
+    # Do we really still need the following?
```
domdinicola: maybe would be useful to list have this function, this a more generic name (maybe coming as parameter)

EquiTrack/partners/tests/test_models.py
```diff
@@ -1594,9 +1594,11 @@ def test_intervention_attachment(self):
         self.assertEqual(six.text_type(instance), u'tv\xe5.txt')
 
     def test_intervention_reporting_period(self):
-        intervention = InterventionFactory.build(number=b'two')
+        intervention = InterventionFactory.build(number='two')
+        six.text_type(intervention)
         instance = InterventionReportingPeriodFactory.build(intervention=intervention)
-        self.assertTrue(six.text_type(instance).startswith(b'two'))
+        six.text_type(instance)
```
domdinicola: do we need this row?

EquiTrack/partners/tests/test_models.py
```diff
@@ -1594,9 +1594,11 @@ def test_intervention_attachment(self):
         self.assertEqual(six.text_type(instance), u'tv\xe5.txt')
 
     def test_intervention_reporting_period(self):
-        intervention = InterventionFactory.build(number=b'two')
+        intervention = InterventionFactory.build(number='two')
+        six.text_type(intervention)
```
domdinicola: do we need this row?

.circleci/config.yml
```diff
@@ -38,7 +43,7 @@ jobs:
             . env1/bin/activate
             cd EquiTrack
             mkdir media
-            bash runtests.sh;
+            tox
```
domdinicola: should we start to use tox and execute only python 2 for now?
my concern is related to the time to execute tests on both py2 and py3.
in case for now we start executing tox -e py2 we are fine!
Meanwhile locally we can execute tox

EquiTrack/t2f/tests/test_travel_details.py
dpoirier: It is imported from django.utils.six (django.utils.six is imported as 'six' in the preceding line) but I'll make it more explicit.

EquiTrack/partners/tests/test_models.py
dpoirier: At least it makes sure ```__str__``` doesn't raise an exception or anything.

EquiTrack/partners/tests/test_models.py
dpoirier: At least it makes sure ```__str__``` doesn't raise an exception or anything.

.circleci/config.yml
dpoirier: I'm skeptical that developers would consistently test against py3 and so we'd be continually playing catch-up trying to fix things that got broken on py3. My vote would be to do all the testing every time, and try to push ahead on squashing migrations and anything else that would speed up all the tests.

Second best would be to at least run flake8 under python3 to catch things that are blatantly incorrect under Python3.

EquiTrack/t2f/tests/test_action_points.py
dpoirier: will fix.

EquiTrack/vision/client.py
dpoirier: Looking at this again, using ```logger.exception``` will log the exception details already, so passing in the stringified exception is redundant. I'll change it to just say something like "Exception in vision client".

EquiTrack/reports/migrations/0015_auto_20180129_1921.py
dpoirier: I'm of the opinion that, unless I'm writing a library, it's better to wait until I need code before writing it.  

My assumption is that this code was here to help debug this migration, so it's no longer needed. I'm going to go ahead and remove it, but I can put it back if my assumption is wrong.

dpoirier:
@domdinicola Updated.  I'll leave the changes-requested label on for you to look at the changes.

robertavram:
I think we'll move this in this week... Also hoping to officially change to Python 3 REALLY soon

PR #1292, opened 2018-03-23
Feature/reset migrations command

domdinicola:


domdinicola:
to test this feature locally

1) remove obsolete apps modeles to delete
python manage.py clear-migrations-table


2) fake initial
python manage.py migrate_schemas --fake-initial  --noinput

domdinicola:
@reinbach can you have a look to this tomorrow?
Also do you have any clue of why test are failing? Seems to be related to the new Attachment model..

EquiTrack/locations/migrations/0001_initial.py
```diff
@@ -28,8 +27,8 @@ class Migration(migrations.Migration):
                 ('api_key', models.CharField(max_length=254)),
```
reinbach: We may want to add `verbose_name` to these fields. Not sure if it is really needed for this table, but now would be the time to add them, if we do.

This applies to all tables.

dpoirier:
Starting with a database I had restored from a production dump from January and migrated up to current develop level, I was able to ```python manage.py clear-migrations-table;python manage.py migrate_schemas --fake-initial --noinput``` successfully.  

I'm not set up to check whether things still work at that point, other than running the tests, which CircleCI does already.

robertavram:
@domdinicola based on what @dpoirier is saying... could we not run `python manage.py clear-migrations-table` then run `python manage.py migrate_schemas --fake --noinput` I'm not sure that fake_initial will be sufficient since there are some migrations that are 0002 because of dependencies.

thoughts? clear table and fake everything.

domdinicola:
My idea was not to migrate again the migrations that are not related to etools, but from different libraries.
The fact that we have 0002 I think is related that sometimes you need another migration from another app to be applied before

ntrncic:
tested it locally with rob's recommendation. Did a quick smoke test on etools, everything works well!

PR #1298, opened 2018-03-27
t2f 4 hact ch4598

domdinicola:
[ch4598]


EquiTrack/partners/models.py
```diff
@@ -614,15 +614,15 @@ def programmatic_visits(cls, partner, update_one=False):
                 travel_type=TravelType.PROGRAMME_MONITORING,
```
ntrncic: in `if update_one` we have to check the end date of the trip then find the quarter. right now we are just looking for current date quarter.

domdinicola: sure!

EquiTrack/partners/models.py
```diff
@@ -596,11 +596,13 @@ def planned_visits(cls, partner):
...
 
     @classmethod
-    def programmatic_visits(cls, partner, update_one=False):
+    def programmatic_visits(cls, partner, event_date=None, update_one=False):
         """
         :return: all completed programmatic visits
         """
-        quarter_name = get_quarter()
+        if not event_date:
+            event_date = datetime.datetime.today()
```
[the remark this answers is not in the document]
domdinicola: it used to be like that (get quarter by default uses today if no date is passed)
i moved the logic into "update_one" case.

are we going to keep this "update_one" mechanism or would we move to a daily task approach?

EquiTrack/partners/models.py
ntrncic: if there is not event date we shouldn't assume todays date. it should always have an event date if updating one.

domdinicola: OK! I added updated the check.
Anyway (as discussed) we will remove the logic to update_one in favor of the task which calculate the json once a day.

PR #1302, opened 2018-03-28
Action points models

roman-karpovich:
Models for upcoming Action Points module.


EquiTrack/action_points/migrations/0001_initial.py
```diff
@@ -0,0 +1,58 @@
...
+import django_fsm
+import model_utils.fields
+
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+        ('contenttypes', '0002_remove_content_type_name'),
```
domdinicola: please generate again this migration

EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,153 @@
...
+    due_date = models.DateField(verbose_name=_('Due Date'), blank=True, null=True)
+    high_priority = models.BooleanField(default=False, verbose_name=_('High Priority'))
+
+    action_taken = models.TextField(verbose_name=_('Action Taken'), blank=True)
+
+    section = models.ForeignKey(Section, verbose_name=_('Section'))
+    office = models.ForeignKey(Office, verbose_name=_('Office'))
+    location = models.ForeignKey(Location, verbose_name=_('Location'), blank=True, null=True)
+    partner = models.ForeignKey(PartnerOrganization, verbose_name=_('Partner'))
+    cp_output = models.ForeignKey(Result, verbose_name=_('CP Output'), blank=True, null=True)
```
domdinicola: - related object: generic fk should be for following models
* result
* partner
* intervention
* travel
* activity
* engagement

no need to have partner, cp_output and intervention here

EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+    def send_email(self, recipient, template_name):
+        context = {
+            'environment': get_environment(),
+            'action_point': self.get_mail_context(),
+            'recipient': recipient.get_full_name(),
+        }
+
+        notification = Notification.objects.create(
```
domdinicola: FYI, we are refactoring Notifications! We will keep you update on this!

EquiTrack/action_points/migrations/0002_auto_20180327_1215.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('post_office', '0004_auto_20160607_0901'),
+        ('action_points', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_email_template, migrations.RunPython.noop)
```
domdinicola: please add the email also in the update_notifications command

EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,153 @@
...
+from action_points.conditions import ActionPointCompleteRequiredFieldsCheck
+from locations.models import Location
+from notification.models import Notification
+from partners.models import PartnerOrganization, Intervention
+from reports.models import Result
+from users.models import Section, Office
+
+
+@python_2_unicode_compatible
+class ActionPoint(TimeStampedModel, models.Model):
```
domdinicola: TimeStampedModel already extends models.Model, please remove it

EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,153 @@
...
+                               verbose_name=_('Author'))
+    assigned_by = models.ForeignKey(settings.AUTH_USER_MODEL, related_name='+', verbose_name=_('Assigned By'))
+    assigned_to = models.ForeignKey(settings.AUTH_USER_MODEL, related_name='assigned_action_points',
+                                    verbose_name=_('Assigned To'))
+
+    status = FSMField(verbose_name=_('Status'), max_length=10, choices=STATUSES, default=STATUSES.open, protected=True)
+
+    description = models.TextField(verbose_name=_('Description'))
+    due_date = models.DateField(verbose_name=_('Due Date'), blank=True, null=True)
+    high_priority = models.BooleanField(default=False, verbose_name=_('High Priority'))
```
domdinicola: I would change this to a choice field, just in case the requirement change in order to have more levels

EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,153 @@
...
+    action_taken = models.TextField(verbose_name=_('Action Taken'), blank=True)
+
+    section = models.ForeignKey(Section, verbose_name=_('Section'))
+    office = models.ForeignKey(Office, verbose_name=_('Office'))
+    location = models.ForeignKey(Location, verbose_name=_('Location'), blank=True, null=True)
+    partner = models.ForeignKey(PartnerOrganization, verbose_name=_('Partner'))
+    cp_output = models.ForeignKey(Result, verbose_name=_('CP Output'), blank=True, null=True)
+    intervention = models.ForeignKey(Intervention, verbose_name=_('PD/SSFA'), blank=True, null=True)
+
+    date_of_complete = MonitorField(verbose_name=_('Date Action Point Completed'), null=True, blank=True,
```
domdinicola: :100:

EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,153 @@
...
+    comments = GenericRelation('django_comments.Comment', object_id_field='object_pk')
+
+    history = GenericRelation('snapshot.Activity', object_id_field='target_object_id',
+                              content_type_field='target_content_type')
+
+    class Meta:
+        ordering = ('related_module', 'related_content_type', 'related_object_id')
+
+    def save(self, **kwargs):
+        if self.related_content_type:
```
domdinicola: I would also add a check "if not self.related_module" in order to not override if the value is set.

EquiTrack/action_points/signals.py
```diff
@@ -0,0 +1,12 @@
...
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from action_points.models import ActionPoint
+
+
+@receiver(post_save, sender=ActionPoint)
+def action_point_updated_receiver(instance, created, **kwargs):
+    if created and instance.related_object is None:
+        instance.send_email(instance.assigned_to, 'action_points/action_point/created_unlinked')
```
domdinicola: I don't see (yet) a place to notify that the action point has been (re)assigned to another user. Am I missing it?

EquiTrack/action_points/models.py
roman-karpovich: Thanks for information!

EquiTrack/action_points/signals.py
roman-karpovich: You're correct, i'll add mailing also for reassigning.

EquiTrack/action_points/models.py
roman-karpovich: I would to keep related module and related object consistent. These fields shouldn't be editable after action point creating, and will be used separately in context of action points dashboard and as nested app.

EquiTrack/action_points/models.py
roman-karpovich: These fields are listed in specs, and i think it's ok to store widely used information in place instead of fetching through generic relation.

EquiTrack/snapshot/utils.py
```diff
@@ -85,12 +85,18 @@ def create_snapshot(target, target_before, by_user):
...
-    )
+    activity_kwargs = {
+        'target': target,
+        'by_user': by_user,
+        'action': action,
+        'data': current_obj_dict,
+        'change': change
+    }
+
+    get_additional_data = getattr(target, 'get_additional_data', None)
```
reinbach: Any reason not to add this additional data to the actual `data`, if could have a unique key that is explicit (eg: `_additional_data`)

EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,153 @@
...
+        STATUSES.open: 'created',
+        STATUSES.completed: 'date_of_complete'
+    }
+
+    KEY_EVENTS = Choices((
+        ('status_update', 'Status Update'),
+        ('reassign', 'Reassign'),
+    ))
+
+    related_module = models.CharField(max_length=20, choices=MODULE_CHOICES, blank=True, null=True)
```
reinbach: Why do we store this information? It seems extra, as it is available in `related_content_type`.
What am I missing? Would save you having to maintain these two fields in sync.

EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+
+    @property
+    def status_date(self):
+        return getattr(self, self.STATUSES_DATES[self.status])
+
+    def __str__(self):
+        return self.reference_number
+
+    def get_additional_data(self, diff):
```
reinbach: Maybe make the name of this method more explicit that it is used by `snapshot` module (eg: `snaphot_additional_data`). I have visions of models having `get_additional_data` for normal operations, and snapshot starts attempting to grab that info :)

EquiTrack/action_points/models.py
reinbach: Maybe rename the field to `date_completed` or `date_of_completion`

EquiTrack/action_points/models.py
roman-karpovich: As i can see from specs, we need to have related module as dropdown in action points dashboard. In this case we need to store it separately from related_object.

EquiTrack/action_points/migrations/0001_initial.py
roman-karpovich: done

roman-karpovich:
reopened this PR as it's contains unneeded migration of audit & tpm emails from emails fixture to update_notifications command.

PR #1308, opened 2018-03-30
Increase attachment api endpoint performance

reinbach:
[ch5015]

Locally the endpoint was taking  > 4s per request
With changes the endpoint now takes < 1s per request


ntrncic:

PR #1309, opened 2018-03-30
Migrate Auditor portal & TPM email templates into update_notifications command.

roman-karpovich:


roman-karpovich:
decided to migrate audit & tpm emails into command for easier integration with action points branch

domdinicola:
Thanks @roman-karpovich!
Actually we were discussing with @robertavram and he was proposing to have fixtures...
Sorry to tell you this just now :-1: 

Maybe this command can become a command that calls a loaddata command for each module (one fixture file for each module). This way would be more clear.

@robertavram ?

robertavram:
@domdinicola yes.. fixtures would be better... let's move after..

PR #1311, opened 2018-03-30
Moved validation to separate package/library

reinbach:
[ch5017]

Moved validation code (validation_mixins) to a separate library (see https://github.com/unicef/etools-validator)

Valiadtor repo still requires a lot more tests, and does need to be packaged. At the moment just referring to github repo in pip requirements file. If we're ok with this approach, I can go ahead and finish those steps up.


EquiTrack/requirements/test.txt
```diff
@@ -1,5 +1,6 @@
 -e git+https://github.com/robertavram/django-storages.git@1.3.1.2#egg=django-storages
 -e git+https://github.com/unicef/djangosaml2.git@v0.16.11.2#egg=djangosaml2
+-e git+https://github.com/unicef/etools-validator#egg=validator
```
robertavram: for requirements, modify `requirements/input/base.in` and run fab update_requirements to update...
... actually just run all the commands in the fab file... this is because it needs to be run in the container... the container has python3 and fabric won;t work.... if you feel inspired and want to add invoke here... i would not complain :)

reinbach: :+1: 
Converting from fabric to invoke, is being done in a separate pull request

PR #1316, opened 2018-04-02
Action points models

roman-karpovich:
copy of #1302 without dependency from #1309


EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,161 @@
...
+    status = FSMField(verbose_name=_('Status'), max_length=10, choices=STATUSES, default=STATUSES.open, protected=True)
+
+    description = models.TextField(verbose_name=_('Description'))
+    due_date = models.DateField(verbose_name=_('Due Date'), blank=True, null=True)
+    high_priority = models.CharField(choices=HIGH_PRIORITY_CHOICES, default=HIGH_PRIORITY_CHOICES.no, max_length=10,
+                                     verbose_name=_('High Priority'))
+
+    action_taken = models.TextField(verbose_name=_('Action Taken'), blank=True)
+
+    section = models.ForeignKey(Section, verbose_name=_('Section'))
```
robertavram: section should Foreign key to reports.Sector

domdinicola: btw we should remove section model @robertavram

EquiTrack/snapshot/models.py
```diff
@@ -41,6 +41,7 @@ class Activity(TimeStampedModel):
     )
     data = JSONField(verbose_name=_("Data"))
     change = JSONField(verbose_name=_("Change"), blank=True)
+    additional_data = JSONField(blank=True, null=True, verbose_name=_("additional data"))
```
domdinicola: @reinbach what do you think here?

EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,161 @@
...
+    office = models.ForeignKey(Office, verbose_name=_('Office'))
+    location = models.ForeignKey(Location, verbose_name=_('Location'), blank=True, null=True)
+    partner = models.ForeignKey(PartnerOrganization, verbose_name=_('Partner'))
+    cp_output = models.ForeignKey(Result, verbose_name=_('CP Output'), blank=True, null=True)
```
domdinicola: I still think that partner, cp_ouput and intervention should be part of the generic FK @ntrncic @robertavram

roman-karpovich: There is some misunderstanding about required fields for action points creating. Would be great if we can see updated list of editable fields for new action point.

about `partner`, `cp_ouput` and `intervention`:
For some models we can't certainly identify required objects. For example `Result` is connected to multiple cp_outputs(`outputs`), partners(`outputs__intervention_links__intervention__agreement__partner`) and interventions(`outputs__intervention_links__intervention`). . Same for `Travel`: we can't select single value, because all required relations are inside `TravelActivity`.

pitsevich: @roman-karpovich @domdinicola @ntrncic @robertavram 

I guess, it's best to pause for a second and clarify a few details on the spec to ensure that everyone is on the same page - we just had a talk with @roman-karpovich and realized that there's a few questions left on our end.

First, of all: 
Do we read the spec correctly, seeing that a new Action Point **created from the Action Points Dashboard** is not related to any object and any eTools module? 
As far as we understand, this Action Point would not have any links to Partner, CP_Output and Intervention?

(If that's the case - then we can remove "related_module" field, because the module can be fetched from related generic object.)


---------------

Second: Regarding "moving" Partner, CP_Output and Intervention to generic relation.

It corresponds with the [DB schema which was drawn during the workshop](https://www.dropbox.com/s/qwbbhrgqdlwg8t4/Screenshot%202018-04-03%2020.24.44.png?dl=0), but it looks like it contradicts the specs (or maybe I've forgot the original intent).

If we would move these fields to Generic FK (so, they'll be together with Engagement, TPM Activity, Trip Activity) - it would mean that these fields won't be set inside TPM/AP/T2F modules, and they also won't be set in A.P. Dashboard (see above).

So, it would be the same as if we would just remove these links from the code, as they are not actually used.

However, as we see in multiple spec sections, these fields are used:

- [example 1](https://www.dropbox.com/s/9w20xruwkfk814g/Screenshot%202018-04-03%2020.35.47.png?dl=0)
- [example 2](https://www.dropbox.com/s/c4rbhu8qncgerwl/Screenshot%202018-04-03%2020.36.02.png?dl=0)
- [example 3](https://www.dropbox.com/s/hhcr7qgdexmienu/Screenshot%202018-04-03%2020.36.51.png?dl=0)
- [example 4](https://www.dropbox.com/s/cxcmjiiv0owxwez/Screenshot%202018-04-03%2020.37.28.png?dl=0)


Hopefully, the questions are clear.
Please help us clarify the confusion :)

EquiTrack/action_points/models.py
```diff
@@ -0,0 +1,161 @@
...
+                               verbose_name=_('Author'))
+    assigned_by = models.ForeignKey(settings.AUTH_USER_MODEL, related_name='+', verbose_name=_('Assigned By'))
+    assigned_to = models.ForeignKey(settings.AUTH_USER_MODEL, related_name='assigned_action_points',
+                                    verbose_name=_('Assigned To'))
+
+    status = FSMField(verbose_name=_('Status'), max_length=10, choices=STATUSES, default=STATUSES.open, protected=True)
+
+    description = models.TextField(verbose_name=_('Description'))
+    due_date = models.DateField(verbose_name=_('Due Date'), blank=True, null=True)
+    high_priority = models.CharField(choices=HIGH_PRIORITY_CHOICES, default=HIGH_PRIORITY_CHOICES.no, max_length=10,
```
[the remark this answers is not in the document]
roman-karpovich: so maybe would be better to rename it to 'priority'  with 'high' /'low' choices?

EquiTrack/snapshot/models.py
roman-karpovich: good idea, i'll do so

EquiTrack/action_points/models.py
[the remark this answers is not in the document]
reinbach: that sounds good to me

EquiTrack/action_points/models.py
ntrncic: @pitsevich @roman-karpovich 
Q1: Yes we should be able to create an Action Point that is not related to any other module/model
Q2: We should keep the foreign keys to different models, we should also add foreign keys to t2f, tpm, ap. Then there is no need for generic foreign key. 

Hope that helps.

pitsevich: @ntrncic Thanks for the responses - we'll review tomorrow in details.

Quick follow-up on Q1:

Just to clarify:  Action Point Dashboard UI should allow to create **only** action points that are **not** linked to module/particular object?

roman-karpovich:
branch updated according to conversations, please review

PR #1326, opened 2018-04-04
update requirements part 1

domdinicola:


domdinicola:
I'll create another PR for DRF

robertavram:
highly recommend against blanked upgrades. I don't have the enough confidence in our tests yet... let's filter through requirements together

domdinicola:
The most of the new requirement are minor or fix releases and I think is really important to upgrade requirements from time to time..
I think we can merge this at the beginning of the sprint, so while we develop we do also test the new requirements.
what do you think @robertavram ?

robertavram:
please comment once you confirm all packages are good to go...

domdinicola:
ready to go!

PR #1330, opened 2018-04-05
Fix InterventionPartnershipDashView

reinbach:
Added last_pv_date field to Intervention and calculate on travel data changes.

Looked into making use of `extra()` but that started looking like a very complex query.

Once we upgrade to Django >1.11 we could make use of better subquery expressions that become available. See https://docs.djangoproject.com/en/2.0/ref/models/expressions/#subquery-expressions

For now I think this may be an interim solution.


robertavram:
the other PR works for now.. waiting for django 1.11 and fixing then...

PR #1332, opened 2018-04-05
Modify InterventionPartnershipDashView view to fix query

reinbach:
Attempting to minimize number of files changed.

This solution, performance wise, is a bit slower than https://github.com/unicef/etools/pull/1330


reinbach:
Replaced by https://github.com/unicef/etools/pull/1334

PR #1336, opened 2018-04-06
Feature/setup.py

domdinicola:


robertavram:
@domdinicola this looks like it needs a ticket in clubhouse and a  description... Make sure @saxix is aware of this work as well... we wouldn't want to duplicate.

domdinicola:
yeah, I was talking to @saxix and I just published it to show it to him (as you can see the branch is old)

PR #1337, opened 2018-04-06
first attempt to fix the 'build_and_test' circleci task

csabadenes:


csabadenes:
Probably this is a wrong approach, and this should be done in the circle config in a virtualenv.

PR #1340, opened 2018-04-09
Location imports utf8 fix

csabadenes:


EquiTrack/locations/tasks.py
```diff
@@ -52,7 +52,7 @@ def create_location(pcode, carto_table, parent, parent_instance,
         except IntegrityError as e:
             logger.info('Not Added: {}', e)
 
-        logger.info('{}: {} ({})'.format(
+        logger.info(u'{}: {} ({})'.format(
```
vkurup: Hmmm, I'd be surprised if this changed any behavior, since `unicode_literals` is imported on line 1.

csabadenes: Thx, yeah, i noticed it later. I think i should explicitly decode in utf-8, but i don't understand why there aren't similar issues elsewhere

csabadenes: My bad. This variable was explicitly UTF-8 encoded elsewhere in the code, i did not notice it at first.

EquiTrack/locations/tasks.py
```diff
@@ -129,8 +129,6 @@ def update_sites_from_cartodb(carto_table_pk):
                 sites_not_added += 1
                 continue
```
csabadenes: I think this line can safely be removed. It seems it was added before the future libraries were adopted.

csabadenes:
Connects to [ch5123]

EquiTrack/locations/tasks.py
```diff
@@ -129,8 +129,6 @@ def update_sites_from_cartodb(carto_table_pk):
...
 
-            site_name = site_name.encode('UTF-8')
-
```
csabadenes: I think this line can safely be removed. It seems it was added before the future libraries were adopted.

EquiTrack/locations/tasks.py
```diff
@@ -52,7 +52,7 @@ def create_location(pcode, carto_table, parent, parent_instance,
         except IntegrityError as e:
             logger.info('Not Added: {}', e)
 
-        logger.info(u'{}: {} ({})'.format(
+        logger.info('{}: {} ({})'.format(
             'Added',
             location.name,
             carto_table.location_type.name
```
vkurup: Looking at the clubhouse error, it looks like this variable holds a non-ascii value, so I think this will need to be changed to `carto_table.location_type.name.decode('utf-8')` (at least under Python 2... if we are fully on Python3 then just ignore me :)

EquiTrack/locations/tasks.py
vkurup: agreed.

EquiTrack/locations/tasks.py
csabadenes: At this point i think the backend is fully python 3.
As far as i can tell, the variable `location.name` was actually holding an encoded unicode byte str, and that was causing the error. It should be fixed by the line of code removed above(`location.name` was filled with the value of  `site_name`).
Thx for the feedback:)

vkurup: :+1:

PR #1341, opened 2018-04-10
added filters for export ch4794

domdinicola:
@domdinicola needs to write a description in which he adds 
[ch4794]
:)


robertavram:
[ch4794]

PR #1344, opened 2018-04-11
[ch5158] Fix production settings/cryptography with Python 3

dpoirier:
Cryptography was expecting the key data to be binary.


robertavram:
well, that was hard :)) thanks for that @dpoirier

PR #1207, opened 2018-02-21
upgrade to django 1.11 [ch396]

domdinicola:

PR #1345, opened 2018-04-11
toward django 1.11 [ch396]

domdinicola:
These are changing required by django 1.11 (and 2.0), compatible with current 1.10.
In order to keep the pr https://github.com/unicef/etools/pull/1207 light (few files) while we wait django-tenant-schema to be ready.


domdinicola:
[ch396]

PR #1348, opened 2018-04-11
Remove django allauth

vkurup:
[ch5185]

Apologies if the template changes are hard to review since I re-indented things at the same time. Let me know if you'd like me to pull that out into a separate commit. Basically, I removed the Partner Login link, added some text about it being 'under construction', and then styled it so that hover was disabled.


robertavram:
👍

PR #1350, opened 2018-04-11
[ch5155] Remove all imports from __future__

dpoirier:
Imports from future were used to bring Python 3 behavior back to Python 2, but we no longer need to support Python 2.


domdinicola:
this has already been implemented in main package refactoring

PR #1353, opened 2018-04-12
fix for py3 [ch5215]

domdinicola:
fix needed for py3


robertavram:
@domdinicola card reference please

PR #1360, opened 2018-04-16
Denormalize Tpm and Ap attachments

reinbach:
[ch5235]

TPM and AP attachments are now also denormalized.

Also added a management command `denormalize_attachments` that will go through and denormalize any attachments that have no already been done.

### DIFF
--- a/EquiTrack/attachments/utils.py
+++ b/EquiTrack/attachments/utils.py
@@ -28,10 +28,8 @@ def get_partner_obj(obj):
         return obj.content_object.agreement.partner
     elif isinstance(obj.content_object, (InterventionAmendment, InterventionAttachment)):
         return obj.content_object.intervention.agreement.partner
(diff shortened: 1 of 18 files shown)


EquiTrack/attachments/utils.py
```diff
@@ -26,8 +28,10 @@ def get_partner_obj(obj):
         return obj.content_object.agreement.partner
     elif isinstance(obj.content_object, (InterventionAmendment, InterventionAttachment)):
         return obj.content_object.intervention.agreement.partner
-    elif isinstance(obj.content_object, (Agreement, Assessment)):
+    elif isinstance(obj.content_object, (Agreement, Assessment, Engagement)):
         return obj.content_object.partner
+    elif isinstance(obj.content_object, TPMActivity):
+        return obj.content_object.tpm_visit.tpm_partner
```
ntrncic: we can get the partner directly from `TPMActivity` base model `Activity`. We could also get the `pd_ssfa_number`

PR #1365, opened 2018-04-17
Feature/fr changes

robertavram:


robertavram:
will also address [CH5145]

PR #1368, opened 2018-04-18
[ch5324] Make FundsCommitmentHeader.responsible_person nullable

dpoirier:
It looks like we're getting data from some REST API for a
FundingCommitmentHeader but the responsible_person field is null. We
recently changed a lot of CharFields to not allow NULL values.  So
change that field back to nullable.


ntrncic:
done by Dom already

PR #1394, opened 2018-04-20
Feature/remove update one functionality [ch5124]

domdinicola:
[ch5124]


EquiTrack/hact/tasks.py
```diff
@@ -1,38 +1,193 @@
...
+            pvq1 = pvq2 = pvq3 = pvq4 = 0
+        else:
+            pv = InterventionPlannedVisits.objects.filter(
+                intervention__agreement__partner=self.partner, year=year,
+                intervention__status__in=[Intervention.ACTIVE, Intervention.CLOSED, Intervention.ENDED]
+            )
+            pvq1 = pv.aggregate(Sum('programmatic_q1'))['programmatic_q1__sum'] or 0
+            pvq2 = pv.aggregate(Sum('programmatic_q2'))['programmatic_q2__sum'] or 0
+            pvq3 = pv.aggregate(Sum('programmatic_q3'))['programmatic_q3__sum'] or 0
+            pvq4 = pv.aggregate(Sum('programmatic_q4'))['programmatic_q4__sum'] or 0
```
dpoirier: I believe this approach will make four separate queries to the database. It might be more efficient to build one queryset with all the aggregations, then extract the values. Django shouldn't hit the database until we start extracting the values, and then it should do one query and get them all.

EquiTrack/hact/tasks.py
```diff
@@ -1,38 +1,193 @@
...
+
+        hact = json.loads(self.partner.hact_values) if isinstance(
+            self.partner.hact_values, six.text_type) else self.partner.hact_values
```
dpoirier: We shouldn't need to use six anymore. ```str``` is the only string/text type in Python 3.

EquiTrack/hact/tasks.py
```diff
@@ -1,38 +1,193 @@
...
 
 import json
-from datetime import datetime
+from datetime import date, datetime
 
 from django.db import connection
-from django.utils import six
+from django.db.models import F, Sum
+from django.db.transaction import atomic
+from django.utils import six, timezone
```
dpoirier: No need to import ```six``` anymore.

EquiTrack/hact/tasks.py
dpoirier: Also, ```hact_values``` is a ```JSONField``` and Django should be extracting the JSON for us. We should be able to just reference the field and get back a Python value.  If that doesn't work, maybe we're doing something wrong somewhere else, like encoding the data as JSON before assigning its value to the field, so it ends up double-JSON-encoded?

dpoirier: It looks like Partner.save() is doing exactly that - assigning the result of json.dumps to the hact_values field.  Maybe we should fix that in another ticket, and clean up all the messy accesses to hact_values in the code?

EquiTrack/hact/tasks.py
```diff
@@ -1,38 +1,193 @@
...
+            self.partner.hact_values, six.text_type) else self.partner.hact_values
+        hact['programmatic_visits']['planned']['q1'] = pvq1
+        hact['programmatic_visits']['planned']['q2'] = pvq2
+        hact['programmatic_visits']['planned']['q3'] = pvq3
+        hact['programmatic_visits']['planned']['q4'] = pvq4
+        hact['programmatic_visits']['planned']['total'] = pvq1 + pvq2 + pvq3 + pvq4
+        self.partner.hact_values = hact
+        self.partner.save()
+
+    def programmatic_visits(self):
```
dpoirier: I see some room for optimizing queries in this function, but it'd make things a lot less readable. Do we know whether performance of this function is going to be significant overall?

EquiTrack/hact/tasks.py
```diff
@@ -1,38 +1,193 @@
...
+        self.partner.hact_values['spot_checks']['completed']['q3'] = trq3 + asc3
+        self.partner.hact_values['spot_checks']['completed']['q4'] = trq4 + asc4
+
+        sc = trip.count() + audit_spot_check.count()  # TODO 1.1.9c add spot checks from field monitoring
+        self.partner.hact_values['spot_checks']['completed']['total'] = sc
+        self.partner.save()
+
+    def audits_completed(self):
+        """
+        :return: all completed audit (including special audit)
```
dpoirier: Comment is inaccurate: method always returns None.

EquiTrack/hact/tasks.py
```diff
@@ -1,38 +1,193 @@
...
+        self.partner.hact_values['programmatic_visits']['completed']['q1'] = pvq1 + tpmv1
+        self.partner.hact_values['programmatic_visits']['completed']['q2'] = pvq2 + tpmv2
+        self.partner.hact_values['programmatic_visits']['completed']['q3'] = pvq3 + tpmv3
+        self.partner.hact_values['programmatic_visits']['completed']['q4'] = pvq4 + tpmv4
+        self.partner.hact_values['programmatic_visits']['completed']['total'] = pv + tpm_total
+
+        self.partner.save()
+
+    def spot_checks(self):
+        """:return: all completed spot checks"""
```
dpoirier: Comment is inaccurate: method always returns None.

EquiTrack/hact/tasks.py
```diff
@@ -1,38 +1,193 @@
...
+        hact['programmatic_visits']['planned']['q2'] = pvq2
+        hact['programmatic_visits']['planned']['q3'] = pvq3
+        hact['programmatic_visits']['planned']['q4'] = pvq4
+        hact['programmatic_visits']['planned']['total'] = pvq1 + pvq2 + pvq3 + pvq4
+        self.partner.hact_values = hact
+        self.partner.save()
+
+    def programmatic_visits(self):
+
+        """:return: all completed programmatic visits"""
```
dpoirier: Comment is inaccurate: method always returns None.

EquiTrack/hact/tasks.py
```diff
@@ -1,38 +1,193 @@
...
 
 
+class PartnerHactSynchronizer(object):
+
+    def __init__(self, partner) -> None:
+        self.partner = partner
+        super().__init__()
+
+    def planned_visits(self):
+        """:return: For current year sum all programmatic values of planned visits
```
dpoirier: Comment is inaccurate: method always returns None.

EquiTrack/hact/tests/test_tasks.py
```diff
@@ -20,3 +35,289 @@ def test_task_update(self):
...
+            intervention=intervention,
+            year=year,
+            programmatic_q1=3
+        )
+        InterventionPlannedVisitsFactory(
+            intervention=intervention,
+            year=year - 1,
+            programmatic_q3=2
+        )
+        self.assertEqual(self.partner_organization.hact_values['programmatic_visits']['planned']['total'], 0)
```
dpoirier: How does this work? We don't appear to call update_hact_values() or planned_visits() anywhere.  If it's still passing, it seems like it's not really testing that the new code is doing anything.

EquiTrack/hact/tasks.py
```diff
@@ -1,38 +1,193 @@
...
+        audits = Audit.objects.filter(partner=self.partner, status=Engagement.FINAL,
+                                      date_of_draft_report_to_unicef__year=datetime.now().year)
+        hact['outstanding_findings'] = sum([
+            audit.pending_unsupported_amount for audit in audits if audit.pending_unsupported_amount])
+        self.partner.hact_values = json.dumps(hact, cls=HactEncoder)
+
+
 @app.task
-def update_hact_values():
+def update_hact_values(country_name=None):
```
dpoirier: I don't see anywhere update_hact_values is called with this new country_name argument; why is it being added?

EquiTrack/hact/tests/test_tasks.py
domdinicola: you're right, i'm calling planned_visits

EquiTrack/hact/tasks.py
domdinicola: I want to be able to call it using a country_name to filter it. If I want to use it as function in the shell

EquiTrack/hact/tasks.py
domdinicola: true.. updating

EquiTrack/hact/tasks.py
domdinicola: true.. updating

EquiTrack/hact/tasks.py
domdinicola: true.. updating

EquiTrack/hact/tasks.py
domdinicola: true.. updating

EquiTrack/hact/tasks.py
domdinicola: ;)

EquiTrack/hact/tasks.py
domdinicola: agree with fix this weird behavior..

EquiTrack/hact/tasks.py
domdinicola: this is actually a daily task... so I don't know what to suggest :sweat_smile:

dpoirier: If it's a background task, so nobody's waiting for it, it's probably fine.  Assuming it doesn't take all day :-)

EquiTrack/hact/tests/test_tasks.py
dpoirier: If the test passed without calling planned_visits(), then it's not really a valid test of planned_visits().  It looks like maybe we should create some data that would result in a non-zero total?

EquiTrack/audit/models.py
```diff
@@ -414,12 +414,6 @@ def save(self, *args, **kwargs):
     def submit(self, *args, **kwargs):
         return super(SpotCheck, self).submit(*args, **kwargs)
 
-    @transition('status', source=Engagement.STATUSES.report_submitted, target=Engagement.STATUSES.final,
```
domdinicola: @roman-karpovich can you confirm I can remove those and use the father's transitions?

roman-karpovich: i guess, they can be removed without any risk

PR #1397, opened 2018-04-23
fix add partner

domdinicola:


domdinicola:
we need this in order to allow the redirect performed by the frontend

PR #1408, opened 2018-04-24
move source code under 'etools' package (preview)

saxix:
# **move source code under 'etools' package**

- move all django applications under 'etools.applications' namespace
- move all libraries/modules under 'etools.libraries' namespace
- move project configuration/ settings under 'etools.config' namespace
- move 'etools' package under src directory
- move all scripts to root folder (manage.py,manage_debug.sh)
### REQUIREMENTS:
- move requirements in src/requirements directory
- add isort/pdbpp/ipython to test.in requirements
### TESTS/CIRCLECI
- updates circleci to use tox and produce html coverage reports
- updates tox.ini and django 1.11/2.0 matrix
- updates .coveragerc
### CODE
- order all imports following iSort rules
- removes all 'from __future__' imports
- fixes settings SITE_NAME and ROOT_URLCONF
- set MEDIA_ROOT and STATIC_ROOT to tmp
- update manage.py to be executed in the root folder
- creates setup.py
- creates MANIFEST.in
### DOCKER
- updates docker related files
- add default value to 'ARG REQUIREMENTS_FILE' in Dockerfile-dev
### OTHERS
- removes runtests.sh
- silence 'tenant_schemas.W003' warning
- updates .gitignore
- change README.md to README.rst
- move isort config to .isort.cfg from setup.cfg
- add comments to flake8


dpoirier:
Why put all the interesting parts down under src/etools/?   Could we put applications, assets, config etc at the top-level?  That will make them faster to get to.

saxix:
it is everything about etools project, so it is best practice only "consume" a clear unambiguous namespace. If we put everything at top level, we raise the risk of name clashes and we reduce clarity. we will have libraries, applications and config at top pythonpath level.

Also note that this should be the first step before we "pull out" applications into dedicated repositories. Read those packages as 'django.contrib'. They could use 'contrib' without `django` prefix, but the risk of name clashes would have raised too much and readability will lower by consequences

dpoirier:
There's no other project here; the whole repo is etools. So it seems to me there's no need to allow for distinguishing etools project's files from other projects.

In the same vein, we own this site project and its namespace. We're not django that has to coexist with every other project out there that uses it.

Even if we want to namespace our own apps under 'etools' just in case someday we might happen to want to use a third-party package that is named 'monitoring' or 'records', we don't need to introduce new 'src' and 'applications' directories too. 

Just to be concrete, I'm thinking something like this would be adequate.  It might not be perfectly pure in its organization, but it seems practical:

README
Procfile
...
.circleci/
assets/
config/
docs/
etools/
etools/action_points/
etools/other apps...
libraries/...
manage.py
manage.sh
requirements/
...

domdinicola:
I like the new organization it allows me to understand better:

1. where and which are the apps and the libraries and distinguish the first from the second
2. where is the project configuration
Btw I think that maybe we can get rid of assets.

about the "unambiguous namespace" I've been there... that's why the library is called azure_graph_api instead of azure (I was getting crazy :sweat_smile: )

:+1: for removing runtests.sh, create setup.py and MANIFEST

@saxix one question... why are requirements there and not in etools?

saxix:
Please ignore the conflicts, we can use this PR for the discussion, and I'll rebase the code as we all agree on what we hav

reinbach:
A minor suggestion, could we have a single dir to group anything that is built/generated eg: `htmlcov`, `dist` etc., labelled something like `build/`

dpoirier:
+1 for ```build/```.

PR #1409, opened 2018-04-25
t2f added filter for location [ch5381]

domdinicola:
[ch5381]


robertavram:
I completely forgot that all filtering is done on backend in t2f... so effort is minimal :))
awesome!

domdinicola:
the backend is ready, but a front end filter is needed indeed

PR #1414, opened 2018-04-26
User model step2

reinbach:
[ch5466]

Second step in moving to custom user model. Need step 1 to have made it to production, before merging this step in.

Email field now has a unique requirement.
Duplicate emails were marked with `dupliate-<email>`


domdinicola:
Do you think would make send to:
- call the model UnicefUser to distinguish it better?
- add the guid and the index number at user level?

src/etools/applications/users/admin.py
```diff
@@ -312,8 +319,7 @@ class SectionAdmin(admin.ModelAdmin):
 
 
 # Re-register UserAdmin
-admin.site.unregister(get_user_model())
-admin.site.register(get_user_model(), UserAdminPlus)
+admin.site.register(User, UserAdminPlus)
```
vkurup: Curious why we switched from `get_user_model()` to `User` here?

src/etools/applications/EquiTrack/tests/test_utils.py
```diff
@@ -72,7 +72,7 @@ def test_simple_instance(self):
                 'groups': [self.group.id],
                 'user_permissions': [self.permission.id],
                 'pk': user.id,
-                'model': 'auth.user',
+                'model': 'users.user',
```
vkurup: Grepping the codebase finds a couple other instances of `auth.user` but I'm not sure if they need to be switched (i.e., not sure if we're using them):

```
./src/etools/applications/EquiTrack/data/audit_users.json:3:        "model": "auth.user",
./src/etools/applications/EquiTrack/data/users.json:4:        "model": "auth.user",
```

src/etools/applications/users/migrations/0002_auto_20180424_1740.py
```diff
@@ -0,0 +1,46 @@
...
+
+from django.db import migrations, models
+from django.db.models import Count
+
+
+def mark_duplicates(apps, schema_editor):
+    User = apps.get_model("users", "user")
+    dupe_emails = set(User.objects.values_list("email", flat=True).annotate(
+        Count("id")
+    ).order_by().filter(id__count__gt=1))
```
vkurup: What does `.order_by()` without a parameter do?

src/etools/applications/utils/permissions/tests/test_permission_models.py
```diff
@@ -1,29 +1,27 @@
-
```
vkurup: Never noticed the spelling error in this filename. Thanks for fixing it!

src/etools/applications/EquiTrack/tests/test_utils.py
reinbach: Good find, I missed thoughts. And it does look like they are used.

src/etools/applications/users/migrations/0002_auto_20180424_1740.py
reinbach: The default ordering is on `first_name` which is not part of this result set, and with aggregation/annotation that causes issues, so we use a blank `order_by` to clear it. It is better explained at https://docs.djangoproject.com/en/2.0/topics/db/aggregation/#interaction-with-default-ordering-or-order-by

src/etools/applications/users/admin.py
reinbach: A cummulation of issues/errors, resulting in me coming to the wrong conclusion :)
Forgot to come back and double check this :(

src/etools/applications/users/migrations/0002_auto_20180424_1740.py
vkurup: I learn something new every day. Thanks for the pointer!

vkurup:
I got a migration error when trying to run migrations on a fresh DB:

```
$ dropdb etools
$ createdb etools
$ git checkout develop
$ python manage.py migrate_schemas
...
$ git checkout user-model-step2
$ python manage.py migrate_schemas
...
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration tpmpartners.0004_auto_20180503_1311 is applied before its dependency users.0002_auto_20180424_1740 on database 'default'.
```

PR #1417, opened 2018-04-26
update requirements part 2

domdinicola:
I do still want to check it (especially carto, and storage)


domdinicola:
@roman-karpovich @pitsevich we are upgrading these requirements FYI
drf-nested-routers, django-post-office

PR #1420, opened 2018-04-27
Update copy attachment management command

reinbach:
[5538]

Attachments modified in the last 25 hrs are updated

### DIFF
--- a/EquiTrack/partners/utils.py
+++ b/EquiTrack/partners/utils.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import datetime
 import logging
 
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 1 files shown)


EquiTrack/partners/utils.py
```diff
@@ -33,6 +34,11 @@ def update_or_create_attachment(file_type, content_type, object_id, filename):
     )
 
 
+def get_from_date():
+    """Return from date to use"""
+    return datetime.date.today() - datetime.timedelta(hours=25)
```
robertavram: can we make this configurable from an env var or maybe the task running can have a default of 25 if no params are passed in. The first time we run it we want to get all docs... if it fails to run for 5 days.. we want to be able to tell it to go back 5 days as a 1-off.

PR #1424, opened 2018-04-27
move source code under 'etools' package (updated to permissions2)

saxix:
- move all django applications under 'etools.applications' namespace
- move all libraries/modules under 'etools.libraries' namespace
- move project configuration/ settings under 'etools.config' namespace
- move 'etools' package under src directory
- move all scripts to root folder (manage.py,manage_debug.sh)
REQUIREMENTS:
- move requirements in src/requirements directory
- add isort/pdbpp/ipython to test.in requirements
TESTS/CIRCLECI
- updates circleci to use tox and produce html coverage reports
- updates tox.ini and django 1.11/2.0 matrix
- updates .coveragerc
CODE
- order all imports following iSort rules
- removes all 'from __future__' imports
- fixes settings SITE_NAME and ROOT_URLCONF
- set MEDIA_ROOT and STATIC_ROOT to tmp
- update manage.py to be executed in the root folder
- creates setup.py
- creates MANIFEST.in
DOCKER
- updates docker related files
- add default value to 'ARG REQUIREMENTS_FILE' in Dockerfile-dev
OTHERS
- removes runtests.sh
- silence 'tenant_schemas.W003' warning
- updates .gitignore
- change README.md to README.rst
- move isort config to .isort.cfg from setup.cfg
- add comments to flake8


dpoirier:
Would it make sense to put requirements under src/config/requirements?

dpoirier:
A few paths in .gitignore should be updated.

dpoirier:
Re: STATIC_ROOT and MEDIA_ROOT - could we put them under /tmp/etools (or something like that) by default?  Just using /tmp/media and /tmp/static assumes the dev system is used only for developing etools, which isn't the case for everyone.

saxix:
@dpoirier 
- requirements  do not need to be packaged and everything under etool is application code that need to be accessed at runtime
- agree on MEDIA/STATIC next run will have directories under /tmp/etools
- great catch on gitignore. will be fixed next run
- flake8 is now easy to run with flake8 src not sure we need a script for one line (in any case will be embedded in tests in step2, and a new dashboard with results/coverage will be available)
- runtests.sh can be restored in step1 but in any case not needed in step2 (only pytest src will be needed to run tests)

vkurup:
I haven't had a chance to look through all of this, but I'm also strongly +1 on a single command for developers to run all linting and tests. This includes flake8, unit tests and a test for missing migrations. I am the one that added runtests.sh to this repo. Before it was present, flake8 was not being run and missing migrations were a common deployment issue, IIRC.

PR #1445, opened 2018-05-03
Improve exception handling

vkurup:
Here's my attempt to improve this issue -> [ch1032]

There are still some try/except blocks which are bigger than we'd like, but for the most part, things are better than when that issue was created. 

Some unrelated changes:
- remove the try/except around xml.ElementTree since in [python 3, cElementTree is deprecated](https://docs.python.org/3.3/whatsnew/3.3.html#xml-etree-elementtree)
- improved our use of logging in some cases

### DIFF
--- a/EquiTrack/publics/tasks.py
+++ b/EquiTrack/publics/tasks.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import csv
+import xml.etree.ElementTree as ET
 from collections import defaultdict
 from datetime import date, datetime
 from decimal import Decimal, InvalidOperation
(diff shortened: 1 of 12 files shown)


EquiTrack/publics/tasks.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import csv
+import xml.etree.ElementTree as ET
```
domdinicola: :+1:

ntrncic:
@vkurup is this ready?

vkurup:
@ntrncic oops, yes, forgot to label it! Ready now.

PR #1446, opened 2018-05-04
Auditor portal updates

roman-karpovich:


ntrncic:
@roman-karpovich please make another PR against Master. Each PR that you do for the next week please make one against develop and one against master.

roman-karpovich:
@ntrncic thanks, understood! I'd opened [PR to master](https://github.com/unicef/etools/pull/1447). Please notify if something else would be required from my side.

EquiTrack/audit/serializers/export.py
```diff
@@ -81,6 +82,17 @@ class Meta:
         ]
 
 
+class AttachmentPDFSerializer(serializers.ModelSerializer):
```
reinbach: Any reason not to move this serializer to `attachments/serializers.py` in case another app wants to reuse this attachment serializer?

EquiTrack/utils/common/serializers/fields.py
```diff
@@ -112,3 +114,40 @@ def update(self, instance, validated_data):
...
+                    exc=exc
+                )
+            )
+            raise type(exc)(msg)
+
+    def to_representation(self, value):
+        value = set(value)
+
+        if self.export_attr:
+            value = map(lambda x: get_attribute_smart(x, self.export_attr), value)
```
reinbach: In stead of having to write lambda we could have the following;

`value = [get_attribute_smart(x, self.export_attr) for x in value]`

I find that a little more readable, but maybe it is just me

EquiTrack/utils/common/serializers/fields.py
```diff
@@ -112,3 +114,40 @@ def update(self, instance, validated_data):
...
+
+        return ', '.join(map(str, filter(lambda x: x, value)))
```
reinbach: Maybe instead do;

`return ', '.join([str(x) for x in value])`

EquiTrack/utils/common/serializers/fields.py
roman-karpovich: ok, i'll do so

EquiTrack/audit/tests/test_views.py
```diff
@@ -676,6 +676,34 @@ def test_list_view(self):
...
+            data={'email': user.email},
+            user=self.unicef_focal_point,
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.data), 1)
+        self.assertEqual(response.data[0]['email'], user.email)
+
+    def test_staff_search(self):
+        UserFactory(auditor=True, partner_firm=self.auditor_firm)
+        user = UserFactory(auditor=True, partner_firm=self.auditor_firm, email='test@unique_example.com')
```
reinbach: I'd change the email address in tests to always be `@example.com` to prevent any leaking of emails.
Change the leading part for uniqueness `eg: test_unique@example.com`

PR #1448, opened 2018-05-07
improved pmp report [ch5638]

domdinicola:
[ch5638]


robertavram:
update from develop please

PR #1457, opened 2018-05-09
Try to untangle dates and datetimes somewhat

dpoirier:
The code had some places where dates and datetimes were
confused.

Some things I tried to keep in mind when making these changes:

* Don't make any changes that would be visible in the APIs, since
  that would probably break clients
* Don't change any actual data types in the database, since again,
  that might break clients

So here's what I ended up doing:

* Renamed some variables whose names contradicted their types
* Renamed model fields whose names contradicted their types
* Added some field declarations for DRF, to keep fields associated
  with renamed model fields from changing names in the API
* Removed unused Quarter model (which I noticed since it seemed
  odd that it used datetimes and not just 1-4 to identify
  quarters, I started changing that, then couldn't find any code
  other than its own test that needed to be updated as a result).


dpoirier:
I couldn't find anything that was using the Quarter model, so I cleaned it away.

dpoirier:
@ntrncic that just leaves changing a few variable names. Is that worth the overhead of a PR?

dpoirier:
Closing this, as after making the requested changes all that would be left would be a few variable name changes and that doesn't seem worth a pull request at this point.

PR #1460, opened 2018-05-09
[ch5698] Add HACT update button to country page

dpoirier:
On the change page in the admin for the users Country model, add a
button that updates the HACT values for the current country, or if the
current country is the global/public country, schedules tasks to
update all the countries.

### DIFF
--- a/src/etools/applications/users/admin.py
+++ b/src/etools/applications/users/admin.py
@@ -1,13 +1,15 @@
 from functools import update_wrapper
 
 from django.conf.urls import url
-from django.contrib import admin
+from django.contrib import admin, messages
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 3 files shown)


src/etools/applications/users/admin.py
```diff
@@ -289,6 +292,16 @@ def execute_sync(country_pk, synchronizer):
             sync_handler.delay(country.name, synchronizer)
         return HttpResponseRedirect(reverse('admin:users_country_change', args=[country.pk]))
 
+    def update_hact(self, request, pk):
+        country = Country.objects.get(pk=pk)
+        if country.schema_name == get_public_schema_name():
```
domdinicola: can you use get_public_schema_name also in execute_sync function

PR #1461, opened 2018-05-09
Update reporting requirements

reinbach:
[5677]

    ignore ids for reporting requirements
    -- query on dates

    amendment logic is wacky
    -- should be flag on intervention

    special can be individual update/delete
    -- move out if want to
    -- not able to delete if dates passed (able to edit)


src/etools/applications/partners/serializers/interventions_v2.py
```diff
@@ -660,7 +662,7 @@ def validate(self, data):
 
         # Only able to change reporting requirements when PD
         # is in amendment status
-        if self.intervention.status not in [self.intervention.DRAFT]:
+        if not self.intervention.in_amendment:
```
robertavram: you are also able to change in draft.... either in draft or if in_amendment.. (both :) )

PR #1463, opened 2018-05-10
[ch4800] Change some messages

dpoirier:
* The message that reads "Field partner_type: Vendor record must belong to PRG2 account group (start from 2500 series)" should read "Vendor number must belong to PRG2 account group"

* The message that reads "Partner Organization already exists with this vendor number" should read "This vendor number already exists in eTools"

NOTE: ntrncic said that this requested change did not look valid: The message that reads " Loading data from Vision Failed, no valid response returned for data:...." should read "The vendor number entered is incorrect"


domdinicola:
not sure about the second message, but the first message was provided by the business

domdinicola:
@ntrncic please revert this
messages were asked by the business

PR #1465, opened 2018-05-10
Hotfix for carto query pagination

csabadenes:


csabadenes:
These changes can result in issues if the Carto dataset ID's are tampered with. For now this is the only solution which works, pagination by offset and limit still hits Carto API limits when the offset gets too high - it seems PgSql evaluates the rows before reaches the offset, and the query times can increase over the API limits.

PR #1467, opened 2018-05-10
adopt pytest as test runner

saxix:


src/etools/applications/partners/admin.py
```diff
@@ -9,11 +9,11 @@
 
 from etools.applications.EquiTrack.admin import ActivityInline, SnapshotModelAdmin
 from etools.applications.partners.exports import PartnerExport
-from etools.applications.partners.forms import (PartnersAdminForm,  # TODO intervention sector locations cleanup
-                                                PartnerStaffMemberForm, SectorLocationForm,)
+from etools.applications.partners.forms import PartnersAdminForm  # TODO intervention sector locations cleanup
```
reinbach: How come you're separating the import with a comment on it from the others?

src/etools/libraries/utils/__init__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from inspect import isclass
+
+
+def fqn(o):
+    """Returns the fully qualified class name of an object or a class
```
dpoirier: I like this. I always worry about typos when we embed code references into strings.

src/etools/applications/partners/admin.py
saxix: it is a script, need some cleanup later

src/etools/libraries/utils/__init__.py
saxix: :)

PR #1468, opened 2018-05-10
update partner rating, added filter by rating [ch1884]  [ch1879]

domdinicola:


domdinicola:
1. store "adapted/adjusted" value in the model since we do filter in the front-end
2. added migration to convert existing values

PR #1496, opened 2018-05-17
docs: remove py3 preparation

domdinicola:


vkurup:
Great minds think alike! I created [ch5814] when I noticed this.

PR #1501, opened 2018-05-18
pytest

saxix:
- use pytest as test runner
- run isort/flake8 during test phase
- static code analysis report available on https://app.codacy.com/app/UNICEF/etools/dashboard
- test and coverage reports available in HTML format in CircleCI
- implement CircleCI matrix django1.10/1.11/2.0

CODE
- updates some tests to fixes some incongruences

@vkurup fixes circleci config
@all some cleanup on imports are needed. plan to fixe them in new PR or just before merging


reinbach:
How come you close the PR and create a new one? We lose our comments, etc, 
Maybe reference the previous PR or do a force git push (not actually sure how that affects the PR though)?

vkurup:
Yes, I was going to ask the same question :) I'd prefer changes to be made on top of the previous branch an extra commit. In addition to preserving comments, it makes it easier to see what actually has changed from one commit to the next. When the entire PR is new, I now have to re-review every change to make sure that things I agreed with before haven't changed in this PR. If there is a concern about too many little commits, a rebase can be done just before merge.

saxix:
these refactoring are massive and executed by a script, in some situations is not possible to force the push and it's needed to recreate the branch. The same happened in step1.

saxix:
@vkurup  there is no need to change the way you write tests, you can continue to use the current way (I did not changed the tests). verbosity can be reduced (removes -v in pytest.ini)
About imports, they are automatically sorted using `isort` we can configure the way it works editing ```.isort.cfg``` I have used a basic configuration just to lint the code.

vkurup:
I guess I was thinking about how I 'run' tests, not write them. Currently I know how to run just 1 particular test or set of tests, which I can run over and over again while I am working on a module, while keeping an eye on the coverage report to make sure that I am covering all new code. I'll need to learn how to do that for pytest, which I'm sure is not that hard.

saxix:
@vkurup  ```.pytest_cache``` added to ```.gitignore```, ```runtests.sh``` now execute pytest

PR #1503, opened 2018-05-18
Add grants list to the cached pmp data on the frontend

csabadenes:
[ch5628]


domdinicola:
do we have a test for this?

csabadenes:
Yes, there is a test in partners/tests/test_api_pmp.py TestPMPDropdownsListApiView - i have modified it a bit. Should i add anything extra?

PR #1506, opened 2018-05-18
Remove djrill (again!)

vkurup:
Part of [ch5697]

Looks like this one was also inadvertently brought back after we removed it in #1003


domdinicola:
thank you!!

domdinicola:
maybe it was me when I introduced piptools.. sorry!

domdinicola:
nice to see that the list of requirement is being cleaned!

PR #1509, opened 2018-05-18
Remove django-hstore and drf-hstore

vkurup:
Part of [ch5697]

AFAICT, we are not using hstore at all anymore, so I could probably remove all of the documentation and deployment references to it.


vkurup:
I've dropped the hstore extension from my local DB and run tests successfully.

robertavram:
👍x 100

PR #1512, opened 2018-05-18
Remove django-autocomplete-light

vkurup:
part of [ch5697]

It was being used in the admin, but the URL was not set correctly, so it wasn't currently working anyway. There was also one API endpoint exposed, but grepping through the frontend apps, I can't see that it was used anywhere.


robertavram:
👍 Love it!

PR #1516, opened 2018-05-22
add & upgrade django gis dependency to DRF 3.8

csabadenes:


src/requirements/local.txt
```diff
@@ -75,7 +76,6 @@ html5lib==1.0b10
 httplib2==0.10.3
 idna==2.6
 imagesize==0.7.1          # via sphinx
-invoke==0.23.0
```
csabadenes: @domdinicola i don't understand why `update-requirements` downgraded `invoke`

domdinicola: weird there were 2 invoke in the file..

PR #1517, opened 2018-05-22
Change special reporting requirements URL to include intervention ID

csabadenes:
Also add unique rule for the same intervention&due_date combo


csabadenes:
Is the `('intervention', 'due_date')` DB unique key overkill?

[ch5784]

PR #1590, opened 2018-05-24
Insignificant change

vkurup:
Just trying to get pyup to see this commit.


vkurup:
Doesn't seem to be working. I've sent an email to the support address.

vkurup:
Not only is the CI part *not* working, but pyup is also sending PRs for updates against our wishes. See #1595, #1596, and #1597. I've turned off the pyup service for now.

PR #1598, opened 2018-05-29
Update PRP Partner api

robertavram:
https://github.com/unicef/etools-partner-reporting-portal/issues/699


robertavram:
fixes: https://github.com/unicef/etools-partner-reporting-portal/issues/612

PR #1599, opened 2018-05-29
Add non-special reporting requirements delete endpoint

csabadenes:
[ch5916]


src/etools/applications/partners/views/interventions_v2.py
```diff
@@ -800,3 +800,24 @@ def post(self, request, intervention_pk, report_type, format=None):
                 self.serializer_list_class(self.get_data()).data
             )
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    def delete(self, request, intervention_pk, report_type, format=None):
+        self.intervention = self.get_object(intervention_pk)
+        self.report_type = report_type
+
+        try:
```
reinbach: I'd suggest pushing/moving this logic to the serializer, and follow the normal flow of how data is validated and then saved/updated/deleted.

I believe there is business logic saying that past reporting requirements cannot be deleted, although they are allowed to be edited. Maybe this logic is limited to Special Reporting Requirements?

src/etools/applications/partners/views/interventions_v2.py
```diff
@@ -800,3 +800,24 @@ def post(self, request, intervention_pk, report_type, format=None):
                 self.serializer_list_class(self.get_data()).data
             )
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    def delete(self, request, intervention_pk, report_type, format=None):
+        self.intervention = self.get_object(intervention_pk)
```
reinbach: These 2 variables are set, but does not appear that they are used further in the code

src/etools/applications/partners/views/interventions_v2.py
```diff
@@ -800,3 +800,24 @@ def post(self, request, intervention_pk, report_type, format=None):
...
+        self.report_type = report_type
+
+        try:
+            deleted_ids = [int(rr["id"]) for rr in request.data["reporting_requirements"] if "id" in rr]
+
+            if len(deleted_ids) == 0:
```
reinbach: You could just use `if deleted_ids:` for this statement

src/etools/applications/partners/views/interventions_v2.py
```diff
@@ -800,3 +800,24 @@ def post(self, request, intervention_pk, report_type, format=None):
...
+
+            ReportingRequirement.objects.filter(
+                intervention=intervention_pk,
+                report_type=report_type,
+                id__in=deleted_ids
+            ).delete()
+            return Response(
+                self.serializer_list_class(self.get_data()).data
+            )
+        except ReportingRequirement.DoesNotExist:
```
reinbach: I don't see how this exception would be reached, as I see there is `filter` on the `ReportingRequirement`? Maybe I am missing something?

csabadenes: @reinbach i believe i addressed your requests. I have added a validation for due dates in past, i'll keep this in WIP until it's confirmed that it is ok.

robertavram:
We don;t want to be able to delete Reporting Requirements 1 by 1... as it can create gaps in the reporting periods. the only way to delete reporting requirements is to change the entire list, and the validation should run on the entire list to make sure there is no non-intentional gap in reporting periods.. We'll keep this branch but close the PR as we might have to adapt sometime soon.

csabadenes:
@robertavram Ok. I guess the delete related code could be moved to the `post` method and called on the requirements which are not present in the list received from the frontend.

PR #1494, opened 2018-05-17
Check all requirements for security vulnerabilities

vkurup:
[ch5159]

This checks the packages in the currently active virtualenv. It's also possible to check packages based on a requirements.txt file, in case we only want to check production requirements, for example.

I expect this to fail CI because of the pysaml issue: clubhouse 5204

PR #1604, opened 2018-05-30
Check requirements for security vulnerabilities

vkurup:
[ch5159]

This brings back #1494 because the online service does not seem to be checking each github commit.


vkurup:
CI fails as expected because of PySAML.

domdinicola:
I think we can close this.
We can use the 'pipenv check' functionality

PR #1613, opened 2018-05-31
[ch4233] tweak partner blocking

domdinicola:


domdinicola:
1. you can mark manually hidden in the admin
2. the hidden flag is an set every time if one of the three flags is set (deleted, blocked, manually blocked)

PR #1615, opened 2018-05-31
Allow country to be overridden by get param

reinbach:
[ch6067]

The idea is that when making an API request you can specify the country/schema you want to query against, and not worry about which country you are currently logged into.

This is more for those users at the regional level, who want to query multiple countries via the API for CSV exports, without having to change their Country via the UI.

Any request with `country` get param will check if country exists and part of the available countries for the user, if so then set the connection to that country


src/etools/applications/EquiTrack/utils.py
```diff
@@ -30,8 +31,22 @@ def get_current_site():
 
 
 def set_country(user, request):
+    from etools.applications.users.models import Country
+
+    country = request.GET.get("country", None)
```
domdinicola: it means that we cannot use country as filter in other views.. do you think we can use another name for the query param?

reinbach: We sure can... maybe `schema`?
It does not look like we have any fields labelled that, that I can see.

domdinicola: yeah maybe schema, then I would add also schema in the get L40

ntrncic:

src/etools/applications/EquiTrack/utils.py
```diff
@@ -30,8 +30,22 @@ def get_current_site():
 
 
 def set_country(user, request):
+    from etools.applications.users.models import Country
+
+    country = request.GET.get(settings.SCHEMA_OVERRIDE_PARAM, None)
+    if country:
+        try:
+            country = Country.objects.get(
+                Q(name=country) | Q(country_short_code=country)
```
domdinicola: I would add 'schema' as well

PR #1635, opened 2018-06-07
[ch6198] HACT: generate hact every day and removed static methods

domdinicola:


src/etools/applications/partners/models.py
```diff
@@ -510,16 +510,15 @@ def latest_assessment(self, type):
...
         try:
-            self.hact_values = json.loads(self.hact_values) if hact_is_string else self.hact_values
+            self.hact_values = self.get_hact_json()
+            self.hact_values = json.dumps(self.hact_values, cls=EToolsEncoder)
         except ValueError as e:
             e.args = ['hact_values needs to be a valid format (dict)']
             raise e
 
         super(PartnerOrganization, self).save(*args, **kwargs)
-        if hact_is_string:
```
domdinicola: I don't think we need this after the save...

domdinicola:
@reinbach I think we can specify the default encoder only from DJ1.11 :(

reinbach:
Well that doesn't help us then! :)

src/etools/applications/partners/models.py
```diff
@@ -761,39 +758,50 @@ def spot_checks(cls, partner, event_date=None, update_one=False):
...
+    def hact_support(self):
+        from etools.applications.audit.models import Audit, Engagement
+
+        hact = self.get_hact_json()
+        audits = Audit.objects.filter(partner=self, status=Engagement.FINAL,
+                                      date_of_draft_report_to_unicef__year=datetime.now().year)
+        hact['outstanding_findings'] = sum([
+            audit.pending_unsupported_amount for audit in audits if audit.pending_unsupported_amount])
+        hact['assurance_coverage'] = self.assurance_coverage
+        self.hact_values = json.dumps(hact, cls=EToolsEncoder)
```
reinbach: No need to set value with `json.dump` let the `save()` method handle it, as it currently does.
Also provides for a simpler update when we upgrade to Django 1.11 and start using the `encoder` on `JSONField`

src/etools/applications/partners/models.py
```diff
@@ -669,23 +667,23 @@ def programmatic_visits(cls, partner, event_date=None, update_one=False):
             hact['programmatic_visits']['completed'][quarter_name] = pvq
             hact['programmatic_visits']['completed']['total'] = pv
         else:
-            pv_year = TravelActivity.objects.filter(
-                travel_type=TravelType.PROGRAMME_MONITORING,
-                travels__traveler=F('primary_traveler'),
-                travels__status__in=[Travel.COMPLETED],
-                travels__end_date__year=timezone.now().year,
-                partner=partner,
+            pv_year = Travel.objects.filter(
```
domdinicola: I needed to change this.
If a Trip has two activities with end date in different quarters it's counted twice.

PR #1636, opened 2018-06-08
Feature/removes custom django storarge package

saxix:
this PR removes the needs of custom django-storages package, using only a custom backend with the standard library. Included tests need azure account (they are skipped in case is missing).


robertavram:
👍

PR #1642, opened 2018-06-11
fix 'createsuperuser' failing due to not asking for the 'email'

csabadenes:


csabadenes:
@reinbach You are welcome.

[ch6202]

domdinicola:
I think we do have a problem also with the admin form @csabadenes @reinbach

csabadenes:
@domdinicola sorry, missed your message.. should be already fixed in https://github.com/unicef/etools/pull/1644

PR #1655, opened 2018-06-15
Another attempt to fix deleting reporting requirements from the backend

csabadenes:


src/etools/applications/partners/serializers/interventions_v2.py
```diff
@@ -768,6 +768,11 @@ def create(self, validated_data):
 
     def delete(self, validated_data):
         for r in validated_data["reporting_requirements"]:
+            if r.get("start_date") < date.today():
```
robertavram: please make start_date, end_date not nullable in the db before checking this... r.get can return None and then an error will be thrown when comparing None to Date

csabadenes: @robertavram Looks like null start dates are allowed for HR report  type, i think i should change the validation rules, because checking for start date makes sense only for QPR. If i make start_date not nullable, it will break HR report type.

robertavram:
see comment in the update validation PR... I think we should avoid deletion of reporting periods (unless last one in the cycle?) otherwise the reporting periods will have gaps and that;s not allowed

robertavram:
@csabadenes please add that validation that checks for gaps in reporting periods if someone tries to delete a reporting period.

csabadenes:
@robertavram Ok

csabadenes:
@robertavram I think it is Ok to leave gaps in the list strictly when deleting, because the deleted gaps should be valid after the user edits the list on the frontend. I've put the two operations in a transaction, so the delete should not happen if the modified list is not valid.

Is it allowed to delete all of the requirements from the list? Currently this is disallowed by validations, and i do not know if it's Ok like that.

csabadenes:
[ch6583]

robertavram:
this was done in some other pr

PR #1667, opened 2018-06-20
Audit & TPM fixes

roman-karpovich:
fixed https://trello.com/c/IzCqL3g0/168-bug-500-error-relatedobjectdoesnotexist-when-calling-api-tpm-visits-pk

### DIFF
--- a/src/etools/applications/audit/serializers/engagement.py
+++ b/src/etools/applications/audit/serializers/engagement.py
@@ -405,6 +405,21 @@ class AuditSerializer(ActivePDValidationMixin, RiskCategoriesUpdateMixin, Engage
     def get_number_of_financial_findings(self, obj):
         return obj.financial_finding_set.count()
 
+    def _validate_financial_findings(self, validated_data):
+        financial_findings = validated_data.get('financial_findings')
+        audited_expenditure = validated_data.get('audited_expenditure')
+        if not financial_findings:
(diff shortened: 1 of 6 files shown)


src/etools/config/settings/base.py
```diff
@@ -100,6 +100,7 @@ def get_from_secrets_or_env(var_name, default=None):
 # DJANGO: FILE UPLOADS
 MEDIA_ROOT = os.environ.get('MEDIA_ROOT', '/tmp/etools/media/')
 MEDIA_URL = '/media/'
+DEFAULT_FILE_STORAGE = 'etools.applications.utils.common.storage.EtoolsFileSystemStorage'
```
ntrncic: we use AZURE STORAGE for files, otherwise it's local file storage. We don't care what country the files are coming from as it's captured in the filepath. I don't think we need this.

src/etools/applications/attachments/models.py
```diff
@@ -29,10 +30,11 @@ class Meta:
 
 
 def generate_file_path(attachment, filename):
-    return 'files/{}/{}/{}/{}'.format(
+    return 'files/{}/{}/{}/{}/{}'.format(
```
ntrncic: we do not need the timestamp as a directory. let the filenames be changed by Azure.

roman-karpovich:
@ntrncic reverted commits related to storage update

PR #1680, opened 2018-06-25
upgrade requirements + refactoring

domdinicola:


.pyup.yml
```diff
@@ -10,5 +10,5 @@ search: False
 
 # specify the requirement files
 requirements:
-  - src/requirements/production.txt
+  - src/requirements/base.txt
   - src/requirements/local.txt
```
vkurup: We should probably remove "local.txt" as well (though we can probably just remove this whole file since we're not using the hosted Pyup service)

domdinicola: sorry @vkurup I didn't get it...

vkurup: Oh, just two things:
1. This PR deletes local.txt from the repo, so we should delete line 14 from this file.
2. We're not actually using pyup.io anymore, so this entire file can be deleted (but that's not directly related to this PR)

domdinicola: Now I get it!
1. I changed to test.txt (develop and test requirements are there now)
2. agree is another task

PR #1687, opened 2018-06-26
Snapshot library

reinbach:
[ch6354]

Created snapshot library package, and now making use of it.
Once merged and released to prod, will be able to remove the `snapshot` app within etools. Required for now to migrate data to new package tables.


domdinicola:
do you think we can run a command running SQL to copy the table for each schema?

reinbach:
As opposed to running the migration that is there?

domdinicola:
in order to use the library in etools, I guess we need a migration command to be run only once in etools (no need to be in the library)

reinbach:
The migration is in the etools snapshot app, not in the library, and is the reason why that app is being kept around till it has been run. So once it has run in production, we will be removing the etools snapshot app completely, and so migration will be removed as well. Library stays clean :)

domdinicola:
nice! I didn't see it! :)

PR #1688, opened 2018-06-26
Enable/readd start date for HR reporting requirements

csabadenes:
My understanding of the requirements were:
 - start date needs to be re-enabled for HR
 - only the first HR has a start date, which is mandatory
 - none of the other than the first HR's can have a start date

Also added another safety check for both QPR and HR.


csabadenes:
[ch6314]

src/etools/applications/partners/serializers/interventions_v2.py
```diff
@@ -645,19 +645,19 @@ class Meta:
...
-                    "start_date": _(
-                        "Start date needs to be on or after PD start date."
-                    )
-                }
-            })
+        self._validate_start_date(requirements)
 
         # Ensure start date is after previous end date
         for i in range(1, len(requirements)):
+            if requirements[i]["start_date"] >= requirements[i]["end_date"]:
```
reinbach: You could break this `if` statement into a method and call it from here and line 692. Just a suggestion.

csabadenes: I have changed few things here, including your suggestion, thanks.

src/etools/applications/partners/serializers/interventions_v2.py
```diff
@@ -645,6 +645,26 @@ class Meta:
         fields = ("reporting_requirements", "report_type", )
 
     def _validate_qpr(self, requirements):
+        self._validate_start_date(requirements)
+        self._validate_date_intervals(requirements)
```
robertavram: I think we need another validation to validate continuity... Like there should be no brakes between reporting period... Hence the deletion of reporting periods should not be allowed if the reporting period being deleted is in the middle of the cycle.

robertavram:
@csabadenes

PR #1696, opened 2018-06-29
Update psycopg2 requirement to use binary package

reinbach:
[ch6519]

Still being required by some sub packages, but as those get updated the old requirement will be removed.


domdinicola:
why don't you update this pr https://github.com/unicef/etools/pull/1680 ?

reinbach:
will do

PR #1709, opened 2018-07-06
[ch1877] Core Value Assessment refactoring

domdinicola:
[ch1877]


src/etools/applications/attachments/migrations/0008_auto_20180711_1859.py
```diff
@@ -0,0 +1,33 @@
...
+# Generated by Django 1.10.8 on 2018-07-11 18:59
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def migrate_core_value_assessment(apps, schema_editor):
+    Attachment = apps.get_model('attachments', 'Attachment')
+    PartnerOrganization = apps.get_model('partners', 'PartnerOrganization')
+    attachments = Attachment.objects.filter(content_type__app_label='partners',
```
reinbach: FYI this works fine in this instance as there is only 1 attachment relation to partner organization, but there are other objects that have more than 1 relation to attachments. So you probably want to get more specific and include the `attachment.code` when filtering like this. Just to prevent any surprises.

src/etools/applications/partners/tasks.py
```diff
@@ -288,11 +288,12 @@ def pmp_indicator_report(writer, **kwargs):
     for country in qs:
         connection.set_tenant(Country.objects.get(name=country.name))
         logger.info(u'Running on %s' % country.name)
-        for partner in PartnerOrganization.objects.filter():
+        for partner in PartnerOrganization.objects.prefetch_related('core_values_assessments'):
             for intervention in Intervention.objects.filter(
                     agreement__partner=partner).select_related('planned_budget'):
                 planned_budget = getattr(intervention, 'planned_budget', None)
                 fr_currencies = intervention.frs.all().values_list('currency', flat=True).distinct()
+                has_assessment = True if getattr(partner.current_core_value_assessment, 'attachment_file', None) else False
```
reinbach: You could probably simplify this with 
`bool(getattr(partner.current_core_value_assessment, 'attachment_file', False))`

src/etools/applications/partners/tests/test_commands.py
```diff
@@ -45,6 +46,9 @@ def setUpTestData(cls):
         cls.partner = PartnerFactory(
             core_values_assessment="sample.pdf"
         )
+        cls.core_values_assessment = CoreValuesAssessmentFactory(
+            core_values_assessment="sample.pdf"
```
reinbach: This should probably be;

`attachment="sample.pdf"`

src/etools/applications/partners/tests/test_tasks.py
```diff
@@ -774,13 +775,13 @@ def setUpTestData(cls):
         cls.file_type_partner = AttachmentFileTypeFactory(
             code="partners_partner_assessment"
         )
-        cls.partner = PartnerFactory(
+        cls.core_value_assessment = CoreValuesAssessmentFactory(
             core_values_assessment="sample.pdf"
```
reinbach: This should probably be changed to;

`attachment="sample.pdf"`

src/etools/applications/partners/tasks.py
domdinicola: :+1:

src/etools/applications/partners/tests/test_commands.py
domdinicola: absolutely

src/etools/applications/partners/tests/test_tasks.py
domdinicola: absolutely

src/etools/applications/attachments/migrations/0008_auto_20180711_1859.py
domdinicola: partners_partner_assessment right?

reinbach: Yep

domdinicola:
yes.. you should use set the attachment_file property

PR #1710, opened 2018-07-09
Fix warnings inside api inspector with updated drf version.

roman-karpovich:
https://trello.com/c/vUeKVV7F/171-warning-in-drf-upgrade


roman-karpovich:
@reinbach this is enough common functionality, which doesn't depends directly from permissions. custom schema inspector just allows to raise api exceptions inside view get_serializer method.

PR #1713, opened 2018-07-10
removes internal 'locations' and use 'unicef_locations' package

robertavram:


src/requirements/base.txt
```diff
@@ -95,6 +95,7 @@ tablib==0.12.1            # via django-import-export
 tenant-schemas-celery==0.1.7
 tornado==4.2.0            # via flower
 unicef_notification==0.2.0
+unicef-locations>=1.1.0
```
domdinicola: can you update the base.in ?

domdinicola: never mind I did it :)

PR #1726, opened 2018-07-16
Add special conditions to export

robertavram:


robertavram:
weird, export labels are not behaving as expected.. @reinbach thoughts? "special conditions pca" is not getting the label right..

robertavram:
@reinbach apparently Dom's last commit fixed it... I thought Header Labels were coming from the custom CSV Renderer, apparently comes out of verbose_name on the model field... who would have thunk :/

PR #1728, opened 2018-07-18
Section Migration: part 4 [do not merge]

domdinicola:


src/etools/applications/partners/migrations/0015_ref_number_year.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+def populate_agreement_ref_num(apps, schema):
+    import re
```
reinbach: How come these imports are inside the function?

src/etools/applications/partners/migrations/0015_ref_number_year.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def populate_intervention_ref_num(apps, schema):
+    import re
+    from django.db import IntegrityError
+    Intervention = apps.get_model("partners", "Intervention")
+
+    for i in Intervention.objects.exclude(document_type="SSFA"):
+        find = re.findall(r'\d{4}', i.number)
+        try:
```
reinbach: This section appears repetitive, you could move this into a function that takes `status_list` parameter. Granted it is just a data migration.

src/etools/applications/partners/migrations/0016_auto_20180717_0700.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('partners', '0015_ref_number_year'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
reinbach: You could merge this migration with `src/etools/applications/partners/migrations/0014_auto_20180716_1547.py`

src/etools/applications/partners/models.py
```diff
@@ -2060,8 +2070,11 @@ def save(self, force_insert=False, **kwargs):
             # to create a reference number we need a pk
             super(Intervention, self).save()
             self.update_reference_number()
+        elif self.status == self.DRAFT:
+            self.update_reference_number()
 
-        self.update_ssfa_properties()
+        if save_from_agreement is False:
```
reinbach: maybe `if not save_form_agreement:`  is more pythonic?

src/etools/applications/partners/migrations/0015_ref_number_year.py
domdinicola: I think that after the migration run in production we can remove the code in the datamigrations to speed up tests

domdinicola: @robertavram any reason?

src/etools/applications/partners/migrations/0015_ref_number_year.py
domdinicola: I think that after the migration run in production we can remove the code in the datamigrations to speed up tests

PR #1739, opened 2018-07-25
Add internal_prc_review attachment to InterventionAmendment

reinbach:
[ch6476]


reinbach:
@acory you want to upload the file at;
`/api/v2/attachments/upload/`
and the resulting `id` provided from that request to be used to update the intervention amendment `internal_prc_review` value with that id.

PR #1746, opened 2018-07-26
Action points updates

roman-karpovich:
updated fields labels according to latest wireframes.
added missing cc's to emails
category not required in action points dashboard


src/etools/applications/action_points/models.py
```diff
@@ -217,7 +217,7 @@ def get_mail_context(self, user=None, include_token=False):
             'object_url': self.get_object_url(user=user, include_token=include_token),
         }
 
-    def send_email(self, recipient, template_name, additional_context=None):
+    def send_email(self, recipient, template_name, additional_context=None, cc=None):
```
domdinicola: why can't be default []?

roman-karpovich: lists are mutable, so they are not recommended to be used as default in kwargs
https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/

PR #1747, opened 2018-07-26
Add middle name to user model

reinbach:
[ch6744]

Is middle name able to be sync'd from VISION, if so, what is the value of the parameter?


src/etools/applications/users/models.py
```diff
@@ -54,8 +55,12 @@ def get_full_name(self):
...
         Return the first_name plus the last_name, with a space in between.
         """
-        full_name = '%s %s' % (self.first_name, self.last_name)
-        return full_name.strip()
+        full_name = ' '.join([
+            self.first_name,
+            self.middle_name,
+            self.last_name,
+        ])
+        return full_name.strip().replace("  ", " ")
```
robertavram: why are we replacing double-spaces? there shouldn't be any double-spaces right?

robertavram: nevermind.. just realized there could be no middle name!

PR #1754, opened 2018-07-27
Switch action points categories in dependency from related module.

roman-karpovich:


roman-karpovich:
frontend changes are here
https://github.com/unicef/etools-action-points/pull/13

PR #1760, opened 2018-07-31
Move action points categories into public schema

roman-karpovich:
https://trello.com/c/YAR2rPBk/25-move-category-model-to-an-inner-global-app-inside-actionpoint-please-use-timestampedmodel


src/etools/applications/action_points/migrations/0007_auto_20180731_0920.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.8 on 2018-07-31 09:20
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def copy_categories(apps, schema_editor):
```
domdinicola: I hope we can avoid this since this is not in production yet

roman-karpovich:
@reinbach this model is tightly connected with action points, so i've created sub-app to move it into public schema.

reinbach:
That makes sense, if only I had read the title of the PR! :)

PR #1761, opened 2018-07-31
upgrade location library

domdinicola:


domdinicola:
@roman-karpovich can you please check if the problem is solved?

PR #1783, opened 2018-08-08
[ch6025] fixed InterventionList result filter

domdinicola:
[ch6025]
fixed InterventionList result filter
removed unused function


domdinicola:
@robertavram should we use 'agreement__country_programme' or 'country_programme'.
the serializer is using 'country_programme'

PR #1789, opened 2018-08-10
[ch7076] removed signed by unicef field

domdinicola:


src/etools/applications/partners/migrations/0020_auto_20180719_1815.py
```diff
@@ -5,26 +5,11 @@
 from django.db import migrations
 
 
-def update_intervention_signed_by(apps, schema):
```
domdinicola: this migration did run already

PR #1791, opened 2018-08-10
Termination and Contingency

robertavram:


acory:
@robertavram There might be a bug on this PR , on save PD (without changing anything related to termination) I get `Cannot change fields while in draft: termination_doc`. Can you check?

robertavram:
are you trying to change termination_doc? can you check the post request?

acory:
No. One easy way to reproduce:
Go to existing Draft PD that already has start and end date , do not modify anything, hit Save.
Request Payload is `{}`, Response is `Cannot change fields while in draft: termination_doc`

robertavram:
I can;t reproduce it for some reason.. what Front End branch should i use?

robertavram:
@acory I've changed a few things on this branch
if you haven't run migrations yet you're good.. otherwise you might have to dump db and restore

I'll share a recent db with you

acory:
FE branch feature--pd-termination. Already ran migrations..

Issue #924, opened 2017-10-05
Conflicting views for deleting ResultLinks


caktus-philip:
At present in the `prp-refactoring` branch, there are 2 different API endpoints for deleting a ResultLink, and they differ significantly in when they do/don't allow deletion. There are 2 different URLs and 2 different views. 

The views are `InterventionResultLinkDeleteView` and `InterventionResultLinkUpdateView`. Here's their implementation -- 
https://github.com/unicef/etools/blob/607c042872b281325806e84b5ce3adaf5a53e18d/EquiTrack/partners/views/interventions_v2.py

Here's the URLs that point to them --
https://github.com/unicef/etools/blob/607c042872b281325806e84b5ce3adaf5a53e18d/EquiTrack/partners/urls_v2.py#L136

https://github.com/unicef/etools/blob/607c042872b281325806e84b5ce3adaf5a53e18d/EquiTrack/partners/urls_v2.py#L105

In a Slack discussion on 5 October 2017, Rob and Nik allowed that there should be only one endpoint, and that while  `InterventionResultLinkUpdateView` isn't quite correct, it's the better of the two and should be preserved while the other is deleted. 

Rob said "[InterventionResultLinkUpdateView] is ok... but missing some validation.... one is status... intervention can only be in draft to delete this... if it's not in draft.. it gets a bit hairy on the deletion validation...

for now... just use the logic that's in there... don't worry about statuses.. and we just enforce that on the frontend

we'll fix later on
"

Nik said, "found this in my notes: to fix this we can add a in_amendment flag on intervention then set it to true when adding amendment, to remove flag we"

PR #1793, opened 2018-08-13
Remove duplicate delete result link endpoint

reinbach:
[ch6745]

Removed `InterventionResultLinkDeleteView` endpoint

`api/v2/interventions/results/(?P<pk>\d+)/` is no longer valid


domdinicola:
since there are 2 endpoints for this.. are we sure we are deleting the right one? I mean, is this used in the frontend?

reinbach:
@domdinicola removing the one as referenced in the ticket to be removed

domdinicola:
https://github.com/unicef/etools/issues/924

PR #1806, opened 2018-08-16
log AD Sync Tasks

domdinicola:


domdinicola:
[ch7180]

src/etools/libraries/azure_graph_api/tasks.py
```diff
@@ -28,28 +32,46 @@ def sync_user(username):
...
     logger.info('Azure Delta Sync Process started')
-    url = cache.get(
-        AZURE_GRAPH_API_USER_CACHE_KEY,
-        '{}/{}/users/delta?$top={}'.format(
-            settings.AZURE_GRAPH_API_BASE_URL,
-            settings.AZURE_GRAPH_API_VERSION,
-            settings.AZURE_GRAPH_API_PAGE_SIZE
+    log = VisionSyncLog(country=Country.objects.get(schema_name="public"), handler_name='UserADSyncDelta')
+    try:
+        url = cache.get(
```
reinbach: How come we cache this `url` and not in `sync_all_users`? maybe move this `url` param into a central location that both of these functions get/set the url?

Actually the only difference besides the usage of cache, appears to be the `handler_name` parameter passed to `VisionSyncLog`

domdinicola: the urls are slightly different there's the '/delta' for the delta task

reinbach: ah missed that

PR #1807, opened 2018-08-17
[ch6934] Return 0 instead of None as default in engagement pending_unsupported_amount

roman-karpovich:


src/etools/applications/audit/models.py
```diff
@@ -410,7 +410,7 @@ def pending_unsupported_amount(self):
             return self.total_amount_of_ineligible_expenditure - self.additional_supporting_documentation_provided \
                 - self.justification_provided_and_accepted - self.write_off_required
         except TypeError:
-            return None
+            return 0
```
domdinicola: no... I would set a default on the model, cause if only one of the fields in None it would get an exception.
total_amount_of_ineligible_expenditure = 40.00
additional_supporting_documentation_provided = 10
justification_provided_and_accepted = 5
write_off_required = None
should return 25.0

domdinicola:
@roman-karpovich is this ready?

domdinicola:
[ch6934]

roman-karpovich:
seems likes that

пт, 24 авг. 2018 г., 21:36 Domenico <notifications@github.com>:

PR #1833, opened 2018-08-29
[ch6478] tweaked admin and added updated groups json

domdinicola:


domdinicola:
@robertavram addressed

PR #1835, opened 2018-08-30
[ch7422] removed trips ap module

domdinicola:
[ch7422]


ntrncic:
Putting this on hold, it is too soon to remove the entire AP model. we had just migrated action points in the last sprint. Let's revisit this in a October sprint.

PR #1854, opened 2018-09-12
Add management command to update attachment seq

reinbach:
[ch7624]

The issue is that the sequence is out of balance with the records in the attachment table. This was due to the records being copied/transferred from old attachment table with `id` value set in order to maintain relationship to the various objects.

This PR provides a management command that resets the sequence values. Run it with;

`python manage.py update_attachment_seq`

You can also copy parts of the management command (`src/etools/applications/attachments/management/commands/update_attachment_seq.py`) and run it in a django shell in another environment if needed.

### DIFF
--- a/src/etools/applications/attachments/management/commands/update_attachment_seq.py
+++ b/src/etools/applications/attachments/management/commands/update_attachment_seq.py
@@ -1,0 +1,25 @@
+from django.core.management.base import BaseCommand
+from django.db import connection
+from django.db.models import Max
+
+from etools.applications.utils.common.utils import run_on_all_tenants
+
+from unicef_attachments.models import Attachment
+
+
+class Command(BaseCommand):
+    """Update seq value for attachment model"""
+
+    def run(self):
+        # get max id for attachments
+        # update seq value to match this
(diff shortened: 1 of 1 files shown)


ntrncic:
I wouldn't put this into the code. essentially we should just run this once. Could you run it in develop and see if it works? @reinbach

reinbach:
@ntrncic 

I ran the following in the terminal on dev without issue;

```
from django.db import connection
from django.db.models import Max
from etools.applications.utils.common.utils import run_on_all_tenants
from unicef_attachments.models import Attachment

def reset_seq():
    row = Attachment.objects.aggregate(max_id=Max("id"))
    cursor = connection.cursor()
    cursor.execute(
        "SELECT setval('unicef_attachments_attachment_id_seq', {}, TRUE)".format(
            row["max_id"]
        )
    )

run_on_all_tenants(reset_seq)```

Which section in the front end is making use of this attachment flow?

ntrncic:
awesome. I think it could be TPM or Auditor portal using it

reinbach:
Mmm... I added attachments in both those sections in multiple places, but alas it did not appear to use the relevant attachment url endpoint. I think I'm still not hitting the correct section.

PR #1865, opened 2018-09-14
Add docs on unicef libraries in use in etools

reinbach:
[ch7464]

An initial start, pretty basic doc. Think it's best to keep the README of each library up to date about the usage, and functionality of the particular library.

Let me know what other information we would like to see in this library doc page.

### DIFF
--- a/docs/libraries.rst
+++ b/docs/libraries.rst
@@ -1,0 +1,74 @@
+Libraries
+=========
+
+Various libraries that are in use/available and their functionality;
+
+unicef-locations
+----------------
+
+https://github.com/unicef/unicef-locations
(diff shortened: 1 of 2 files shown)


docs/libraries.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Django package that handles file attachments for models.
+
+
+djangosaml2
+-----------
+
+https://github.com/unicef/djangosaml2
+
+A maintenance fork of the original and no longer maintained djangosaml2 library.
```
domdinicola: I would say an "adapted" version. The original repo is maintained

PR #1895, opened 2018-10-06
More location remap quickfixes

csabadenes:
[ch7983]

Changes & fixes done in this PR:
 - revalidate the MPTT location tree before starting the location import
 - try to refetch failed pages when requesting paginated data from CartoDB
 - remap locations row by row and not in one batch, and allow the location import to work with archived locations too. These changes make possible more complicated remap scenarios(encountered in Sudan, for example), where the same location is remapped to a second location before a third locations is remapped back to it.
 - write-lock location table rows before the import, to prevent deadlock scenarios caused by another process trying to write the same data
 - few extra checks in the location cleanup step


csabadenes:
@ntrncic, the CH card is #7983

src/etools/libraries/locations/tasks.py
```diff
@@ -168,26 +173,28 @@ def update_sites_from_cartodb(carto_table_pk):
...
-                    Location.objects.filter(p_code__in=list(orphaned_old_pcodes)).update(is_active=False)
+                    Location.objects.filter(
+                        p_code__in=list(orphaned_old_pcodes),
+                        is_active=True,
+                    ).update(
+                        is_active=False
+                    )
 
             # rebuild location tree
             Location.objects.rebuild()
```
robertavram: can you also move this into the transaction?

src/etools/libraries/locations/tasks.py
```diff
@@ -225,21 +233,29 @@ def cleanup_obsolete_locations(self, carto_table_pk):
     remapped_pcodes += [remap_row['new_pcode'] for remap_row in remapped_pcode_pairs]
     deleteable_pcodes = set(database_pcodes) - (set(new_carto_pcodes) | set(remapped_pcodes))
 
+    revalidated_deleteable_pcodes = []
     for deleteable_pcode in deleteable_pcodes:
         try:
-            deleteable_location = Location.objects.all_locations().get(p_code=deleteable_pcode)
+            deleteable_location = Location.objects.all_locations().get(
+                p_code=deleteable_pcode,
+                gateway=carto_table.location_type,
```
robertavram: can we let this fail in case there are multiple duplicated p_codes in the same country?
We've talked to Jan and we understand this is a possibility currently but would rather fix it than accommodate for it.

src/etools/libraries/locations/tasks.py
```diff
@@ -225,21 +233,29 @@ def cleanup_obsolete_locations(self, carto_table_pk):
...
+            )
         except Location.DoesNotExist:
             logger.warning("Cannot find orphaned pcode {}.".format(deleteable_pcode))
         else:
             if deleteable_location.is_leaf_node():
-                logger.info("Deleting orphaned and unused location with pcode {}".format(deleteable_location.p_code))
-                deleteable_location.delete()
+                secondary_parent_check = Location.objects.all_locations().filter(parent=deleteable_location.id)
```
robertavram: could you use .exists() method

src/etools/libraries/locations/tasks.py
```diff
@@ -225,21 +233,29 @@ def cleanup_obsolete_locations(self, carto_table_pk):
...
+                remap_history_check = LocationRemapHistory.objects.filter(
+                    Q(old_location=deleteable_location) | Q(new_location=deleteable_location)
+                )
```
robertavram: .exists()

src/etools/libraries/locations/tasks.py
```diff
@@ -225,21 +233,29 @@ def cleanup_obsolete_locations(self, carto_table_pk):
     remapped_pcodes += [remap_row['new_pcode'] for remap_row in remapped_pcode_pairs]
     deleteable_pcodes = set(database_pcodes) - (set(new_carto_pcodes) | set(remapped_pcodes))
 
+    revalidated_deleteable_pcodes = []
```
robertavram: can you add a comment on what you;re achieving with "revalidated_deletable_pcodes"?

src/etools/libraries/locations/tasks.py
csabadenes: I think it is inside the transaction, it's hard to see with that many indents

csabadenes:
@robertavram , i think i have addressed the requested changes

PR #1902, opened 2018-10-18
[ch8162] old sdk: upgrade sentry

domdinicola:
old_sdk


ntrncic:
@domdinicola  Please provide a clubhouse card so it can be tested when it goes to develop and staging to make sure we are still getting sentry errors

domdinicola:
[ch8162]

PR #1904, opened 2018-10-22
[ch7881] POSynchronizer_removed_donor_and_grant

domdinicola:


domdinicola:
[ch7881]

ntrncic:
fix tests. why are we removing donor and grant?

PR #1905, opened 2018-10-22
7841 changed PDF PCA filename

domdinicola:


domdinicola:
[ch7841]

src/etools/applications/partners/views/v1.py
```diff
@@ -39,18 +43,18 @@ def get_context_data(self, **kwargs):
             return {"error": "Cannot find document with given query parameter lang={}".format(lang)}
 
         try:
-            agreement = Agreement.objects.get(id=agr_id)
+            self.agreement = Agreement.objects.get(id=agr_id)
```
ntrncic: why use self here?

domdinicola: I set self.agreement cause I can use it in get_pdf_filename method

ntrncic:
@domdinicola why is is for next sprint?

src/etools/applications/partners/views/v1.py
```diff
@@ -28,6 +28,10 @@ class PCAPDFView(LoginRequiredMixin, PDFTemplateView):
         "ifrc_french": "partners/pca/ifrc_french_pdf.html"
     }
 
+    def get_pdf_filename(self):
+        return '{0.reference_number}-{0.partner}-{1.profile.country.schema_name}.pdf'.format(
```
robertavram: use agreement_number here instead of reference_number

PR #1908, opened 2018-10-23
allow proper packaging

saxix:
fixes setup.py, removes dependency from git+djangosaml2 and uses unicef-djangosaml2 properly available on pypi

This PR allow etools to be used as dependency, enabling datamart upgrades.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,12 +31,13 @@ def get_requirements(env):
                 continue
             line = line[:-1]
             if line.startswith('-e git+'):
-                url = line.replace('-e ', '')
+                url = line.replace('-e git+', '')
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -45,7 +46,10 @@ def get_requirements(env):
 
 install_requires = get_requirements('base')
 test_requires = get_requirements('test')
-
+# TODO: remove me
+# print(111, "setup.py:49", dependency_links)
+# print(111, "setup.py:49", install_requires)
+print(111, "setup.py:49", test_requires)
```
reinbach: Any reason this has not been removed?

PR #1917, opened 2018-10-26
SSFA Header not move to signed and add year to admin

ntrncic:
[ch7786]
[ch7805]

### DIFF
--- a/src/etools/applications/partners/validation/agreements.py
+++ b/src/etools/applications/partners/validation/agreements.py
@@ -183,6 +183,9 @@ class AgreementValid(CompleteValidation):
         return True
 
     def state_signed_valid(self, agreement, user=None):
+        if agreement.agreement_type == agreement.SSFA and \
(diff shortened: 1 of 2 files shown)


src/etools/applications/partners/validation/agreements.py
```diff
@@ -183,6 +183,9 @@ def state_draft_valid(self, agreement, user=None):
         return True
 
     def state_signed_valid(self, agreement, user=None):
+        if agreement.agreement_type == agreement.SSFA and \
+                agreement.interventions.all().first().status == 'draft':
```
domdinicola: do we assume is always only one comment from SSFA?
we don't need to do .all() if we call .first() after

ntrncic: @domdinicola so there is basic validation that runs first. so if it gets to here then it means that there is only one ssfa for this agreement.

PR #1923, opened 2018-10-30
[ch8232] tpm visit list view added full name to focal point

domdinicola:
[ch8232]


reinbach:
I think you win the award for longest repo name :)

PR #1924, opened 2018-10-30
Fix for result_type filter in OutputListAPIView

roman-karpovich:
Fixed broken filter by result_type.
By default, combining Result.outputs and result_type__name='Outcomes' gives empty list.

```
In [13]: print(Result.outputs.all().filter(result_type__name='Outcome').query)
SELECT ... FROM ... WHERE ("reports_resulttype"."name" = Output AND "reports_resulttype"."name" = Outcome) ORDER BY "reports_result"."name" ASC
```


robertavram:
if you need this for FAM or AP can you create a different view or another version of the API.. this might break clients that rely on output_name given by OutputListSerializer.

roman-karpovich:
> if you need this for FAM or AP can you create a different view or another version of the API.. this might break clients that rely on output_name given by OutputListSerializer.

Yes, we need it for Field Monitoring module to show outcomes in dropdowns, but this is basic functionality, which seems to be broken. I left it unchanged except result_type filter, which is correct now and left filter by outputs by default, as was before. Don't sure if anobody use this special filter, because if so, this should be reported much earlier.

roman-karpovich:
@robertavram Does i understood correctly, that I should make fixed copy of this view inside views/v3 with different url? This seems to be very agressive versioning for such fix, which can make future maintaining much harder.

robertavram:
@roman-karpovich I agree it's aggressive but I feel like there are other clients that might break because the name property from OutputListSerializer was changed. I think some clients rely on the "output_name" format

roman-karpovich:
@robertavram agreed, that's why i added serializer method to retrieve correct name

``
    def get_name(self, obj):
        if obj.result_type == ResultType.OUTPUT:
            return obj.output_name
        else:
            return obj.result_name
```

if it's output, format will be exactly the same as before. else, result name will be provided to avoid assertion error.

robertavram:
OHHHHHH... my bad.. i didn't notice that it's inheriting the methods from minimal...

merging.. 
thanks!

roman-karpovich:
Thanks!

PR #1927, opened 2018-10-31
Django Admin: added git changeset date

domdinicola:
this will allow us to know which is the current version in the environment


robertavram:
I suggest that this is something we add somehow in the docker image build process. @saxix has been simplifying code to not include anything by packaged apps for other apps where you would not have access to git commits. I order to standardize and use better practice, the version number of the app should be added sometime during docker build

PR #1929, opened 2018-10-31
move sentry to tenant_support

domdinicola:


reinbach:
Why is this under `tenant_support`?
I'd think this is code related to `tenant`/`schemas` etc. While sentry is error handling?

domdinicola:
@reinbach correct, but the Sentry Client is only adding the tenant to the informations... what do you think?

reinbach:
My understanding is that we're just adding `tenant` data point to the data information being sent to sentry, and not really making it compatible with tenants. 
I see the possibility that we add more data points over time.

I'd probably move it to the EquiTrack folder.

domdinicola:
I guess if we have to add other information we would extend this class in the EquiTrack folder...
this class in not dependent of etools.. can be used in a generic tenant application

saxix:
IMHO both Domenico and Greg are right. Sentry currently depends on tenant, but it also could depends on other infos of etools. It is a 'cross-cutting' component (like celery - no tasks,  celery config I mean). For this reason it should be in a top level package (etools.sentry or etools.lib.sentry). 
etools does not have a real 'core' package; even if it has been built around  Equitrack, it is more a functional module than a system module.

PR #1947, opened 2018-11-07
admin actions

domdinicola:


robertavram:
Hey Dom, can you write the description of this PR what it's aiming to do, and connect it to a card?

domdinicola:
Hi @robertavram , I wanted to talk to you and @ntrncic  about this, I can show you quickly how does this work.
Basically it add a set of actions to the standard django admin actions, one of them is the "Mass Update". With this we may change at once the same attribute for multiple instances.
I was thinking about this for the mass update of sections.

PR #1948, opened 2018-11-07
[ch7673] country office admin permissions

domdinicola:
[ch7673]


src/etools/applications/EquiTrack/data/groups.json
```diff
@@ -200,27 +200,24 @@
     "fields": {
       "name": "Country Office Administrator",
       "permissions": [
-        738,
+        744,
```
robertavram: I'm not sure what's happening here...  is there a way to write in a comment the changes you made to get this?

domdinicola: basically here we "tailored" manually the permissions needed in staging and dump the fixture of the group requested "Country Office Administrator"

robertavram: Ugh. ok. Merging.. but let's figure out a way to be more explicit on what permissions we're setting in the code.. this is impossible to validate against business requirements from the code :)

PR #1958, opened 2018-11-13
moved FSMTransitionActionMixin in permission2 app [refactoring]

domdinicola:


domdinicola:
@reinbach better now?

PR #1964, opened 2018-11-14
Intervention Attachments

robertavram:
`api/v2/interventions/(?P<intervention_pk>\d+)/attachments/ ` -> List/
Create (GET, POST)
`api/v2/interventions/attachments/(?P<pk>\d+)/ ` -> Update/ Delete
(PATCH, DELETE)

- [x] Endpoints created
- [x] Tests updated


robertavram:
On hold until FE is done, as it PMP will error out without.

PR #1974, opened 2018-11-16
Send suspension notification when agreement suspended

reinbach:
[ch7539]

### DIFF
--- a/src/etools/applications/partners/views/agreements_v2.py
+++ b/src/etools/applications/partners/views/agreements_v2.py
@@ -33,6 +33,7 @@ from etools.applications.partners.serializers.exports.agreements import (
     AgreementExportSerializer,
 )
 from etools.applications.partners.validation.agreements import AgreementValid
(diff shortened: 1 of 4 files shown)


src/etools/applications/partners/views/agreements_v2.py
```diff
@@ -168,6 +169,10 @@ def update(self, request, *args, **kwargs):
             # refresh the instance from the database.
             instance = self.get_object()
 
+        # if agreement becomes suspended, send notification
```
ntrncic: would it be better if it's a post_save signal on the model. What if something goes wrong and the transaction is reverted, the email still gets sent.

reinbach: Would be nice to use the `post_save` signal, but we require the user who is making the update to know who to send the email to. I don't see how we can pass this information to the signal. The updating user information is not on the model/object.

PR #1977, opened 2018-11-19
[ch8647] user sync: fixed user without profile + changed celery queue

domdinicola:
[ch8647]

### DIFF
--- a/src/etools/applications/users/tasks.py
+++ b/src/etools/applications/users/tasks.py
@@ -4,14 +4,13 @@ import json
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group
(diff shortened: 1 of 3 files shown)


src/etools/applications/users/tasks.py
```diff
@@ -139,7 +138,7 @@ def create_or_update_user(self, record):
                 profile = user.profile
             except ObjectDoesNotExist:
                 logger.warning(u'No profile for user {}'.format(user))
-                return
+                profile = UserProfile.objects.create(user=user)
```
reinbach: Any reason not to change this `try/except` to be a `get_or_create` on the `UserProfile`?

ntrncic: I agree, we should do get_or_create unless I'm missing something.

PR #1986, opened 2018-11-26
[Field monitoring] Permissions for settings & planning (simplified)

roman-karpovich:


src/etools/applications/field_monitoring/fm_settings/tests/test_views.py
```diff
@@ -381,11 +482,21 @@ def test_create(self):
         self.assertEqual(len(response.data['methods']), 2)
         self.assertEqual(len(response.data['partners_info']), 1)
 
-    def test_create_without_partner(self):
+    def test_create_unucef(self):
```
reinbach: Looks like method name is misspelled `test_create_unicef`

PR #1994, opened 2018-11-30
Add filter on Special Condition PCA, and add to exports

reinbach:
[ch8843]

Able to filter on agreements with `?special_condition_pca=True`

`special_conditions_pca` now shows up in the agreement export endpoints


acory:
Reminder to allow small case for special_conditions_pca= true  on export, now I get error  ["'true' value must be either True or False."]

reinbach:
@acory special_conditions_pca filtering now allows "true" and "false"

PR #1997, opened 2018-12-03
Change the agreement API to return status as part of the reference number

reinbach:
[ch8833]

`agreement_number` now includes `status` as part of the value


ntrncic:

PR #2001, opened 2018-12-03
[ch8917] djangorestframework 39

domdinicola:
[ch8917]


src/etools/applications/partners/tests/test_api_interventions.py
```diff
@@ -1757,6 +1761,7 @@ def test_create_amendment_when_already_in_amendment(self):
         )
 
     def _make_request(self, user=None, data=None, request_format='json', **kwargs):
+        print(data, 1234)
```
reinbach: Probably want to remove this

domdinicola: ops!

PR #2010, opened 2018-12-04
[ch8880] permission2 fix

domdinicola:


src/etools/applications/audit/transitions/conditions.py
```diff
@@ -97,7 +97,7 @@ def get_errors(self, instance, *args, **kwargs):
 
                 try:
                     model_field = instance._meta.get_field(field)
-                    if not value or value == model_field.default:
+                    if value is None or value == model_field.default:
```
domdinicola: this allow values like 0.0

@roman-karpovich why did you check value == model_field.default?

roman-karpovich: Here we check that value was changed. Else there is no sense to add fields with having default to condition.

PR #2059, opened 2018-12-17
PCA Special Condition export value Yes/No

reinbach:
[ch8895]

Show Yes/No for `special_condition_pca` field in agreements export


domdinicola:
@reinbach why can't we use the friendly render?

PR #2060, opened 2018-12-17
[Field monitoring] csv exports for settings

roman-karpovich:
exports for:
- sites
- log issues
- planned checklist


src/etools/applications/field_monitoring/fm_settings/export/serializers.py
```diff
@@ -0,0 +1,127 @@
...
+    attachments = serializers.SerializerMethodField()
+
+    def get_related_to(self, obj):
+        return dict(obj.RELATED_TO_TYPE_CHOICES).get(obj.related_to_type, '')
+
+    def get_name(self, obj):
+        return str(obj.related_to.name) if obj.related_to else ''
+
+    def get_attachments(self, obj):
+        return ', '.join(map(
```
reinbach: Maybe just me, but I find a list expression easier to read;
`", ".join(["{} - {}".format(a.file_type, urljoin(site_url(), a.url)) for a in obj.attachments.all()])`

roman-karpovich: thanks! i'll update according to the notes.

src/etools/applications/field_monitoring/fm_settings/export/renderers.py
```diff
@@ -0,0 +1,84 @@
...
+            'category': _('Category'),
+            'checklist_item': _('Checklist Item'),
+            'by_partner': _('By Partner'),
+            'specific_details': _('Specific Details'),
+        }
+
+        for m in self.methods:
+            labels['selected_methods.{}'.format(m.name)] = m.name
+
+        for m in self.methods:
```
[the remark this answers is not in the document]
roman-karpovich: in fact, there are two different blocks, and as for me, it will be harder to read if they will be combined. 
we can't just make it planar and write something like 
```
        for m in self.methods:
            labels['selected_methods.{}'.format(m.name)] = m.name
            
            if m.is_types_applicable:
                labels['recommended_method_types.{}'.format(m.name)] = _('Rec. {} types').format(m.name)
```
as far as keys ordering should be preserved to avoid duplicating them inside header property.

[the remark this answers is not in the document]
reinbach: :+1:

src/etools/applications/field_monitoring/fm_settings/export/serializers.py
```diff
@@ -0,0 +1,127 @@
...
+        if len(partners_info) == 0:
+            return ''
+
+        by_partner = any([i.partner for i in partners_info])
+        if not by_partner:
+            return partners_info[0].specific_details
+
+        data = [
+            '{info.partner.name}: {info.specific_details}'.format(info=info)
+            for info in obj.partners_info.all()
```
pitsevich: use partners_info here

PR #2083, opened 2018-12-28
t2f more cleaning

domdinicola:


src/etools/applications/t2f/admin.py
```diff
@@ -28,8 +28,13 @@ class TravelAdmin(admin.ModelAdmin):
...
     )
+    raw_id_fields = (
+        'traveler',
+        'supervisor'
+    )
 
 
+@admin.register(models.TravelActivity)
 class TravelActivityAdmin(admin.ModelAdmin):
     model = models.TravelActivity
```
reinbach: The `model` attribute was removed in the other admin classes, probably want to do that here as well

domdinicola: ops I missed it

PR #2084, opened 2018-12-28
t2f - remove clearance model

domdinicola:


domdinicola:
@ntrncic we need frontend to be adapted

PR #2089, opened 2018-12-31
Refactor t2f permission matrix

reinbach:
[ch9260]
[ch9297]

Permission matrix was a yaml file that needed to be parsed, changes pre-parse the file into a json file, which is used by `PermissionMatrixView` endpoint called by the FrontEnd.

The `PermissionMatrix` now makes use of a permissions dictionary that was generated from the yaml file.

The yaml file has been left in place with the scripts used to pre-parse it. It appears that the yaml file was derived from db tables, which are no longer present. The code around that has been removed.

```
/api/t2f/travels/<pk>/
(/etools.applications.t2f.views.travel:TravelDetailsViewSet.retrieve)
- before: ~5s
- after: ~200ms

/api/t2f/permission_matrix/
(/etools.applications.t2f.views.generics:PermissionMatrixView.get)
- before: ~5.5s
- after: ~60ms
```


src/etools/applications/t2f/permissions.py
```diff
@@ -0,0 +1,7248 @@
+permissions = {'Anyone': {'approved':
```
domdinicola: maybe you can use the status constants here Travel.APPROVED ecc

src/etools/applications/t2f/helpers/permission_matrix.py
```diff
@@ -40,15 +43,28 @@ def get_user_role_list(user, travel=None):
...
 
         with open(path) as permission_matrix_file:
-            permission_matrix = yaml.load(permission_matrix_file.read())
+            permission_matrix = json.loads(permission_matrix_file.read())
         cache.set(PERMISSION_MATRIX_CACHE_KEY, permission_matrix)
 
     return permission_matrix
 
 
+def get_permission_matrix_old():
```
domdinicola: do we still need this?

src/etools/applications/t2f/helpers/permission_matrix.py
```diff
@@ -60,30 +76,15 @@ def __init__(self, travel, user):
...
-
-        permissions = defaultdict(bool)
-        for user_type, um in permission_matrix.items():
-            if user_type not in self._user_roles:
-                continue
-
-            for state, sm in um.items():
-                if state != self.travel.status:
-                    continue
+        perms = defaultdict(bool)
```
domdinicola: :+1: way cleaner

src/etools/applications/t2f/helpers/permission_matrix.py
```diff
@@ -98,3 +99,107 @@ def has_permission(self, permission_type, model_name, field_name):
...
+    Default expected permission is;
+    - `edit`: False
+    - `view`: True
+
+    For each record check if it is different to the default
+    if so, take note
+
+    Structure of permission_matrix is;
+    Travel: {
+       "User": {
```
domdinicola: isn't this the role?

src/etools/applications/t2f/helpers/permission_matrix.py
domdinicola: I mean after the migration to the new one we can get rid of it, right?
should we add a TODO?

src/etools/applications/t2f/permissions.py
reinbach: This is a computer generated file from the old permission matrix. So we're not going to be editing this. Next step would be to summarize this file into something much smaller, then I think we should make use of `Travel.APPROVED` etc.

src/etools/applications/t2f/helpers/permission_matrix.py
reinbach: Yes we can get rid of this once we're happy everything works correctly

PR #2090, opened 2019-01-02
[ch9062] trips change export backend filter 

domdinicola:
[ch9062]


domdinicola:
remove partnership from static

PR #2095, opened 2019-01-03
Update trips to use etools attachments library

reinbach:
[ch9279]

To copy current files to attachments, use the following django command;

`python manage.py copy_t2f_travel_attachments`


ntrncic:
please update migrations and resolve conflict @reinbach

PR #2101, opened 2019-01-07
[CH #9293] Allow Country Save&Delete only in the Global schema

csabadenes:
[ch9293]


domdinicola:
I was thinking to something more light
https://github.com/unicef/etools/pull/2105

csabadenes:
@domdinicola yeah it achieves the same thing but much simpler. i guess this one can be closed then.

PR #2103, opened 2019-01-08
Flake

domdinicola:


src/etools/applications/audit/serializers/risks.py
```diff
@@ -244,7 +244,7 @@ def to_representation(self, instance):
 
     @staticmethod
     def calculate_risk(category):
-        raise NotImplemented()
+        raise NotImplementedError()
```
reinbach: Any reason that this `NotImplementedError` is actually called, relevant to the others?

domdinicola: no idea... I guess we can remove the call

PR #2117, opened 2019-01-14
tenant support library

domdinicola:


src/etools/applications/EquiTrack/util_scripts.py
```diff
@@ -19,25 +18,6 @@ def printtf(*args):
         f.write('\n')
 
 
-def set_country(name):
```
ntrncic: please leave this here.

PR #2122, opened 2019-01-16
Intervention dashboard view performance tweaks

reinbach:
Update intervention dashboard view query to make use of subqueries for the following data points;
- `last_pv_date`
- `has_final_partnership_review`
- `action_points`

`days_last_pv` is calculated in the serializer.

Locally I see an improvement in request time from ~1.6s to ~.5s for the api endpoint `/api/v2/interventions/partnership-dash/`


src/etools/applications/partners/views/dashboards.py
```diff
@@ -64,6 +87,9 @@ def get_queryset(self):
             Count("frs__currency", distinct=True),
             max_fr_currency=Max("frs__currency", output_field=CharField(), distinct=True),
             multi_curr_flag=Count(Case(When(frs__multi_curr_flag=True, then=1))),
+            has_final_partnership_review=Subquery(final_partnership_review_qs),
```
robertavram: Nice!!! I'll give it a nice test bbefore merging... but this is so much better!

PR #2125, opened 2019-01-17
REFACTORING move factory to test library

domdinicola:


domdinicola:
@reinbach do you have any idea of why now test are failing?

PR #2129, opened 2019-01-18
moved sentry client

domdinicola:


src/etools/config/settings/production.py
```diff
@@ -10,7 +10,7 @@
     'dsn': get_from_secrets_or_env('SENTRY_DSN'),  # noqa: F405
 }
 # Override default client, in order to send extra data to Sentry
-SENTRY_CLIENT = 'etools.applications.utils.sentry.client.EToolsSentryClient'
+SENTRY_CLIENT = 'etools.sentry.EToolsSentryClient'
```
reinbach: Looks like the file was moved to `etools/config`, so not sure this is correct.

domdinicola: ops

PR #2152, opened 2019-02-04
Set label for internal prc review type

reinbach:
[ch9232]

We can run the following in each environment, if we don't want a data migrations;

```
from etools.libraries.tenant_support.utils import run_on_all_tenants
from unicef_attachments.models import FileType


def update_internal_prc_review_file_type():
    try:
        file_type = FileType.objects.get(
            code="partners_intervention_amendment_internal_prc_review",
        )
    except FileType.DoesNotExist:
        pass
    else:
        file_type.label = "Internal PRC Review"
        file_type.save()


run_on_all_tenants(update_internal_prc_review_file_type)
```


domdinicola:
Do we want to create a data-migration or we run this once and that's it?

reinbach:
We sure can do that, I don't have the permissions to do that, but have updated the comments in this PR with the necessary snippet of code to do that.
Who ever can run that in the various environments, can then close this PR

PR #2155, opened 2019-02-05
Pmp notifications

reinbach:
[ch9311]


src/etools/applications/partners/utils.py
```diff
@@ -436,3 +437,51 @@ def send_agreement_suspended_notification(agreement, user):
...
+    """Send an email to PD/SHPD/SSFA's focal point(s) if signed
+    and start date is past with no FR added"""
+    frs_count = FundsReservationHeader.objects.filter(
+        intervention=OuterRef("pk")
+    ).annotate(count=Count("vendor_code")).values("count")
+    intervention_qs = Intervention.objects.filter(
+        status=Intervention.SIGNED,
+        start__lt=datetime.date.today(),
+    ).annotate(
+        frs_count=Subquery(frs_count)
```
domdinicola: good to have this on board now

src/etools/applications/partners/utils.py
```diff
@@ -436,3 +437,51 @@ def send_agreement_suspended_notification(agreement, user):
...
+        ]
+        send_notification_with_template(
+            recipients=recipients,
+            template_name="partners/intervention/past-start",
+            context={
+                "reference_number": intervention.reference_number,
+                "title": intervention.title,
+                "url": "{}{}".format(
+                    settings.HOST,
+                    intervention.get_object_url(),
```
robertavram: i think get_object_url is not where we want to send them. 
The url format should look like: "/pmp/interventions/{pk}/details"

src/etools/applications/partners/utils.py
robertavram: can this be done with "frs__isnull=False" on the filter.. I think we should go with that :) @reinbach is having a lot of fun with subqueries :)

src/etools/applications/partners/utils.py
```diff
@@ -436,3 +437,51 @@ def send_agreement_suspended_notification(agreement, user):
             "pd_list": pd_list,  # section, pd_number, link
         }
     )
+
+
+def send_intervention_draft_notification():
+    """Send an email to PD/SHPD/SSFA's focal point(s) if in draft status"""
+    for intervention in Intervention.objects.filter(status=Intervention.DRAFT):
```
robertavram: we should add another filter, to filter only `created < today - 7days`

robertavram:
@reinbach there are a couple of failing tests.. not sure why

PR #2171, opened 2019-02-12
moved helper

domdinicola:


robertavram:
why?

ntrncic:
don't understand why the change. closing this pr.

domdinicola:
it was just a proposal
to put emphases that these are helping function: and since we use it a lot in rancher shell it was easier to do/remember:

from etools.helper import *
set_country('Italy')

PR #2173, opened 2019-02-13
[ch9019] AP export custom comment

domdinicola:
[ch9019]

### DIFF
--- a/src/etools/applications/action_points/export/serializers.py
+++ b/src/etools/applications/action_points/export/serializers.py
@@ -1,7 +1,14 @@
 from rest_framework import serializers
+from unicef_restlib.fields import FunctionRelatedField
+
+
+def comment_export_function(obj):
+    user = obj.user if obj.user else '-'
+    return f'{user} ({obj.submit_date.strftime("%d %b %Y")}) {obj.comment}'
 
 
 class ActionPointExportSerializer(serializers.Serializer):
+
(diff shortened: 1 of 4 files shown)


src/etools/applications/action_points/export/serializers.py
```diff
@@ -22,4 +29,5 @@ class ActionPointExportSerializer(serializers.Serializer):
     related_ref = serializers.CharField(source='related_object.reference_number', read_only=True, allow_null=True)
     related_object_str = serializers.CharField()
     related_object_url = serializers.CharField()
-    action_taken = serializers.SlugRelatedField(source='comments', many=True, read_only=True, slug_field='comment')
+    action_taken = FunctionRelatedField(source='comments', many=True, read_only=True,
```
robertavram: why not add this vanilla DRF? in a class function?

domdinicola: the model (Comment) is in the library 'django_comments'.
about nested... it's an export

the alternative would have been a serializer method iterating the comments, but i thought this FunctionRelatedField would be a nice to have in case we need in other contexts

domdinicola: the alternative would be

action_taken = serializers.SerializerMethodField()

def get_action_taken(self, obj):
  return "; ".join(["{} ({}): {}".format(c.user, c.date, c.comment) for c in obj.comments.all()])

PR #2174, opened 2019-02-14
Intervention attachment admin

reinbach:
[ch9315]

Clean up attachments in partners admin, ensure that attachments show up correctly in frontend, when changed in the admin section.

Issue adding `InterventionAmendment` as inline item to `Interventions`, as we have a inline within inline issue. Still looking to see if there is a work around without adding fields to the `InterventionAmendment` model.


src/etools/applications/partners/admin.py
```diff
@@ -35,14 +33,43 @@
 )
 
 
-class InterventionAmendmentAttachmentFileInline(AttachmentSingleInline):
-    verbose_name_plural = _("Attachment")
+class AttachmentSingleInline(AttachmentSingleInline):
```
domdinicola: maybe we can call this Mixin cause has to be used together with another inline class
also I would add here by default: code=None

reinbach: This is an interim step, I plan to move this code to `unicef-attachments` library.

If I have `code = None`, the code will actually work, I think I would rather have it blow up to prevent incorrect setup. Maybe add a check during `__init__` and raise `ImproperlyConfigured` exception if no code attribute.

domdinicola: ok! fine with me

PR #2185, opened 2019-02-18
update develop with Master

domdinicola:


reinbach:
How come we need to review `master` to `develop` merge? I would think that should automatically happen

domdinicola:
is moving the hotfix in develop

PR #2190, opened 2019-02-19
alpine based docker image

saxix:
new Dockerfile Alpine based

- new 'docker' directory
- fixes static file root directory issue
- add proper entrypoint
- size ~339MB uncompressed
- run services as `etools` user instead of `root`
- optionally run migrations
- updates wait-for-it to works with busybox


docker/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
+
+
+if [[ "$*" == "workers" ]];then
+    celery worker -A etools.config.celery --loglevel=DEBUG --concurrency=4 --purge --pidfile run/celery.pid
+elif [[ "$*" == "beat" ]];then
+    celery beat -A etools.config.celery --loglevel=DEBUG --pidfile run/celerybeat.pid
+elif [[ "$*" == "etools" ]];then
```
robertavram: web, worker, beater, flower

domdinicola:
I like:
- dockerignore
- run not as super user
- entrypoint (commands are under version control)

don't like:
- migrations (we may need to fake)

robertavram:
great idea.. we'll revert to this.. closing for now to clean up as this is a bit outdated.

PR #2193, opened 2019-02-20
[ch9278] Generic django admin fixes

csabadenes:
[ch9278]

 - fix outdated libraries, bad/mistyped filters
 - fix TravelForm, loading the initial value for 'travel reference number' needs to be wrapped in a transaction
 - set required fields for some of the admin forms
 - remove "Add" button from Snapshot, Post Office, Partner Planned Engagement


csabadenes:
Question: I have removed  the 'Add' button from Partner/Planned Engagements, because "partner" is set as read-only, and we cannot add new Engagement without selecting a partner. Therefore the "Add" action makes no sense. 
^ Is this ok?

csabadenes:
The tests should pass after https://github.com/unicef/unicef-attachments/pull/4 is merged

src/etools/applications/t2f/forms.py
```diff
@@ -1,5 +1,21 @@
...
+
+class TravelForm(forms.ModelForm):
+
+    class Meta:
+        model = Travel
+        fields = '__all__'
+
+    def get_initial_for_field(self, field, field_name):
+        if self.initial.get(field_name, field.initial) == make_travel_reference_number:
+            with transaction.atomic():
```
domdinicola: can we move this in the make_travel_reference_number? so we don't need to define a form

src/etools/libraries/snapshot/admin.py
```diff
@@ -0,0 +1,14 @@
+from django.contrib import admin
+
+from unicef_snapshot.admin import ActivityAdmin
+from unicef_snapshot.models import Activity
+
+
+class EtoolsSnapshotActivityAdmin(ActivityAdmin):
```
domdinicola: maybe we should do this in the library

src/etools/libraries/post_office/admin.py
```diff
@@ -0,0 +1,14 @@
+from django.contrib import admin
+
+from post_office.admin import LogAdmin
+from post_office.models import Log
+
+
+class EtoolsPostOfficeLogAdmin(LogAdmin):
```
domdinicola: we don't need to override this

src/etools/libraries/post_office/tests/test_admin.py
```diff
@@ -0,0 +1,17 @@
...
+from etools.libraries.post_office.admin import EtoolsPostOfficeLogAdmin
+
+from post_office.models import Log
+
+
+class MockRequest:
+    pass
+
+
+class TestPOLogAdmin(BaseTenantTestCase):
```
domdinicola: we don't need to override this

src/etools/applications/t2f/forms.py
csabadenes: I can move it, but it will result in nested transactions in other places where `make_travel_reference_number` is used. I don't think that should be an issue anyway, but please let me know if that's ok.. i have made the suggested changes locally, but i thought it's better to ask before proceed.

src/etools/libraries/snapshot/admin.py
csabadenes: @domdinicola @reinbach  i have moved the snapshot related change to t he library, here is the PR:
https://github.com/unicef/unicef-snapshot/pull/1

src/etools/libraries/post_office/admin.py
csabadenes: Ok, i am removing the override. (without it, the error will persist. i guess it should be fixed by the postoffice library.)

src/etools/libraries/post_office/tests/test_admin.py
csabadenes: Ok, i am removing the override. (without it, the error will persist. i guess it should be fixed by the postoffice library.)

domdinicola:
let's wait for new version of unicef-attachments

domdinicola:
@csabadenes can you update unicef-attachemnt and unicef-snapshot with latest version (@reinbach FYI)

reinbach:
@domdinicola `unicef_attachments` requires review for merging from `develop` to `master`

csabadenes:
@domdinicola @reinbach i will be updating as soon as unicef_attachments is merged.

domdinicola:
@reinbach attachment has been approved

csabadenes:
@robertavram the conflicts should be resolved

ntrncic:
we need searches in email, fund reservations, interventions, agreements, trips, engagements, tpm visits, tpm activities

csabadenes:
@ntrncic i am looking into the issue. As far as i can see more than half of the pages you mentioned already have search filters.

PR #2195, opened 2019-02-21
Fix broken deep attributes in traits after factory-boy upgrade

roman-karpovich:
added metaclass to avoid importing tons of factories for every instance status


domdinicola:
@roman-karpovich thanks Roman, really appreciated!

PR #2196, opened 2019-02-21
Travel Activity is required for T2FActionPoint models

reinbach:
[ch9782]

If `travel_activity` is not set record is not considered a T2FActionPoint, rather just a ActionPoint.
Hence the snapshot fails to find the record, when creating the delta.

### DIFF
--- a/src/etools/applications/audit/admin.py
+++ b/src/etools/applications/audit/admin.py
@@ -1,12 +1,23 @@
-
 from django.contrib import admin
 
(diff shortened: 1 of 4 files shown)


domdinicola:
@reinbach  I guess we need that also for EngagementActionPoint

reinbach:
> @reinbach I guess we need that also for EngagementActionPoint

Yes we do

PR #2212, opened 2019-03-01
Document type shpd to hpd

reinbach:
[ch9741]


src/etools/applications/partners/migrations/0032_auto_20190301_1029.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 2.0.9 on 2019-03-01 10:29
+
+from django.db import migrations, models
+
+
+def update_document_type(apps, schema_editor):
+    Intervention = apps.get_model("partners", "intervention")
+    for intervention in Intervention.objects.filter(document_type="SHPD").all():
+        intervention.document_type = "HPD",
```
domdinicola: why don't use update?
actually we don't what to change the reference_number (as per requirement) maybe we can use mass_update
@reinbach @ntrncic

src/etools/applications/partners/models.py
```diff
@@ -1587,11 +1587,11 @@ class Intervention(TimeStampedModel):
         (TERMINATED, "Terminated"),
     )
     PD = 'PD'
-    SHPD = 'SHPD'
+    SHPD = 'HPD'
```
domdinicola: should we change also the name of the constant?

reinbach: Nope, the reason we have a constant is so we don't have to change the name of it, just its value

src/etools/applications/partners/migrations/0032_auto_20190301_1029.py
reinbach: I still need to get into the habit of using `.update()`, will update this

src/etools/applications/partners/models.py
domdinicola: what about bulk_update https://docs.djangoproject.com/en/2.2/ref/models/querysets/#django.db.models.query.QuerySet.bulk_update?

domdinicola: ops is 2.2

domdinicola:
@ntrncic we will need to run update-notifications and collateral effect is that all draft intervention will "move" to HPD

PR #2216, opened 2019-03-05
Validate overlapping dates in Approved, and Completed status as well

reinbach:
[ch10004]

Ensure overlapping dates are validated/checked in Approved, and Completed status changes as well.

There is a `validate_itinerary` method, which is used for transition to `Submitted` status, I'd recommend moving the validation in the serializer to this, and using this validation for the other transitions `Approved` and `Completed`.

Kept this PR simply as I believe we're wanting to make a quick release.


reinbach:
Start date is now allowed to be equal to end date of previous itineraries

PR #2220, opened 2019-03-05
[ch9800] provide 400 error when Add New Audit Staff Team Member

domdinicola:
[ch9800]


src/etools/applications/audit/serializers/auditor.py
```diff
@@ -40,10 +40,15 @@ def validate(self, attrs):
         validated_data = super().validate(attrs)
         user_pk = validated_data.pop('user_pk', None)
 
-        if not self.instance:
+        if self.instance and user_pk:
```
reinbach: I think we can clean up these `if/else` statements a bit more. The `user_pk` is being used all over.

Maybe the following may be clearer;
```
if user_pk:
  if self.instance:
    ...
  else:
   ...
else:
  if "user" not in validated_data:
   ...
```

domdinicola: sure

domdinicola: @reinbach you comment hsa been addressed

src/etools/applications/audit/serializers/auditor.py
```diff
@@ -40,15 +40,14 @@ def validate(self, attrs):
...
+        if user_pk:
+            if hasattr(user_pk, 'purchase_order_auditorstaffmember'):
+                firm = user_pk.purchase_order_auditorstaffmember.auditor_firm
+                raise serializers.ValidationError({'user': _('User is already assigned to ') + str(firm)})
+            if not self.instance:
                 validated_data['user'] = user_pk
-
-            if 'user' not in validated_data and not user_pk:
-                raise serializers.ValidationError({'user': _('This field is required.')})
+        elif 'user' not in validated_data and not user_pk:
```
reinbach: You probably don't need the `and not user_pk` part as you should not get here based on the initial `if user_pk`

PR #2238, opened 2019-03-13
[ch10122] use __quarter in Partner HACT functions

csabadenes:
[ch10122]


domdinicola:
there's another place where to substitute

csabadenes:
@domdinicola i have replaced the checks for programmatic_visits too. The task description also mentioned planned_visits_to_hact, but i can't find what i could replace there.

PR #2261, opened 2019-03-26
Add results framework endpoint

reinbach:
[ch10174]

new endpoint added;
`api/v2/reports/interventions/results/<intervention_id>/?format=docx_table`


acory:
The file has no extension and the name is 'download'.  Is this ok?

robertavram:
@reinbach let's name this file "{pd_ref_number}_results.docx"

PR #2264, opened 2019-03-27
simply pipfile [list only direct requirements]

domdinicola:


domdinicola:
for now we decided to follow this approach and freeze the direct requirements version in the pipfile.
In this way we are sure we don't update any or major version of direct requirements without notice.

PR #2269, opened 2019-03-28
[ch9810] assign auditor group automatically

domdinicola:
[ch9810]


src/etools/applications/audit/views.py
```diff
@@ -504,6 +505,7 @@ def perform_create(self, serializer, **kwargs):
         if not instance.user.profile.country:
             instance.user.profile.country = self.request.user.profile.country
         instance.user.profile.countries_available.add(self.request.user.profile.country)
+        instance.user.groups.add(Group.objects.get(name='Auditor'))
```
reinbach: Can we move the name `Auditor` to a constant? and make use of a `get_or_create` if some one changes that name in the future, a change will need to be made here

domdinicola: @reinbach does it look better?

PR #2273, opened 2019-03-29
Intervention planned visits updates now reliant on pk for visit

reinbach:
[ch4326]

Previous updates were based on the year of the visit


src/etools/applications/partners/serializers/interventions_v2.py
```diff
@@ -95,7 +95,7 @@ def is_valid(self, raise_exception=True):
         try:
             self.instance = self.Meta.model.objects.get(
                 intervention=self.initial_data.get("intervention"),
-                year=self.initial_data.get("year"),
+                pk=self.initial_data.get("id"),
```
robertavram: @reinbach I'm not sure what this is doing. Currently it works as expected in dev. Could you clarify?

reinbach: @robertavram If you edit a planned visit and change the year, then a new record is created instead of updating it, due to the above query checking against the year of the record instead of the primary key.

PR #2274, opened 2019-03-29
Update attachment file type static results

reinbach:
[ch10183]

Only provide the document types that are available in the `AttachmentFlat` model. These are all the document types that are in use.


robertavram:
Let's jump on a call at some point tomorrow to try to answer a few questions.

src/etools/applications/partners/views/v2.py
```diff
@@ -97,10 +97,7 @@ def get(self, request):
         # hardcoded to partners app attachments for now
         # once tpm and audit come online we can remove the filter
         # on attachment file types
-        attachment_types = AttachmentFileType.objects.values_list(
```
robertavram: let's add a new field "attachmeent_types_active" to represent the attachment types that are in use.. @marko911 can you please use this in FAM frontend for setting the attachment types. (see issue related to this PR)

PR #2307, opened 2019-04-22
[ch11123] filter locations export

domdinicola:
[ch11123]


robertavram:
please do the isort separately from any features though.

PR #2313, opened 2019-04-23
add workspace: add country available for superusers

domdinicola:


src/etools/applications/users/models.py
```diff
@@ -109,11 +109,11 @@ class Country(TenantMixin):
                                        blank=True)
 
     # TODO: rename the related name as it's inappropriate for relating offices to countries.. should be office_countries
-    offices = models.ManyToManyField('Office', related_name='offices', verbose_name=_('Offices'))
+    offices = models.ManyToManyField('Office', related_name='offices', verbose_name=_('Offices'), blank=True)
 
-    threshold_tre_usd = models.DecimalField(max_digits=20, decimal_places=4, default=None, null=True,
+    threshold_tre_usd = models.DecimalField(max_digits=20, decimal_places=4, blank=True, null=True,
```
robertavram: this used to be used for Trips. one day we should get rid of it.

domdinicola: agree, i'll do some clean soon :)

PR #2317, opened 2019-04-24
[ch11233] Management App tweaks

domdinicola:
added purge queue
added sync all users
[ch11233]


src/etools/applications/management/views/general.py
```diff
@@ -23,6 +24,18 @@ def get(self, request, format=None):
         return Response({'success': 'key removed if it was present'})
 
 
+class PurgeQueue(APIView):
```
robertavram: I'm nervous about purging the queue with a click.. We should really never purge the queue!! Important emails could be lost or we'd lose a task that was supposed to run without it failing.

domdinicola: it's only for super users... but I guess maybe is better avoid to have this functionality

PR #2323, opened 2019-04-25
[ch6560] move to python 3.7

domdinicola:


domdinicola:
@reinbach can you test it locally?

reinbach:
> @reinbach can you test it locally?

@domdinicola I went and rebuilt etools in py37 venv and ran the tests, and all looked good locally.

PR #2327, opened 2019-04-26
Update result framework serializer 

reinbach:
[ch10174]

Make use of baseline, target from applied indicators


domdinicola:
@reinbach test are failing

PR #2328, opened 2019-04-29
[ch11606] admin: filter horizontal for many to many fields

domdinicola:
[ch11606]


robertavram:
The naming of the PR is confusing.. I see the location changes.. but i see some other filters sprinkled in everywhere.

src/etools/applications/t2f/admin.py
```diff
@@ -75,6 +79,9 @@ class ItineraryItemAdmin(admin.ModelAdmin):
         'origin',
         'destination'
     )
+    filter_horizontal = (
+        'airlines',
```
robertavram: why do we need to filter by airline?

domdinicola: It is not filtering... (the field is already in the form)
is using filter horizontal for 'manyToMany', looks like is better for the user experience (since is a multiselect, please forget to use CTRL and deselect already selected items)

domdinicola:
is using filter horizontal for 'manyToMany', looks like is better for the user experience (since is a multiselect, please forget to use CTRL and deselect already selected items)

PR #2333, opened 2019-05-01
[ch10965] [ch10704] TPM enhancements

domdinicola:
- changed unique message [ch10965]
- add to TPM group automatically on creation
- changed username generation
- admin panel: TPM staff member use raw_id


domdinicola:
[ch10704]

PR #2336, opened 2019-05-02
Action points export Ref as link

reinbach:
[ch11360]


domdinicola:
why did you need to add another serializer?

reinbach:
To perform the reference number to link conversion

src/etools/applications/action_points/serializers.py
```diff
@@ -276,3 +277,18 @@ def transform_dataset(self, dataset):
...
+    reference_number_link = serializers.SerializerMethodField()
+
+    class Meta(ActionPointListSerializer.Meta):
+        fields = ActionPointListSerializer.Meta.fields + [
+            "reference_number_link",
+        ]
+
+    def get_reference_number_link(self, obj):
+        return "https://{}{}".format(
+            self.context['request'].get_host(),
```
robertavram: can we use the django settings to get the host instead? it would simplify things a bit and not require the request context to be sent.

reinbach: django settings would give us the backend URL, and I believe we want the front end URL

robertavram: The settings.host should be the right host (staging, dev, etools.unicef.org) and the reverse will probably give you the api endpoint no? check ActionPoint.get_object_url method to generate the frontend url.

robertavram: or.. did you test this locally and it works?

reinbach: If different front ends accessed this endpoint then it would work.
I've gone ahead and moved it use `settings.HOST` var

PR #2344, opened 2019-05-09
Gracefully handle redis being down

reinbach:
[ch11092]

If there is a connection error with redis/cache, then continue as normal.

To test locally, comment out the `CACHE` setting in `src/etools/config/settings/local.py` and make sure redis is not up.

The following tests will fail, as they are actually testing caching;

```
src/etools/applications/environment/tests/test_models.py:85:    def test_is_active_function_cached(self):
src/etools/applications/environment/tests/test_models.py:100:    def test_user_set_is_cached(self):
src/etools/applications/environment/tests/test_models.py:109:    def test_empty_user_set_is_cached(self):
src/etools/applications/environment/tests/test_models.py:117:    def test_group_set_is_cached(self):
src/etools/applications/environment/tests/test_models.py:276:    def test_is_active_function_is_cached(self):
src/etools/libraries/locations/tests/test_views.py:73:    def test_api_location_list_cached(self):
```

### DIFF
--- a/src/etools/config/settings/base.py
+++ b/src/etools/config/settings/base.py
@@ -67,7 +67,7 @@ def get_from_secrets_or_env(var_name, default=None):
 # DJANGO: CACHE
 CACHES = {
     'default': {
-        'BACKEND': 'redis_cache.RedisCache',
+        'BACKEND': 'etools.libraries.redis_cache.base.eToolsCache',
         'LOCATION': get_from_secrets_or_env('REDIS_URL', 'redis://localhost:6379/0')
     }
 }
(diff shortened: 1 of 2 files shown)


robertavram:
@reinbach did you manage to look at all the implementation and make sure we don;t fail in other ways if cache returns nothing and exception is not thrown

robertavram:
@saxix can you do a quick review on this as well?

reinbach:
@robertavram yes, did go through all cache calls, and they all appear to be fine and handle `None` being returned by any cache call

PR #2354, opened 2019-05-16
[ch12015] force lowercase when creating new external users

domdinicola:


domdinicola:
[ch12015]

src/etools/applications/firms/serializers.py
```diff
@@ -43,6 +43,9 @@ def create(self, validated_data):
         validated_data.setdefault('username', email)
         return super().create(validated_data)
 
+    def validate_email(self, value):
+        return value.lower()
```
robertavram: I would not use the validate method for lowering this.. maybe a clean method.
However I strongly believe that we need to do this on the user model also. I notice that the domain gets automatically lowered by the User manager, we should probably override to automatically lower everything else.. maybe have a custom email field that does this on the serializer as well.. I think both approaches are needed.. we need a migration script most likely to move all users to lowercase or where conflicts pop up we need to address them by hand. lmk what you think

domdinicola: clean method is for forms not for serializers if remember correct
and yes, I think we should add a constrain on the model as well.
we can write a script to update existing users to lower case and list out the exception we are getting.
maybe we also want to add another script to align username and emails for all the users

robertavram: yeah, clean would manually have to be called.. I really dislike that a validator will change the value of the field. We can create a custom EmailField where we overload 'to_internal_value' to lower before running the rest of the code. maybe? @domdinicola @reinbach ?

reinbach: custom EmailField would probably be best I think

domdinicola: I'll setup a quick meeting to discuss this next week

domdinicola: @robertavram save of existing users may fail

src/etools/applications/t2f/management/commands/et2f_init.py
```diff
@@ -235,6 +237,7 @@ def _load_business_areas(self):
                 self.stdout.write('Business area found: {}'.format(ba[1]))
 
     def _get_or_create_admin_user(self, username, password):
+        print(username, 12234, password)
```
reinbach: Probably want to remove this? :)

domdinicola: yep

PR #2357, opened 2019-05-20
Update trip permissions

reinbach:
[ch11571]

Disable edit permissions for Travel Focal Point user in completed state for end_date and start_date fields


src/etools/applications/t2f/helpers/permission_matrix.py
```diff
@@ -36,16 +37,38 @@ def get_user_role_list(user, travel=None):
...
+            to_format[user][state] = data
+    return to_format
+
+
 def get_permission_matrix():
     permission_matrix = cache.get(PERMISSION_MATRIX_CACHE_KEY)
     if not permission_matrix:
-        path = os.path.join(
-            os.path.dirname(t2f.__file__),
-            "permission_matrix.json",
```
domdinicola: can we get rid of this file then?

reinbach: file has been removed

domdinicola:
@reinbach please fix conflicts,
so we generate the matrix on the fly and we cache it right?

reinbach:
@domdinicola  Yes

PR #2372, opened 2019-05-24
Changes to intervention validations

robertavram:


robertavram:
@domdinicola fix for that weird FR sync error

PR #2387, opened 2019-05-29
[ch9740] migrate pd attachment type

domdinicola:
[ch9740]


src/etools/applications/partners/migrations/0038_auto_20190529_1844.py
```diff
@@ -0,0 +1,26 @@
...
+
+from django.db import migrations
+
+
+def update_targets(apps, schema_editor):
+    InterventionAttachment = apps.get_model("partners", "InterventionAttachment")
+    FileType = apps.get_model("partners", "FileType")
+    partnership_review = FileType.objects.filter(name='Partnership Review')
+    other = FileType.objects.filter(name='Other')
+    if partnership_review.exists() and other.exists():
```
reinbach: What if we have a "Partnership Review" and not "Other"?
Do we want to ensure that "Other" exists? Maybe with the following?
```
if partnership_review.exists():
  other = FileType.objects.get_or_create(name="Other")
```

domdinicola: we are running the update only if both exist and we are going to run this only once.. i think it should be fine

src/etools/applications/partners/migrations/0038_auto_20190529_1844.py
```diff
@@ -0,0 +1,26 @@
...
+        partnership_review = partnership_review.first()
+        other = other.first()
+        InterventionAttachment.objects.filter(type=partnership_review).update(type=other)
+        partnership_review.delete()
+
+
```
robertavram: no need for this one. rename next migration to `38`

PR #2390, opened 2019-05-30
t2f report permission

reinbach:
[ch12396]

Report permissions are limited to just model level


ntrncic:
@marko911 please use this backend branch and confirm it works as you expect on your side..
Let's make 100% sure we're not going to find another regression on permissions. Delaying production push until everyone's sure.
@reinbach and @marko911 please work together, let me know if you need any assistance.
We'll merge as soon as you guys confirm locally works like a charm.
from Rob and I previously commented on the wrong PR.

marko911:
Tested, passes the  flow.

PR #2391, opened 2019-05-30
Use update_or_create and not get_or_create for init attachment filetypes

reinbach:
[ch12439]


domdinicola:
@reinbach is this the reason why the denormalize command did not work for updating attachments in document library?

robertavram:
@marko911 please use this backend branch and confirm it works as you expect on your side..
Let's make 100% sure we're not going to find another regression on permissions. Delaying production push until everyone's sure.
@reinbach and @marko911 please work together, let me know if you need any assistance.
We'll merge as soon as you guys confirm locally works like a charm.

PR #2392, opened 2019-05-31
Remove PD/PCA attachment file type options from fixture

reinbach:
[ch10183]

There is a migration that cleans up these file types (PD/PCA), but In staging it appears the Ghana was added after this migration was run.


src/etools/applications/core/data/attachments_file_types.json
```diff
@@ -169,16 +149,6 @@
             "code": "tpm_report"
         }
     },
-    {
```
domdinicola: do we need to remove also from tpm?

reinbach: I thought it needed to be removed from all? Is that not the case?

domdinicola: @ntrncic do we need `PCA` document type in TPM? Shall we check with the Business Owner?

ntrncic: no we don't need it. @reinbach @domdinicola

PR #2396, opened 2019-06-05
Update value_number validator to attempt Decimal conversion as validation test

reinbach:
[ch12529]


reinbach:
@csabadenes good call

PR #2399, opened 2019-06-09
[ch12697] fixed POSync

domdinicola:


domdinicola:
@ntrncic @robertavram maybe we need an hotfix for this

domdinicola:
[ch12697]

PR #2402, opened 2019-06-10
hotfix 7.0.1

domdinicola:


domdinicola:
the VisionDataTenantSynchronizer is adding the country to the logging and setting the tenant.
this is exactly the old code but splitted between VisionDataTenantSynchronizer (dependent on tenant) and VisionDataSynchronizer (in the library)

PR #2405, opened 2019-06-18
Delegated FRs

robertavram:
Add delegated FRs


src/etools/applications/funds/synchronizers.py
```diff
@@ -206,7 +223,11 @@ def header_sync(self):
         to_create = []
         for item in fr_numbers_from_records:
             record = self.header_records[item]
-            to_create.append(FundsReservationHeader(**record))
+            if self.DELEGATED:
+                fr_instance = FundsReservationHeader(delegated=True, **record)
+            else:
```
domdinicola: if self.DELEGATED:
     record['delegated'] = True
fr_instance = FundsReservationHeader(**record)

iI think is more readable

robertavram:
[ch7941]

PR #2406, opened 2019-06-18
[ch12707] add coverage and use global country for tpm and po sync

domdinicola:


robertavram:
@domdinicola why are we removing the ability to call for all countries?

domdinicola:
@robertavram Both sync are in the public schma.
TPMPartner has a many to many rel but is populated with data of the current schema (check TPMPartner.activate)

src/etools/applications/audit/purchase_order/tests/test_synchronizers.py
```diff
@@ -80,12 +81,7 @@ class TestUpdatePurchaseOrders(BaseTenantTestCase):
     def setUpTestData(cls):
         cls.country = Country.objects.first()
 
-    @mock.patch("etools.applications.vision.tasks.logger.exception")
-    def test_update_purchase_orders_no_country(self, mock_logger_exception):
-        """Ensure no exceptions if no countries"""
-        update_purchase_orders(country_name="Wrong")
-        self.assertEqual(mock_logger_exception.call_count, 0)
-
+    @override_settings(PUBLIC_SCHEMA_NAME='test')
```
reinbach: Maybe make use of the constant `SCHEMA_NAME` in `src/etools/applications/core/tests/cases.py` instead of hardcoding `test`

PR #2412, opened 2019-06-21
Reference numbers moved to db

reinbach:
[ch12761]


src/etools/applications/audit/models.py
```diff
@@ -217,15 +221,11 @@ def unique_id(self):
             self.id
         )
 
-    @property
-    def reference_number(self):
```
domdinicola: :100:

src/etools/applications/action_points/models.py
```diff
@@ -195,7 +205,7 @@ def get_mail_context(self, user=None):
         return {
             'person_responsible': self.assigned_to.get_full_name(),
             'assigned_by': self.assigned_by.get_full_name(),
-            'reference_number': self.reference_number,
+            'reference_number': self.get_reference_number(),
```
domdinicola: is this needed?

reinbach: Yes, as the initial save `created` param is `True` on the signal and it bubbles up to this code, and it may not have the `reference_number` value set

src/etools/applications/action_points/signals.py
```diff
@@ -20,5 +18,6 @@ def action_point_updated_receiver(instance, created, **kwargs):
     if relevant_ap_instance:
         if created:
             instance.send_email(instance.assigned_to, email_template, cc=[instance.assigned_by.email])
-        elif instance.tracker.has_changed('assigned_to'):
-            instance.send_email(instance.assigned_to, email_template)
+        elif not instance.tracker.has_changed('reference_number'):
```
robertavram: what happens if both get changed on the same save?

src/etools/applications/audit/serializers/engagement.py
```diff
@@ -181,12 +181,12 @@ class EngagementLightSerializer(serializers.ModelSerializer):
...
     )
     status_date = serializers.ReadOnlyField(source='displayed_status_date', label=_('Date of Status'))
-    unique_id = serializers.ReadOnlyField(label=_('Unique ID'))
+    reference_number = serializers.ReadOnlyField(label=_('Unique ID'))
 
     class Meta:
         model = Engagement
         fields = [
-            'id', 'unique_id', 'agreement', 'po_item',
+            'id', 'reference_number', 'agreement', 'po_item',
```
robertavram: the frontends might be expecting unique_id

src/etools/applications/t2f/views/travel.py
```diff
@@ -162,8 +162,6 @@ def get_queryset(self):
 
         qs = qs.annotate(status=Case(When(travels__traveler=F('primary_traveler'),
                                           then=F('travels__status')), output_field=CharField()))\
-            .annotate(reference_number=Case(When(travels__traveler=F('primary_traveler'),
```
robertavram: this is worrisome .. no idea why in the world this was done. I'm afraid that the FE might be doing some weird magic based on records with no ref numbers. (mind blown 💥 )

domdinicola:
@reinbach can we split PRs by module?
no need to touch t2f for now

src/etools/applications/action_points/signals.py
reinbach: the idea is that the reference number is set/saved when the record is initially created as a subsequent save after the record is created. The `assigned_to` value should not be changing in that subsequent save.

src/etools/applications/audit/serializers/engagement.py
reinbach: +1 I'll keep this the same then

reinbach:
This PR split into multiple PRs;
- https://github.com/unicef/etools/pull/2472
- https://github.com/unicef/etools/pull/2473
- https://github.com/unicef/etools/pull/2474
- https://github.com/unicef/etools/pull/2475

PR #2414, opened 2019-06-21
CH 12818 - fix unit test warnings.

csabadenes:
- also replace 'datetime.replace' with 'datetime.combine', it should keep the code working correctly if nanosecond support is introduced in the future - https://bugs.python.org/issue15443

[ch12818]


src/etools/applications/partners/utils.py
```diff
@@ -440,9 +440,10 @@ def send_agreement_suspended_notification(agreement, user):
 
 def send_intervention_draft_notification():
     """Send an email to PD/SHPD/SSFA's focal point(s) if in draft status"""
+    sdate_diff = make_aware(datetime.datetime.combine(now(), datetime.time.min) - datetime.timedelta(days=7))
     for intervention in Intervention.objects.filter(
             status=Intervention.DRAFT,
-            created__lt=(datetime.datetime.now(datetime.timezone.utc) - datetime.timedelta(days=7)).date(),
```
domdinicola: can we just remove  .date() ?

csabadenes: I think if we do that, the datetime would not be round down to midnight, and it could lead to inconsistencies.

csabadenes: also using `datetime.timezone.utc` was most likely wrong, it should have been the server's timezone settings.

PR #2423, opened 2019-07-01
[ch12990] sections: intervention with indicators

domdinicola:


domdinicola:
[ch12990]

src/etools/applications/partners/serializers/interventions_v2.py
```diff
@@ -265,6 +266,27 @@ class Meta:
         )
 
 
+class BasicInterventionListSerializer(serializers.ModelSerializer):
+    partner_name = serializers.CharField(source='agreement.partner.name')
+    indicators = serializers.SerializerMethodField()
+
+    def get_indicators(self, obj):
+        qs = [ai for rl in obj.result_links.all() for ll in rl.ll_results.all() for ai in ll.applied_indicators.all()]
```
reinbach: This looks like there may be some performance issues here, seeing that we are doing 3 queries per row.
I think it may be a good idea to attempt to get the indicators in the queryset on the view?

domdinicola: @reinbach I was thinking to use Subquery, but looks like it can only return one value...
Do you have any ideas to make it increase the performaces?

reinbach: @domdinicola I think you may want to look at [Django Database Functions](https://docs.djangoproject.com/en/dev/ref/models/database-functions) and look into extending them.

Samples: https://stackoverflow.com/a/40478702/1608220 or https://stackoverflow.com/a/31337612/1608220

Or maybe make use the postgresql [StringAgg function](https://docs.djangoproject.com/en/dev/ref/contrib/postgres/aggregates/#stringagg)

domdinicola: @reinbach can you have a look to it?

reinbach: @domdinicola actually I think in this scenario we can just to a `prefetch_related` to cut down on the subsequent db queries. Will result in the data being loaded into memory, not sure if that now is a performance/resource issue?

src/etools/applications/partners/views/interventions_v2.py
```diff
@@ -215,6 +216,21 @@ def list(self, request):
...
+    queryset = Intervention.objects.all()
+    serializer_class = BasicInterventionListSerializer
+    permission_classes = (PartnershipManagerPermission,)
+
+    filters = (
+        ('sections', 'sections__in'),
+    )
+
+    def get_queryset(self):
+        return super().get_queryset().select_related('agreement__partner').prefetch_related(
```
domdinicola: I did override that here.. maybe we can remove one of those (here we do also select_related on partners)

reinbach: Sorry, missed that. Yeah I'll undo what I did

PR #2430, opened 2019-07-04
Update django-storage to 1.7.1

reinbach:
[ch10966]

Note that for your local instances you will want to uninstall, then re-install all `azure*` libs, as suggested by library mainterains. Or just rebuild your container/venv.


domdinicola:
can we use the testing branch which generates an image so we can test it in dev environment?

PR #2432, opened 2019-07-04
Add task to send email for overdue action points

reinbach:
[ch12499]


src/etools/applications/action_points/tasks.py
```diff
@@ -0,0 +1,51 @@
...
+
+logger = get_task_logger(__name__)
+
+
+@app.task
+def notify_overdue_action_points():
+    """Send a notification to assignee of an action for each
+    overdue action point (day after due date)
+    """
+    for country in Country.objects.exclude(name='Global').all():
```
domdinicola: would be nice to remove .all() and exclude based on `schema_name` and `get_public_schema_name`

reinbach: How come you want to remove `.all()`

I've updated to make use of the `get_public_schema_name`

domdinicola: @reinbach  do we need .all() ? the exclude should be enough right?

reinbach: Trying to be explicit, what do we have against using `.all()` ? :)

PR #2458, opened 2019-08-02
PSEA

domdinicola:


domdinicola:
@reinbach I don't think is a good idea to call the model Engagement... we already have another class in `audit` app called Engagement, would be better to find another name in order to avoid confusion.

reinbach:
@domdinicola How about `Assessment`?

domdinicola:
@reinbach we have Assessment in `partner`..

reinbach:
@domdinicola I'd say it not really a big deal, the app encapsulates it fine.

src/etools/applications/psea/models.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def rating(self):
+        result = Answer.objects.filter(assessment=self).aggregate(
+            assessment=Sum("rating__weight")
+        )
+        if result:
+            return result["assessment"]
+        return None
+
+    def get_reference_number(self):
```
robertavram: ref number can be "{}/{}PSEA{}".format(country_short_code, current_year, id of record)

src/etools/applications/psea/models.py
```diff
@@ -0,0 +1,316 @@
...
+
+    class Meta:
+        verbose_name = _('Evidence')
+        verbose_name_plural = _('Evidence')
+
+    def __str__(self):
+        return f'{self.label}'
+
+
+class Indicator(TimeStampedModel):
```
robertavram: I think we'll have to come up with some better naming for some of these models.

src/etools/applications/psea/models.py
reinbach: :+1:

src/etools/applications/psea/models.py
reinbach: They seem appropriate/applicable to me

src/etools/applications/psea/serializers.py
```diff
@@ -0,0 +1,490 @@
...
+
+class BaseAssessmentSerializer(serializers.ModelSerializer):
+    permissions = serializers.SerializerMethodField(read_only=True)
+
+    class Meta:
+        model = Assessment
+
+    def get_permissions(self, obj):
+        user = self.context['request'].user
+        if self.context["view"].action == "list":
```
robertavram: I think there is some confusion here.. we shouldn't serialize permissions for the users in List.. not sure why we're doing this.. One thing we need is to filter out all assessments that user does not belong in if the user is not unicef.
I'll make the change

PR #2462, opened 2019-08-05
Doc Type cleanup

reinbach:
[ch10172]

Update `unicef-attachment` to 0.6.0 which adds ability to group attachment file types cleanly. Allows removal up duplicates


robertavram:
@reinbach is this ready? does it address changes with the new moule/s as well? PSEA? etc
also did we release a newer version of attachments in the meantime?

reinbach:
> @reinbach is this ready? does it address changes with the new moule/s as well? PSEA? etc
> also did we release a newer version of attachments in the meantime?

Updates have been made, should be all set now

PR #2463, opened 2019-08-05
[Ch12698] Fix partner import keyerror when 'partner_type_desc' is missing

csabadenes:
[ch12698]

### DIFF
--- a/src/etools/applications/partners/synchronizers.py
+++ b/src/etools/applications/partners/synchronizers.py
@@ -260,8 +260,6 @@ class PartnerSynchronizer(VisionDataTenantSynchronizer):
             'GOVERNMENT': 'Government',
             'UN AGENCY': 'UN Agency',
         }
-        if 'PARTNER_TYPE_DESC' in partner:
-            return type_mapping.get(partner['PARTNER_TYPE_DESC'].upper(), None)
 
     @staticmethod
     def get_partner_rating(partner):
--- a/src/etools/applications/partners/synchronizers.py
+++ b/src/etools/applications/partners/synchronizers.py
@@ -260,7 +260,8 @@ class PartnerSynchronizer(VisionDataTenantSynchronizer):
             'GOVERNMENT': 'Government',
             'UN AGENCY': 'UN Agency',
         }
-        return type_mapping.get(partner['PARTNER_TYPE_DESC'].upper(), None)
+        if 'PARTNER_TYPE_DESC' in partner:
+            return type_mapping.get(partner['PARTNER_TYPE_DESC'].upper(), None)
 
     @staticmethod
     def get_partner_rating(partner):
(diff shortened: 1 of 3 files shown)


domdinicola:
what happen if there's no `partner_type` are we still getting a 500?
maybe we can user `_filter_records` to mark the record as invalid

csabadenes:
@domdinicola , as far as i can see, if there's no partner type, the partner is saved as hidden/deleted/blocked and the partner is logged as 'skipped'.

I have noticed another issue.. the data which caused this exception has no 'partner_type_desc' at all, and it should have been filtered by '_filter_records', because 'PARTNER_TYPE_DESC' is mandatory. So in the end there should be no exception.. i'll change the task to 'WIP' and recheck some things.

src/etools/applications/partners/views/partner_organization_v2.py
```diff
@@ -471,6 +471,12 @@ def create(self, request, *args, **kwargs):
 
         country = request.user.profile.country
         partner_sync = PartnerSynchronizer(business_area_code=country.business_area_code)
+
+        partner_resp = partner_sync._filter_records([partner_resp]).pop()
```
domdinicola: do we need the `pop` here?

src/etools/applications/partners/synchronizers.py
```diff
@@ -260,8 +260,6 @@ def get_partner_type(partner):
             'GOVERNMENT': 'Government',
             'UN AGENCY': 'UN Agency',
         }
-        if 'PARTNER_TYPE_DESC' in partner:
-            return type_mapping.get(partner['PARTNER_TYPE_DESC'].upper(), None)
```
domdinicola: restore the return

src/etools/applications/partners/views/partner_organization_v2.py
domdinicola: i would do `if not partner_sync._filter_records([partner_resp])`

csabadenes: true, the reassignment was unnecessary and also in some cases wrong, i have fixed it

domdinicola: @csabadenes address this

src/etools/applications/partners/synchronizers.py
csabadenes: i noticed that mistake bit later, i have restored it.

csabadenes: i have also added few extra tests to check for missing required fields and invalid partner type

src/etools/applications/partners/views/partner_organization_v2.py
csabadenes: it should be fixed by now

PR #2472, opened 2019-08-14
Action Points reference number moved to db

reinbach:
[ch12761]

See https://github.com/unicef/etools/pull/2412


src/etools/applications/action_points/signals.py
```diff
@@ -20,5 +19,6 @@ def action_point_updated_receiver(instance, created, **kwargs):
     if relevant_ap_instance:
         if created:
             instance.send_email(instance.assigned_to, email_template, cc=[instance.assigned_by.email])
-        elif instance.tracker.has_changed('assigned_to'):
-            instance.send_email(instance.assigned_to, email_template)
+        elif not instance.tracker.has_changed('reference_number'):
```
domdinicola: @reinbach why are we adding this?

src/etools/applications/action_points/models.py
```diff
@@ -195,7 +204,7 @@ def get_mail_context(self, user=None):
         return {
             'person_responsible': self.assigned_to.get_full_name(),
             'assigned_by': self.assigned_by.get_full_name(),
-            'reference_number': self.reference_number,
+            'reference_number': self.get_reference_number(),
```
domdinicola: why we can't we use the reference_number? Is because the email can be send before the creation?

reinbach: Yes, an email is sent out on `created` signal. Reference number only set in subsequent save.

src/etools/applications/action_points/signals.py
reinbach: This is to ignore the subsequent save that happens when setting the reference number, on this save just the reference number is changed. So do not want to accidentally send out `assign_to` email

robertavram: hmm.. Maybe I'm wrong but it looks to me that the purpose of l23 originally was to send another email on a subsequent save only if the "assgined_to" was changed.. so if 20 saves in.. someone changes the pereson responsible for the action point, that person will recieve the email.. I think that logic will break with this proposed change.

src/etools/applications/action_points/models.py
robertavram: on `get_reference_number` method can we return current ref number if exists on record if not the new number generated.. that way we can rest assured that there's no funny business happening.. especially if we change the logic for gen ref numbers in the future.

src/etools/applications/action_points/signals.py
reinbach: If we don't have this logic the `ActionPointsEmailsTestCase.test_create_unlinked` test results in 3 emails being sent out.

src/etools/applications/action_points/models.py
reinbach: done

PR #2475, opened 2019-08-14
TPM reference number moved to db

reinbach:
[ch12761]

See #2412


domdinicola:
@reinbach can you update the migration here?

domdinicola:
@reinbach please resolve conflict and regenerate migration

PR #2477, opened 2019-08-19
TPMActivities API enpoint

domdinicola:


src/etools/applications/tpm/serializers/visit.py
```diff
@@ -103,6 +103,16 @@ def create(self, validated_data):
         return super().create(validated_data)
 
 
+class TPMActivityLightSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = TPMActivity
+        fields = [
+            'id', '__str__', 'date', 'tpm_visit', 'is_pv', 'partner', 'intervention', 'cp_output', 'section',
```
reinbach: Are the consumers of this endpoint expecting `__str__` as a key, maybe rename that?

PR #2479, opened 2019-08-20
[Ch13203] Intervention amendment notifications

csabadenes:
[Ch13203]


src/etools/applications/partners/views/interventions_v2.py
```diff
@@ -298,12 +299,14 @@ def update(self, request, *args, **kwargs):
             nested_related_names=nested_related_names,
             **kwargs
         )
-
         validator = InterventionValid(instance, old=old_instance, user=request.user)
         if not validator.is_valid:
             logging.debug(validator.errors)
             raise ValidationError(validator.errors)
 
+        if old_instance and instance.in_amendment == False and old_instance.in_amendment == True:
```
domdinicola: you can use 
`not instance.in_amendment and instance.in_amendment`

csabadenes: I will change it now, thx

src/etools/applications/partners/views/interventions_v2.py
```diff
@@ -298,12 +299,14 @@ def update(self, request, *args, **kwargs):
             nested_related_names=nested_related_names,
             **kwargs
         )
-
         validator = InterventionValid(instance, old=old_instance, user=request.user)
         if not validator.is_valid:
             logging.debug(validator.errors)
             raise ValidationError(validator.errors)
 
+        if old_instance and not instance.in_amendment and old_instance.in_amendment:
```
robertavram: Let's add a flag, feature-gating this bit.. we should be able to control whether in this country these kinds of emails get sent. Disable them temporarily. and they need to be disabled for now. you can use environment.switch

csabadenes: Ok. I will add the flag asap.

csabadenes: @robertavram i have added a tenantswitch check

PR #2487, opened 2019-09-04
Update unicef-attachment requirement to 0.5.2

reinbach:
[ch14119]

Update attachment permissions, now just requires IsAuthenticated, but can be overwritten if needed


reinbach:
Can someone with the necessary permissions, re-run the circleci test

PR #2493, opened 2019-09-09
Check permissions for assessor out of the matrix

robertavram:


robertavram:
@reinbach this is an example on how to check assessment permission matrix in regards to modifying Assessor. it works well for me.. I think it can probably even be generalized into a mixin at some point.. maybe...

src/etools/applications/psea/permission_matrix/assessment_permissions.csv
```diff
@@ -4,7 +4,7 @@ Field no,Verbose Name,Field Name,Group,Condition,Status,Action,Allowed
 1.2.2,Assessment Date,assessment_date,Audit Focal Point,,assigned,Edit,TRUE
 1.2.3,Assessment Date,assessment_date,Audit Focal Point,,in_progress,Edit,TRUE
 1.2.4,Assessment Date,assessment_date,Audit Focal Point,,submitted,Edit,TRUE
-1.3.1,Assessor,assessors,Audit Focal Point,,draft,Edit,TRUE
+1.3.1,Assessor,assessor,UNICEF Audit Focal Point,,draft,Edit,TRUE
```
reinbach: @robertavram How come you changed the Group to `UNICEF Audit Focal Point` for the Assessor, and left the Group for Assessment as `Audit Focal Point`, is that correct?

robertavram: @reinbach no it's not but I only worried about the one field in this PR.. they should all be changed.. that's the actual name of the django permission group. if it can't find that group for the user (if Audit Focal Point group instead of UNCEF Audit Focal Point) the condition with return False

robertavram: I wanted to more clearly illustrate the changes for the one thing :)

PR #2506, opened 2019-09-26
fix display rating and validation messages

robertavram:
[ch15094]


robertavram:
@reinbach did you notice this?

src/etools/applications/psea/serializers.py
```diff
@@ -141,15 +141,18 @@ def get_available_actions(self, obj):
         is_focal_group = user.groups.filter(
             name__in=[UNICEFAuditFocalPoint.name],
         ).exists()
-        user_belongs = obj.user_belongs(user)
+        user_is_assessor = obj.user_belongs(user)
```
reinbach: Probably want `obj.user_is_assessor(user)`

PR #2509, opened 2019-09-27
PSEA Updates

reinbach:


src/etools/applications/psea/views.py
```diff
@@ -228,7 +230,30 @@ def _set_status(self, request, assessment_status):
...
+            )
+        # assigned status is automatically moved to in progress
+        elif instance.status == Assessment.STATUS_IN_PROGRESS:
+            send_notification_with_template(
+                recipients=instance.assessor.get_recipients(),
+                template_name="psea/assessment/assigned",
+                context={"assessment": instance}
+            )
+        elif instance.status == Assessment.STATUS_REJECTED:
+            send_notification_with_template(
```
robertavram: @reinbach I think we can use "TRANSITION_SIDE_EFFECTS" Property on the Assessment Model.

I don't think I've ever tested that in real life but I think it should work.

see implementation in partners.models L1109

PR #2510, opened 2019-09-27
PSEA Update

reinbach:
Update assessments to make use of TRANSITION_SIDE_EFFECTS


robertavram:
does it work as expected? @reinbach ? :)

reinbach:
@robertavram yes it is, there are tests in place, which continued to pass after changing to this method/logic

robertavram:
super.. I've made some changes to the old version.. once merged in develop maybe we can adjust this one :D ... it seems like nested objects dont ref well.

src/etools/applications/psea/validation.py
```diff
@@ -102,3 +105,35 @@ def assessment_focal_point_user(assessment, user):
...
+        recipients=assessment.get_focal_recipients(),
+        template_name="psea/assessment/submitted",
+        context=assessment.get_mail_context(user)
+    )
+
+
+def assessment_rejected(assessment, old_instance=None, user=None):
+    send_notification_with_template(
+        recipients=assessment.get_recipients(),
+        template_name="psea/assessment/rejected",
```
robertavram: @reinbach I think this template is also looking for "rejection_comment" and will end up blank

src/etools/applications/psea/views.py
```diff
@@ -230,44 +228,7 @@ def _set_status(self, request, assessment_status):
...
-                )
-            # assigned status is automatically moved to in progress
-            elif instance.status == Assessment.STATUS_IN_PROGRESS:
-                send_notification_with_template(
-                    recipients=instance.assessor.get_recipients(),
-                    template_name="psea/assessment/assigned",
-                    context=context
-                )
-            elif instance.status == Assessment.STATUS_REJECTED:
-                context["rejected_comment"] = instance.get_rejected_comment()
```
robertavram: this one :)

src/etools/applications/psea/validation.py
reinbach: Oops :(

PR #2524, opened 2019-10-08
[Ch15316] location remap issues

csabadenes:
[ch15316]
Change location remap flow:
 - separate location remap step from location import step, first remap locations from the remap table, then do the import after the remaping has happened. This solves a current bug happening when a new location is being imported with a pcode that is present in the remap table as an old pcode, and potentially other issues. The import process hopefully is clearer after this change.
 - allow to include every potential remappable pcode in the remap table, and filter them on the fly based on the location usage in the database. Until now, all of the rows from the remap table were considered as active. After this change, only those rows are used where the the remapped location is actually 'in use', this eliminates the need that the remap table has to be constantly adapted to the state of the locations database. Solves the issue of the remap table getting constantly outdated.
 - this remap reorganization also should solve a bug introduced recently, when some locations have been marked inactive after the remap, and could not be re-imported.


src/etools/libraries/locations/task_utils.py
```diff
@@ -364,8 +348,58 @@ def create_location(pcode, carto_table, parent, parent_instance, remapped_old_pc
...
+    """
+    :param carto_table:
+    :param new_pcode: pcode the others will be remapped to
+    :param remapped_pcodes: pcodes to be remapped and archived/removed
+
+    :return: [(new_location.id, remapped_location.id), ...]
+    """
+
+    remapped_locations = Location.objects.all_locations().filter(p_code__in=list(remapped_pcodes))
+    if not remapped_locations:
```
reinbach: May be clearer to have `if not remapped_locations.exists():` as remapped_locations is a queryset

You could rename the variable to `remapped_locations_qs` to be clearer, or if you're happy to load the query results now, change the above line to;

`remapped_locations = Location.objects.all_locations().filter(p_code__in=list(remapped_pcodes)).all()`

and ignore what I said in the first line, and the results will be cached, so when using the result set later in the code, not another hit

src/etools/libraries/locations/tasks.py
```diff
@@ -100,24 +102,58 @@ def update_sites_from_cartodb(carto_table_pk):
...
+                                aggregated_remapped_pcode_pairs[carto_pcode].append(remap_row['old_pcode'])
+
+                    # aggregated_remapped_pcode_pairs - {'new_pcode': ['old_pcode_1', old_pcode_2, ...], ...}
+                    for remapped_new_pcode, remapped_old_pcodes in aggregated_remapped_pcode_pairs.items():
+                        remapped_location_id_pairs = remap_location(
+                            carto_table,
+                            remapped_new_pcode,
+                            remapped_old_pcodes
+                        )
+                        # crete remap history, and remap relevant models to the new location
```
reinbach: Spelling `crete` -> `create`

src/etools/libraries/locations/task_utils.py
csabadenes: I am going with your first suggestion, because i don't see this query reused somewhere else, to profit from chaching. Thanks.

PR #2529, opened 2019-10-09
Require staff members set when assigning if assessor is vendor type

reinbach:
[ch15382]

### DIFF
--- a/src/etools/applications/psea/validation.py
+++ b/src/etools/applications/psea/validation.py
@@ -55,6 +55,15 @@ class AssessmentValid(CompleteValidation):
         self.check_rigid_fields(assessment, related=True)
         return True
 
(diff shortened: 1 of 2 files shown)


src/etools/applications/psea/validation.py
```diff
@@ -55,6 +55,15 @@ def state_draft_valid(self, assessment, user=None):
         self.check_rigid_fields(assessment, related=True)
         return True
 
+    def state_assigned_valid(self, assessment, user=None):
+        # make sure assessor has staff members assigned if vendor type
+        if assessment.assessor.assessor_type == assessment.assessor.TYPE_VENDOR:
+            if not assessment.assessor.auditor_firm_staff.exists():
+                raise StateValidationError(["Staff member(s) required"])
```
robertavram: 👍 so elegant!

PR #2532, opened 2019-10-10
Create Location and Office merger utility

domdinicola:


src/etools/applications/management/handlers/base.py
```diff
@@ -0,0 +1,56 @@
+from django.db import transaction
+from django.db.models.fields.related_descriptors import ForwardManyToOneDescriptor, ManyToManyDescriptor
+
+from model_utils.tracker import DescriptorWrapper
+
+from etools.applications.core.util_scripts import set_country
+
+
+class SimpleHandler:
```
reinbach: Shouldn't this extend an `ABC` class? And maybe rename to `ABCHandler` to be clearer?

domdinicola: done

robertavram:
keep the branch. We'll need to provide the same functionality as sections has. UI and all.

PR #2534, opened 2019-10-10
Add parent_name to locations dropdown in api

reinbach:
[ch15331]


domdinicola:
@reinbach do you think we need to do a select related (to minimize the number of queries)?

reinbach:
@domdinicola it appears I updated the wrong endpoint, and this fix will be done in the locations library, a new PR (https://github.com/unicef/unicef-locations/pull/12) coming in the future

PR #2541, opened 2019-10-17
[ch13376] externals module filter

domdinicola:
[ch13376]


src/etools/libraries/djangolib/views.py
```diff
@@ -0,0 +1,27 @@
+import functools
+import operator
+
+from django.db.models import Q
+
+
+class ExternalModuleFilterMixin:
+
```
robertavram: please add docstrings on usage and behavior

PR #2547, opened 2019-10-23
Tweak OfficeViewSet get_queryset

reinbach:
Attempting to fix issue with all offices being returned by endpoint `/api/offices/` instead of those just within the users selected `country`

Not able to reproduce this exact issue locally, but was not getting any offices.

Locally was seeing strange issue with returned qs from super() not matching
qs in ModelManager


domdinicola:
hey Greg, maybe is related to this https://github.com/unicef/etools/pull/2541
we need the super for ExternalModuleFilterMixin

reinbach:
@domdinicola it does appear related, still need this fix.
Just tested `develop` which has #2541 merged in, and still seeing the issue locally on the `develop` branch

PR #2551, opened 2019-10-29
Arcgis online location import

csabadenes:
[ch9359]

Add Arcgis online as a data source for location import


robertavram:
closing for now, we're looking for a better solution for locations in general

PR #2561, opened 2019-10-31
Amendment signed date validation changes

reinbach:
[ch13116]


src/etools/applications/partners/serializers/agreements_v2.py
```diff
@@ -26,6 +26,30 @@ class Meta:
         model = AgreementAmendment
         fields = "__all__"
 
+    def validate(self, data):
+        data = super().validate(data)
+        if self.context.get('skip_global_validator', None):
```
domdinicola: just curious, what is this `skip_global_validator`

reinbach: `skip_global_validator` is used in `unicef-validator` library. It that is set, then is is expected that there is a validator in place already, so not to call current serializer validator

PR #2569, opened 2019-11-07
Assessment date not required in initial states

reinbach:
[ch16419]


src/etools/applications/psea/tests/test_views.py
```diff
@@ -473,6 +473,29 @@ def test_post(self):
...
+        partner = PartnerFactory()
+        assessment_qs = Assessment.objects.filter(partner=partner)
+        self.assertFalse(assessment_qs.exists())
+
+        response = self.forced_auth_req(
+            "post",
+            reverse('psea:assessment-list'),
+            user=self.focal_user,
+            data={
+                "partner": partner.pk,
```
robertavram: pass an assessment_date key w empty string as value in here. I think that's the problem

robertavram: @reinbach

PR #2571, opened 2019-11-07
Ensure country override set when adding external user to tenant

reinbach:
[ch15959]

### DIFF
--- a/src/etools/applications/users/serializers_v3.py
+++ b/src/etools/applications/users/serializers_v3.py
@@ -178,7 +178,7 @@ class ExternalUserSerializer(MinimalUserSerializer):
         if instance.profile.countries_available.count() == 1:
             if (
                     not instance.profile.country_override and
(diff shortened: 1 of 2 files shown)


robertavram:
hold this please.. I don't think it's correct

src/etools/applications/users/serializers_v3.py
```diff
@@ -175,6 +175,13 @@ def _add_to_country(self, instance):
         exists, country = self._in_country(instance)
         if not exists:
             instance.profile.countries_available.add(country)
+        if instance.profile.countries_available.count() == 1:
+            if (
+                    not instance.profile.country_override and
+                    country.name not in ["UAT", "Global"]
```
domdinicola: can we use schema_name instead of name?

reinbach: Sure thing

PR #2578, opened 2019-11-07
Change emails for psea

robertavram:


src/etools/applications/psea/notifications/assessment-assigned.py
```diff
@@ -5,17 +5,13 @@
...
     Vendor Number: {{ partner_vendor_number }}
 
     Vendor Name: {{ partner_name }}
 
-    PSEA Risk Rating: {{ overall_rating }}
-
-    Date of Assessment: {{ assessment_date }}
-
-    Please update the Vendor Master Data in VISION accordingly
+    Please click <a href="{{ url }}">this link</a> to complete the report.<br/><br/>
```
reinbach: I believe this is a text version email, not sure HTML will work, probably depends on the email client

robertavram: Yes.. I think you're correct.. we'll probably have to try.. if not.. not sure how to hyprelink it

domdinicola: i don't think you can hyperlink it here... usually we just display the link

PR #2583, opened 2019-11-11
Reporting requirements can be added for contingency PDs in signed state

reinbach:
[ch15014]


src/etools/applications/partners/serializers/interventions_v2.py
```diff
@@ -878,6 +878,8 @@ def validate(self, data):
                     raise serializers.ValidationError(
                         _("Changes not allowed when PD is terminated.")
                     )
+            elif self.intervention.contingency_pd and self.intervention.status == Intervention.SIGNED:
```
robertavram: don't forget to make the change in the permission matrix as well (FE relies on that to drive the UI)

reinbach:
> permission matrix

Done

PR #2594, opened 2019-11-13
add libc (gunicorn requirement)

domdinicola:


domdinicola:
https://stackoverflow.com/questions/58786695/how-to-address-oserror-libc-not-found-raised-on-gunicorn-exec-of-flask-app-in

looks like from gunicorn 20.0 is needed

PR #2596, opened 2019-11-18
attempt to fix gunicorn

domdinicola:


domdinicola:
latest version of gunicorn is not compatible with alpine
https://github.com/benoitc/gunicorn/issues/2160

PR #2603, opened 2019-11-21
Spot check sections/offices fields added

reinbach:
[ch15365]


src/etools/applications/audit/serializers/engagement.py
```diff
@@ -380,7 +384,7 @@ class Meta(EngagementSerializer.Meta):
             'internal_controls', 'findings', 'face_form_start_date', 'face_form_end_date',
             'amount_refunded', 'additional_supporting_documentation_provided',
             'justification_provided_and_accepted', 'write_off_required', 'pending_unsupported_amount',
-            'explanation_for_additional_information',
+            'explanation_for_additional_information', 'sections', 'offices',
```
reinbach: I've added the `sections` to the serializer, but this field is not showing up in the serializer data? Does anyone have any idea why this is the case? I'm stumped!

domdinicola:
@reinbach I guess you have to setup the permissions in the `update_audit_permissions` file
then run the command in order to update the permissions

src/etools/applications/audit/serializers/engagement.py
domdinicola: @reinbach I guess you have to setup the permissions in the update_audit_permissions file
then run the command in order to update the permissions

reinbach: @domdinicola thanks, looks like that may be it

robertavram: SeparatedReadWriteField magic!

domdinicola: that's the future @robertavram !

src/etools/applications/audit/views.py
```diff
@@ -444,6 +444,7 @@ class SpotCheckViewSet(EngagementManagementMixin, EngagementViewSet):
     queryset = SpotCheck.objects.all()
     serializer_class = SpotCheckSerializer
     csv_export_serializer = SpotCheckDetailCSVSerializer
+    serializer_action_classes = {}
```
domdinicola: just curious.. why do we need this?

reinbach: In an inherited View the `list` action is mapped to a specific serializer, which does not include the `offices`/`sections` fields. So this was to clear that mapping

domdinicola:
in the card https://app.clubhouse.io/unicefetools/story/15365/spot-check-record-to-capture-respective-section-and-office I see in the comments that it would be tied only to One Office... did this change?

reinbach:
> in the card https://app.clubhouse.io/unicefetools/story/15365/spot-check-record-to-capture-respective-section-and-office I see in the comments that it would be tied only to One Office... did this change?

In a subsequent comment @ntrncic updated that to be many to many for Offices as well

PR #2605, opened 2019-11-22
Fix admin user office

reinbach:
If schema is not selected in admin, user admin interface errors out as it is unable to find `UserTenantProfile`

Add migration for `reports.Office` sequence ids needed to be reset due to old ids being copied across

Correct reports `UserTenantProfile ` data migration


src/etools/applications/reports/migrations/0023_auto_20191014_1546.py
```diff
@@ -1,27 +1,33 @@
...
     for each country that UserProfile has available
     Use the connection to determine the current country
     """
-    UserProfile = apps.get_model("users", "userprofile")
-    UserTenantProfile = apps.get_model("reports", "usertenantprofile")
-    Office = apps.get_model("reports", "office")
-    country = connection.tenant
-    country_name= getattr(country, "name", None)
-    if country_name and country_name != get_public_schema_name():
+    if connection.tenant.schema_name != "test":
```
domdinicola: test or public?

src/etools/applications/reports/migrations/0023_auto_20191014_1546.py
```diff
@@ -1,27 +1,33 @@
...
-            if country in profile.countries_available.all():
-                office = Office.objects.get(pk=profile.office.pk)
-                UserTenantProfile.objects.create(
-                    profile=profile,
-                    office=office,
-                )
+            if profile.office and country in profile.countries_available.all():
+                try:
+                    office = Office.objects.get(pk=profile.office.pk)
+                except Office.DoesNotExist:
```
domdinicola: I think we should not catch this... we have to find these issues

src/etools/applications/reports/migrations/0024_auto_20191122_1559.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 2.2.7 on 2019-11-22 15:59
+
+from django.db import connection, migrations
+from django.db.models import Max
+
+
+def reset_office_seq(apps, schema_editor):
+    Office = apps.get_model("reports", "office")
```
domdinicola: we can add here the `if connection.tenant.schema_name not in ["test", "public"]

src/etools/applications/reports/migrations/0023_auto_20191014_1546.py
domdinicola: `if connection.tenant.schema_name not in ["test", "public"]

src/etools/applications/users/admin.py
```diff
@@ -18,6 +19,15 @@
 from etools.libraries.azure_graph_api.tasks import sync_user
 
 
+def get_office(obj):
```
domdinicola: :100:

src/etools/applications/reports/migrations/0023_auto_20191014_1546.py
reinbach: We don't actually want to catch these as the offices for the user may be in other schemas, and not available in the current schema

src/etools/applications/reports/migrations/0024_auto_20191122_1559.py
reinbach: Done

src/etools/applications/reports/migrations/0023_auto_20191014_1546.py
reinbach: Updated it to both

PR #2623, opened 2019-12-09
[ch15594] [ch15952] FAM/PSEA redirect

domdinicola:
[ch15594] [ch15952]

### DIFF
--- a/src/etools/applications/users/views.py
+++ b/src/etools/applications/users/views.py
@@ -300,6 +300,9 @@ class ModuleRedirectView(RedirectView):
                 return '/tpm/'
 
             if Auditor.as_group() in self.request.user.groups.all():
-                return '/ap/'
+                if False:
+                    return '/psea/'
+                else:
+                    return '/ap/'
 
         return super().get_redirect_url(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


src/etools/applications/users/views.py
```diff
@@ -299,7 +301,17 @@ def get_redirect_url(self, *args, **kwargs):
...
+
+                if Engagement.objects.filter(status__in=[Engagement.PARTNER_CONTACTED, Engagement.REPORT_SUBMITTED],
+                                             agreement__auditor_firm__staff_members=self.request.user):
+                    return '/ap/'
+                elif Assessment.objects.filter(Q(partner__psea_assessment__assessor__user=0) |
+                                               Q(partner__psea_assessment__assessor__auditor_firm_staff__user=0)):
+                    return '/psea/'
+                elif Engagement.objects.filter(agreement__auditor_firm__staff_members=self.request.user):
+                    return '/ap/'
+                else:
```
reinbach: Maybe not default to `/psea/`, already checking for that type above, but raise an assertion error as we have an undefined Engagement?

domdinicola: I think the requirement comes from https://app.clubhouse.io/unicefetools/story/15952/invalid-redirect

reinbach: Ah, in that case carry on

PR #2656, opened 2020-01-25
[Field Monitoring] data collection structure update

roman-karpovich:
The main goal is to re-use structure for offline data collection


roman-karpovich:
added simple usage example: ContactBookExampleTestCase
resulting structure json: https://gist.github.com/roman-karpovich/84f253f54d7a84e582840a5362c92576

robertavram:
This is for step 2... generalizing.. it's most likely going to be useful in the next couple of months.. for now, it would be great to keep current structure.. just nested in one call rather than for each card.

Frontend logic is already built for that.. we could keep it simple for now.

robertavram:
So far.. everything looks wonderful!

robertavram:
@reinbach concurs!

PR #2657, opened 2020-01-27
Field Monitoring: use evident finding values instead of json

roman-karpovich:
masked real values with property having type check


robertavram:
as per call, closing.

PR #2658, opened 2020-01-29
[Field Monitoring] Fix unlinked pd/ssfa validation

roman-karpovich:
ignoring outputs not connected to the current country programme while validating monitored entities relations
clubhouse#17914

### DIFF
--- a/src/etools/applications/field_monitoring/planning/activity_validation/validations/basic.py
+++ b/src/etools/applications/field_monitoring/planning/activity_validation/validations/basic.py
@@ -30,7 +30,8 @@ def interventions_connected_with_partners(i):
 
     diff = set(partners) - set(i.partners.values_list('id', flat=True))
     if diff:
-        error_text = _('Unlinked PD/SSFA detected. Please remove corresponding partners first: {}')
+        error_text = _("You've selected a PD/SSFA and unselected some of it's corresponding partners, "
+                       "please either remove the PD or add the partners back before saving: {}")
         wrong_partners = PartnerOrganization.objects.filter(id__in=diff).values_list('name', flat=True)
         raise BasicValidationError(error_text.format(', '.join(wrong_partners)))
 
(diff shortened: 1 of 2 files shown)


src/etools/applications/field_monitoring/planning/activity_validation/validations/basic.py
```diff
@@ -38,7 +38,10 @@ def interventions_connected_with_partners(i):
 
 
 def interventions_connected_with_cp_outputs(i):
-    cp_outputs = i.interventions.values_list('result_links__cp_output', flat=True)
+    current_cp = CountryProgramme.main_active()
+    cp_outputs = i.interventions.filter(
```
robertavram: @roman-karpovich , i think you related this to this card: https://app.clubhouse.io/unicefetools/story/17914/entities-to-monitor-adding-a-pd-ssfa-with-no-cp-output-results-in-error-message-unlinked-pd-ssfa-detected-please-remove-corresponding#activity-18138

which is a different issue.. 
basically if you relate an intervention that has no `result_links` then the following line
`diff = set(cp_outputs) - set(i.cp_outputs.values_list('id', flat=True))`
will effectively look like: `{None} - {1,2,3 etc}` which results in `{None}` which is a truethy value and the exception is thrown.

robertavram: That being said, I think excluding non-active CP outputs from validation is not a bad idea in principle let's go with it and we'll easily revert later if needed

src/etools/applications/field_monitoring/planning/activity_validation/validations/basic.py
```diff
@@ -38,7 +38,10 @@ def interventions_connected_with_partners(i):
...
+        result_links__cp_output__country_programme=current_cp
+    ).values_list('result_links__cp_output', flat=True)
```
robertavram: Since you're here.. you might be able to correct the error message as well instead of
`Unlinked PD/SSFA detected. Please remove corresponding outputs first: {}'
it can say
`You've selected a PD/SSFA and unselected some of it\'s corresponding outputs, please either remove the PD or add the outputs back before saving: {}'

roman-karpovich: updated texts. i'm not sure about same check for unlinked partners, please check
https://github.com/unicef/etools/pull/2658/commits/7a32facadc05332c549e46acfe2ae1f528ed8806#diff-4ca57aacd65e2fc2459a25c7c4ea55dfR33

PR #2659, opened 2020-02-03
fix hact dumps

robertavram:


domdinicola:
FYI https://github.com/unicef/etools/pull/2652

domdinicola:
"Object of type Decimal is not JSON serializable"

PR #2667, opened 2020-02-19
[FIeld Monitoring] Update data collection methods fixture

roman-karpovich:
[ch18215]
added Measurement & Participatory appraisal to fm methods fixture
fixture should be loaded with `python manage.py tenant_loaddata field_monitoring_methods`

### DIFF
--- a/src/etools/applications/core/data/field_monitoring_methods.json
+++ b/src/etools/applications/core/data/field_monitoring_methods.json
@@ -25,5 +25,23 @@
             "short_name": "Obs.",
             "use_information_source": false
         }
+    },
+    {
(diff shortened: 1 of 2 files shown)


src/etools/applications/core/data/field_monitoring_methods.json
```diff
@@ -25,5 +25,23 @@
             "short_name": "Obs.",
             "use_information_source": false
         }
+    },
+    {
```
robertavram: @roman-karpovich these are not needed.

robertavram:
observation is meant to include measurement and PA. if things change we'll add.. for now keeping out

PR #2670, opened 2020-02-19
[Field Monitoring] updates to checklist blueprint structure

roman-karpovich:


robertavram:
what are the extra types used for?

roman-karpovich:
> what are the extra types used for?

floating_attachments - to indicate attachments input should be rendered not as input, but at the top of the card
abstract - technical groups; frontend goes deeper instead of rendering
card - render as card
collapse - makes card collapseable

robertavram:
> > what are the extra types used for?
> 
> `floating_attachments` - to indicate attachments input should be rendered not as input, but at the top of the card
> `abstract` - technical groups; frontend goes deeper instead of rendering
> `card` - render as card; since we use complex structure which doesn't match exactly to layout, we decided to control rendering evidently
> `collapse` - makes card collapseable



> > what are the extra types used for?
> 
> `floating_attachments` - to indicate attachments input should be rendered not as input, but at the top of the card
> `abstract` - technical groups; frontend goes deeper instead of rendering
> `card` - render as card; since we use complex structure which doesn't match exactly to layout, we decided to control rendering evidently
> `collapse` - makes card collapseable

Hey @roman-karpovich 

We've talked about it and we're wondering if we can't be even more explicit on this..
instead of declaring `extra` and then again `type`, and since it seems that this is always referring to styling, we could easily just place a `styling` key on top level of the object (same level as `title` or `type`) that is the array you need.
example:
instead of: 
```
"structure": {
    "type": "group",
    "extra": {"type":  ["abstract"]},
    "name": "root",
...
```
do this:
```
"structure": {
    "type": "group",
    "styling":  [], // or ["card"], ["card", "collapsable"] etc
    "name": "root",
...
```
Notice that I've left the styling as an empty list (or None) if there is no styling required. ( effectively replacing the abstract word `abstract`)

What do you think. We have consensus on this side if it makes sense please go ahead. if not, happy to continue here or jump on a call.

roman-karpovich:
sure, seems good. since we widely use them, it would be good to transform
styling to common feature

пн, 24 февр. 2020 г., 18:35 Robert Avram <notifications@github.com>:

PR #2677, opened 2020-03-13
[Field Monitoring] Offline collect endpoint improvements & fixes

roman-karpovich:
fixed empty checklist created if something went wrong during form validation;
fixed 500 if findings are missing for provided ids;
fixed user searching: now it's allowed to pass email through body;
added user mapping to checklist attachments during offline form submit;
updated target questions to be not required as we can attach only narrative/attachments without filling in questions


robertavram:
Is this still WIP? Or ready?

roman-karpovich:
i'm unable to check form completiin on our backend because we need frontend
hotfix. so i'll mark as ready a bit later today

пт, 13 мар. 2020 г., 15:29 Robert Avram <notifications@github.com>:

roman-karpovich:
seems working in our test environment, should be much stable now

PR #2682, opened 2020-03-19
Add termination letter requirement for PCA agreements marked terminated

reinbach:
[ch17886]

### DIFF
--- a/src/etools/applications/partners/models.py
+++ b/src/etools/applications/partners/models.py
@@ -1163,6 +1163,13 @@ class Agreement(TimeStampedModel):
         code='partners_agreement',
         blank=True
     )
+    termination_doc = CodedGenericRelation(
+        Attachment,
+        verbose_name=_('Termination document for PCAs'),
+        code='partners_agreement_termination_doc',
+        blank=True,
+        null=True
+    )
     start = models.DateField(
         verbose_name=_("Start Date"),
         null=True,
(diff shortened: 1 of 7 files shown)


src/etools/applications/partners/models.py
```diff
@@ -1319,11 +1326,18 @@ def transition_to_suspended(self):
     def transition_to_cancelled(self):
         pass
 
+    @transition(field=status,
+                source=[SIGNED],
+                target=[TERMINATED, SUSPENDED],
+                conditions=[])
```
robertavram: thing is.. this validation will work even if we don't specify transitional validation, since the field is required on the Treminated status.. but for the sake of clarity, maybe redundant (not sure if needed) but since you added this.. I would tweak it a bit: I would keep only one transition_to_terminated, as the condition for transition.. add the function you created in validation/agreements.py L47 ... slightly tweaked to return False (raise Transition error) by default.. and only if the status is signed and the doc attached should it return True.

src/etools/applications/partners/tests/test_validation_agreements.py
```diff
@@ -96,6 +97,27 @@ def test_valid(self):
...
+    def test_valid(self):
+        terminable_statuses = [Agreement.SIGNED]
+        for terminable_status in terminable_statuses:
+            agreement = AgreementFactory(status=terminable_status)
+            a = AttachmentFactory(
+                code='partners_agreement_termination_doc',
+                content_object=agreement,
+            )
+            agreement.termination_doc.add(a)
+            self.assertTrue(agreements.transition_to_terminated(agreement))
```
robertavram: this will basically test if the function behaves.. it would be good to add or replace this test with a request test.. making a request to the endpoint to transition to terminated.. that way it will be more complete.

reinbach: :+1:

PR #2685, opened 2020-03-23
Pmp prp mandatory integration

reinbach:
[ch17422]

### DIFF
--- a/src/etools/applications/partners/validation/interventions.py
+++ b/src/etools/applications/partners/validation/interventions.py
@@ -169,8 +169,8 @@ def transition_to_active(i):
             _('PD cannot be activated if the Partner is Blocked in Vision')
         ])
 
-    # PD/SSFA require results framework and reporting requirements
-    if i.document_type in [i.PD, i.SSFA]:
+    # PD/SHPD/SSFA require results framework and reporting requirements
+    if i.document_type in [i.PD, i.SSFA, i.SHPD]:
         if not i.reporting_requirements.exists() or not i.result_links.exists():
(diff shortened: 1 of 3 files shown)


src/etools/applications/partners/validation/interventions.py
```diff
@@ -168,6 +168,18 @@ def transition_to_active(i):
         raise TransitionError([
             _('PD cannot be activated if the Partner is Blocked in Vision')
         ])
+
+    # PD/SSFA require results framework and reporting requirements
+    if i.document_type in [i.PD, i.SSFA]:
```
robertavram: everything here.. not just PD, SSFA.. SHPD is actually just a type of PD

reinbach: :+1:

src/etools/applications/partners/validation/interventions.py
```diff
@@ -169,8 +169,8 @@ def transition_to_active(i):
             _('PD cannot be activated if the Partner is Blocked in Vision')
         ])
 
-    # PD/SSFA require results framework and reporting requirements
-    if i.document_type in [i.PD, i.SSFA]:
+    # PD/SHPD/SSFA require results framework and reporting requirements
+    if i.document_type in [i.PD, i.SSFA, i.SHPD]:
```
robertavram: this will be true every time :) can be removed.

reinbach: you say that now, until we add a new type that doesn't belong!! :)

PR #2687, opened 2020-04-02
[Field Monitoring] set location to be required in draft

roman-karpovich:
[ch19310]

### DIFF
--- a/src/etools/applications/field_monitoring/planning/activity_validation/permissions_matrix.csv
+++ b/src/etools/applications/field_monitoring/planning/activity_validation/permissions_matrix.csv
@@ -47,7 +47,7 @@ Field no,Field Name,Group,Condition,Status,Action,Allowed
 ,reject_reason,All Users,is_ma_related_user,*,edit,TRUE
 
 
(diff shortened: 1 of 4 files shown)


src/etools/applications/field_monitoring/planning/activity_validation/permissions_matrix.csv
```diff
@@ -47,7 +47,7 @@ Field no,Field Name,Group,Condition,Status,Action,Allowed
 ,reject_reason,All Users,is_ma_related_user,*,edit,TRUE
 
 
-,location,All Users,,assigned,required,TRUE
+,location,All Users,,draft,required,TRUE
```
robertavram, line 50: you can just add * instead of a status.. that will make it required in all statuses.

As it's written here it won't be required in any other status except of draft.

There are two ways to use the permission matrix in this case.. either 1. write multiple lines and mark requiired true for each status on a different line, or 2. since it'll be required in all statuses you can simply add * instead of the state.

roman-karpovich:
> perm matrix status

done

PR #2707, opened 2020-04-28
Update target_display to handle missing indicator on AppliedIndicator

reinbach:
[ch18196]

`Indicator` field is allowed to be None on `AppliedIndicator`

### DIFF
--- a/src/etools/applications/reports/models.py
+++ b/src/etools/applications/reports/models.py
@@ -659,7 +659,7 @@ class AppliedIndicator(TimeStampedModel):
 
     @cached_property
     def target_display(self):
-        ind_type = self.indicator.display_type
+        ind_type = self.indicator.display_type if self.indicator else None
         numerator = self.target.get('v', self.target)
         denominator = '-'
         if ind_type == IndicatorBlueprint.RATIO:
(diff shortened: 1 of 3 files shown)


src/etools/applications/reports/models.py
```diff
@@ -659,7 +659,7 @@ class AppliedIndicator(TimeStampedModel):
 
     @cached_property
     def target_display(self):
-        ind_type = self.indicator.display_type
+        ind_type = self.indicator.display_type if self.indicator else None
```
robertavram, line 662: I thought indicator is required for any applied indicator

robertavram: I think it should be.

reinbach: The `indicator` field on `AppliedIndicator` model is nullable, and is marked as not required in the serializer.

So during validation on `AppliedIndicatorSerializer` it errors out if there is no `indicator` value.

Do you want me to make `indicator` required on the serializer (https://github.com/unicef/etools/blob/develop/src/etools/applications/reports/serializers/v2.py#L146)

robertavram: i forgot about cluster indicators.. this is the only scenario in which indicator is nullable. Let's get on a call and figure this out.. I think this might be an issue

PR #2722, opened 2020-06-09
Add new fields to prp partner endpoint

reinbach:
[ch17412]

Not sure I have the path to engagements correct, should this go through agreements instead?


src/etools/applications/partners/serializers/prp_v1.py
```diff
@@ -59,9 +61,23 @@ class Meta:
             "phone_number",
             "basis_for_risk_rating",
             "core_values_assessment_date",
-            "type_of_assessment"
+            "type_of_assessment",
+            "sea_risk_rating_name",
+            "psea_assessment_date",
+            "overall_risk_rating",
```
robertavram, line 67: this refers to the `highest_risk_rating` on the partner model.. not through the engagements. add that instead, and also add `highest_risk_rating_type`

reinbach: :+1:

PR #2723, opened 2020-06-12
Add ISO3 Codes to Country Model

reinbach:
[ch18577]

Not able to find a ISO3 code match for the following country records;

Global
UAT
South Sudan
Syria Cross Border
MENARO
FRG
Geneva PFP
Palestine
ESARO
ROSA
WCARO
LACRO
Libya
ECARO
Cabo Verde
Kosovo
Programme Division


src/etools/applications/users/models.py
```diff
@@ -85,7 +85,15 @@ class Country(TenantMixin):
     name = models.CharField(max_length=100, verbose_name=_('Name'))
     country_short_code = models.CharField(
         max_length=10,
-        default='', blank=True, verbose_name=_('Short Code')
+        default='',
+        blank=True,
+        verbose_name=_('UNICEF Country Reference Code')
+    )
+    iso3_code = models.CharField(
```
robertavram, line 92: can you please add this field to the PRP serializer and ensure you can see it in the admin

robertavram:
merging with a todo.

src/etools/applications/users/models.py
reinbach: Field is in the endpoint used by PRP see `CountrySerializer`

Have no added it to admin

PR #2730, opened 2020-06-18
include hact calculation for FM visits

robertavram:


src/etools/applications/field_monitoring/planning/models.py
```diff
@@ -407,6 +410,18 @@ def methods(self):
...
         )
 
+    def update_one_hact_value(self):
+        """
+            Every time an activity transitions to completed, all Partners associated with that activity
+            will increase the completed PV count if applicable
+        """
+
+        aq_qs = self.questions.filter(question__is_hact=True).filter(overall_finding__value__isnull=False)
+        partner_orgs = [aq.partner for aq in aq_qs.all() if aq.partner]
```
reinbach, line 420: Don't want to move the `if aq.partner` into a `filter` on the aq_qs? Then no need for a list expression

src/etools/applications/field_monitoring/analyze/views.py
```diff
@@ -31,24 +30,25 @@ def get(self, request, *args, **kwargs):
...
+
             ).prefetch_related('cp_outputs', 'interventions'),
             to_attr='visits'
         ),
-    ).annotate(
-        completed_visits=Count('monitoring_activities', filter=_completed_activities_filter),
     ).order_by('id').distinct()
 
 
+
```
reinbach, line 50: These extra lines are probably going to fail lint validations, 
This also applies to some of the lines you've removed.

src/etools/applications/field_monitoring/planning/models.py
robertavram: makes sense.. iterative process leftovers

src/etools/applications/field_monitoring/analyze/views.py
robertavram: new computer.. coding in the equivalent of notepad.. I'll address all and clean up next week

PR #2731, opened 2020-06-23
Ensure IsAuthenticated permissions is used

reinbach:
[ch17961]

The default permission class is `IsAuthenticated`

Fixed all permission_classes that attempted to override this default still included `IsAuthenticated` permissions.

The following places were specific about providing "open" access;
- https://github.com/unicef/etools/blob/develop/src/etools/applications/field_monitoring/data_collection/views.py#L71
- https://github.com/unicef/etools/blob/develop/src/etools/applications/core/views.py#L25
- https://github.com/unicef/etools/blob/develop/src/etools/applications/partners/views/prp_v1.py#L23

Are these places ok to be more "open"?

### DIFF
--- a/src/etools/applications/core/views.py
+++ b/src/etools/applications/core/views.py
@@ -5,6 +5,7 @@ from django.shortcuts import redirect
 from django.urls import reverse
 from django.views.generic import RedirectView
 
+from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.views import APIView
 from rest_framework_jwt.serializers import jwt_encode_handler, jwt_payload_handler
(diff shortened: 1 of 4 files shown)


robertavram:
ugh.. before we push to prod
https://github.com/unicef/etools/blob/develop/src/etools/applications/field_monitoring/data_collection/views.py#L71
the first link is def problematic there, we absolutely need auth, read todo 2 lines above. @ntrncic we can't push to prod before we fix this.


https://github.com/unicef/etools/blob/develop/src/etools/applications/core/views.py#L25
second link, here we should use is_authenticated because it's actually using the user to issue the token.. (I think we assumed at the time that authentication is enforced elsewhere (url function?)) 

last link is enforcing user.is_authenticated.. so it's ok..

PR #2736, opened 2020-07-01
Update answer attachment within create function

reinbach:


robertavram:
@reinbach any ticket relating to this? do you remember what the issue was here?

reinbach:
> @reinbach any ticket relating to this? do you remember what the issue was here?

This was me attempting to fix an issue that was happening in prod. I think we can close this PR, as we want with requiring the FE to make another request. @ntrncic thoughts?

robertavram:
@reinbach there's an issue with attachments that is being reported in prod.. but I have no idea what it is.. do you remember the issue that we're trying to fix here?

reinbach:
@robertavram the issue was that the response from the BE did not have the updated data for the attachment. But a refresh did.
This hack was an attempt to make sure the first response did have the updated data.

This issue is not happening locally (or in dev, I believe) just in prod

robertavram:
Fascinating, yeah.. that's not what's being reported. The attachments are
disappearing for some people, apparently, can't reproduce. no idea how this
can happen. I'm getting on a call with a user that's supposedly
experiencing this and I will update after.

robertavram:
nope

PR #2738, opened 2020-07-06
Send relevant email on FM Activity actions

reinbach:
[ch21567]
[ch21571]
[ch21573]
[ch21569]
[ch21572]
[ch21568]

### DIFF
--- a/src/etools/applications/field_monitoring/planning/models.py
+++ b/src/etools/applications/field_monitoring/planning/models.py
@@ -392,7 +392,9 @@ class MonitoringActivity(
             self.init_offline_blueprints()
 
         # send email to users assigned to fm activity
-        recipients = list(self.team_members.all()) + [self.person_responsible]
+        recipients = set(
+            list(self.team_members.all()) + [self.person_responsible]
+        )
         for recipient in recipients:
             self._send_email(
                 recipient.email,
--- a/src/etools/applications/field_monitoring/planning/models.py
+++ b/src/etools/applications/field_monitoring/planning/models.py
@@ -391,18 +391,6 @@ class MonitoringActivity(
             # trigger effects manually? or rewrite this effect?
             self.init_offline_blueprints()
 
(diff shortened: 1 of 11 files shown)


robertavram:
Can you not use side effects?

robertavram:
Can you not use side effects?

Just read your comment. I thought direct transitions also send side effects. As long as the validator is called.

reinbach:
None of the transition methods appear to be called. Still looking into why not, and what I'm missing.

src/etools/applications/field_monitoring/planning/models.py
```diff
@@ -343,9 +387,20 @@ def auto_accept_staff_activity(self):
         if self.monitor_type == self.MONITOR_TYPE_CHOICES.staff:
             self.accept()
             self.save()
-            # todo: direct transitions doesn't trigger side effects. trigger effects manually? or rewrite this effect?
+            # todo: direct transitions doesn't trigger side effects.
+            # trigger effects manually? or rewrite this effect?
             self.init_offline_blueprints()
 
+        # send email to users assigned to fm activity
+        recipients = list(self.team_members.all()) + [self.person_responsible]
```
roman-karpovich, line 395: maybe use set here to avoid possible email duplication

reinbach: :+1:

robertavram:
The way we implemented it, it's not designed to call the functions decorated functions. So nothing in the functions gets executed unless you call the function itself, if you do that, you'll transition but effectively circumvent the validator. I would not place any content in the transition functions and use side effects. Although, not completely sure how it was implemented in fm

reinbach:
I keep forgetting that the validator is handling it :(

roman-karpovich:
> None of the transition methods appear to be called. Still looking into why not, and what I'm missing.

they are defined here just for test purposes. in fact side effects are to be used for treal ransitions

PR #2746, opened 2020-07-14
[ePD] [ch21867] Generic comments app

roman-karpovich:
[ch21867]


roman-karpovich:
@reinbach i'm not sure if we really need FSM to control states for comments instead of simple charfield. please take a look

src/etools/applications/comments/models.py
```diff
@@ -0,0 +1,35 @@
...
+        ('deleted', _('Deleted')),
+    )
+
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, verbose_name=_('User'),
+                             on_delete=models.PROTECT, related_name='comments')
+    state = models.CharField(verbose_name=_('State'), max_length=10, choices=STATES, default=STATES.active)
+    instance_related_ct = models.ForeignKey(ContentType, verbose_name=_('Content Type'), on_delete=models.CASCADE)
+    instance_related_id = models.IntegerField(verbose_name=_('Object ID'))
+    instance_related = GenericForeignKey(ct_field='instance_related_ct', fk_field='instance_related_id')
+    users_related = models.ManyToManyField(settings.AUTH_USER_MODEL, blank=True, related_name='mentions')
```
reinbach, line 24: I don't see a scenario where a comment is related to more than one user. The user who creates the comment is "owner" of the comment, and they are the ones that are able to edit/delete that comment.

src/etools/applications/comments/models.py
```diff
@@ -0,0 +1,35 @@
...
+    related_to_description = models.TextField(blank=True)
+    related_to = models.CharField(max_length=100)
```
reinbach, line 26: The `related_to` field is the main key that is used to relate the comment to a specific object. There is the chance/potential that a comment is not related to a model, or the model object does not have a clear way to create a generic relation to the model. (e.g. comment app is moved into a micro service).

This field is required to both create and to get a list of comments. All requests from the Frontend will be referenced by this `related_to` field.

src/etools/applications/comments/models.py
```diff
@@ -0,0 +1,35 @@
...
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from model_utils import Choices
+from model_utils.models import TimeStampedModel
+
+
+class Comment(TimeStampedModel, models.Model):
```
reinbach, line 11: A user can "reply" to a comment, so we will want to have a parent field that relates to Comments model.

We will need to determine the permissions/logic around a user wanting to delete their comment, when there are children/replies to the comment.

src/etools/applications/comments/models.py
roman-karpovich: as i can see from draft data models document, this field is going to be used for users tagging in comment

src/etools/applications/comments/models.py
roman-karpovich: sounds good

src/etools/applications/comments/urls.py
```diff
@@ -0,0 +1,16 @@
+from django.conf.urls import include, url
```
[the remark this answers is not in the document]
roman-karpovich: updated

src/etools/config/urls.py
```diff
@@ -75,6 +75,7 @@
     url(r'^locations/cartodbtables/$', CartoDBTablesView.as_view(), name='cartodbtables'),
     url(r'^locations/autocomplete/$', LocationQuerySetView.as_view(), name='locations_autocomplete'),
     url(r'^api/v1/field-monitoring/', include('etools.applications.field_monitoring.urls')),
+    url(r'^api/v1/comments/', include('etools.applications.comments.urls')),
```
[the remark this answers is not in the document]
roman-karpovich: updated

src/etools/applications/comments/models.py
roman-karpovich: added parent field with corresponding validations

src/etools/applications/comments/models.py
reinbach: Ah, I missed that completely. Ignore me then :)

PR #2750, opened 2020-07-22
ePD users

reinbach:
[ch21985]


src/etools/applications/users/views_v3.py
```diff
@@ -56,42 +58,51 @@ def retrieve(self, request, pk=None):
...
-        queryset = super().get_queryset().filter(
-            profile__country=user.profile.country, is_staff=True).prefetch_related(
-            'profile', 'groups', 'user_permissions').order_by('first_name')
-
-        return queryset
+                qs = qs.filter(id__in=user_ids)
+
+        if self.is_partner_staff():
+            emails = []
+            for p in self.partners():
```
roman-karpovich, line 90: maybe we should prefetch staff members. in worst case there will be lot of db queries here

src/etools/applications/users/tests/test_views_v3.py
```diff
@@ -162,6 +172,22 @@ def test_minimal_verbosity(self):
...
+        partner_user = UserFactory(email=partner_staff.email)
+        partner_user.profile.partner_staff_member = True
+        partner_user.profile.save()
+        self.assertTrue(get_user_model().objects.count() > 1)
+        response = self.forced_auth_req(
+            'get',
+            self.url,
+            user=partner_user
+        )
+        response_json = json.loads(response.rendered_content)
```
roman-karpovich, line 187: just my curiosity: do we really need to parse rendered content? maybe it will be easier to use response.data

reinbach: Nope you're right, `response.data` works fine here

src/etools/applications/users/views_v3.py
reinbach: The idea is that if a user is hopefully only connected one partner, and so this should only be a single extra query for the staff members of that partner. Unless I'm missing something.

PR #2755, opened 2020-08-03
[ePD] Intervention pd activity details and timings management

roman-karpovich:
[ch22038]
[ch21924]


src/etools/applications/partners/utils.py
```diff
@@ -627,3 +629,27 @@ def sync_partner_staff_member(partner: PartnerOrganization, staff_member_data: P
     profile.country = profile.country or connection.tenant
     profile.save()
     profile.countries_available.add(connection.tenant)
+
+
+def get_quarters_range(start: datetime.date, end: datetime.date) -> List[Tuple[datetime.date, datetime.date]]:
+    """[)"""
```
reinbach, line 635: Not sure what is up with the doc string?

roman-karpovich: i tried to mention here that we include first date and exclude last from range. i'll update this doc

PR #2756, opened 2020-08-03
ePD management budget

reinbach:
[ch22062]


src/etools/applications/partners/views/interventions_v3.py
```diff
@@ -120,7 +129,28 @@ class InterventionPDOutputsListCreateView(InterventionPDOutputsViewMixin, ListCr
 
 
 class InterventionPDOutputsDetailUpdateView(InterventionPDOutputsViewMixin, RetrieveUpdateDestroyAPIView):
-    pass
+    def perform_update(self, serializer):
```
roman-karpovich, line 132: we don't need intervention here now. now serializers remember intervention on initialization and validate it on save

src/etools/applications/partners/serializers/interventions_v3.py
```diff
@@ -254,3 +254,9 @@ class Meta:
             "unicef_focal_points",
             "unicef_signatory",
         )
+
+
+class InterventionManagementBudgetSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = InterventionManagementBudget
+        fields = "__all__"
```
roman-karpovich, line 262: not sure it's good to use __all__ for new api's
from [drf guidelines](https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include): "It is strongly recommended that you explicitly set all fields that should be serialized using the fields attribute. This will make it less likely to result in unintentionally exposing data when your models change."

reinbach: :+1:

src/etools/applications/partners/views/interventions_v3.py
reinbach: :+1:

PR #2758, opened 2020-08-04
[ePD] Add endpoint to get dropdowns data

roman-karpovich:
[ch21993]


src/etools/applications/partners/views/v3.py
```diff
@@ -44,3 +56,69 @@ def map_serializer(self, serializer):
...
+class PMPDropdownsListApiView(APIView):
+    permission_classes = (IsAuthenticated, IsAdminUser | IsPartnerUser)
+
+    @cached_property
+    def partner(self):
+        return PartnerOrganization.objects.filter(staff_members__email=self.request.user.email).first()
+
+    def get_signed_by_unicef_users(self, request) -> list:
+        return list(get_user_model().objects.filter(
+            groups__name__in=['Senior Management Team'],
```
reinbach, line 70: It may be prudent to move this group name to a constant/setting value. I think it is used else where in the code as well

roman-karpovich: yep. there a lot of places where groups are used as strings. maybe it will make sense to do changes separately?

reinbach: Yes separately is good, slowly clean up the rest of the places

PR #2761, opened 2020-08-06
ePD supply item

reinbach:
[ch22063]


src/etools/applications/partners/models.py
```diff
@@ -2497,6 +2512,61 @@ def __str__(self):
...
+                    self.partner_contribution_local += activity.cso_cash
+                    self.unicef_cash_local += activity.unicef_cash
+        try:
+            programme_effectiveness = 0
+            partner_contribution_local = self.partner_contribution_local
+            unicef_cash_local = self.unicef_cash_local
+            if not init:
+                init_totals()
+                init = True
+            for i in range(1, 4):
```
robertavram, line 2538: I would move this logic in a function on the ManagementBudget model.. call the totals from there since it'll always be a One2One model. That way it will be easier to retrieve those total values as well.

src/etools/applications/partners/models.py
```diff
@@ -2497,6 +2512,61 @@ def __str__(self):
...
+                self.unicef_cash_local += unicef_value
+                programme_effectiveness += unicef_value
+        except InterventionManagementBudget.DoesNotExist:
+            self.partner_contribution_local = partner_contribution_local
+            self.unicef_cash_local = unicef_cash_local
+
+        # in kind totals
+        if self.intervention.supply_items.exists():
+            self.in_kind_amount_local = 0
+            for item in self.intervention.supply_items.all():
```
robertavram, line 2560: I like that you're iterating and adding, in case prefetch is used an aggregation query is avoided.

src/etools/applications/partners/models.py
```diff
@@ -2735,6 +2805,13 @@ class InterventionManagementBudget(TimeStampedModel):
         null=True,
     )
 
+    def save(self, *args, **kwargs):
```
robertavram, line 2808: as mentioned above, It could be cleaner to add 3 methods on this model. `get_unicef_total`, `get_partner_total`, `get_total` and use these methods instead from the intervention budget.
Although currently I can't really think whether we need to store that percentage calculation or just do it on serialize-out, saving it would not be a bad idea to make it easier for datamart exports. Up to you on the approach .

src/etools/applications/partners/models.py
reinbach: Where are you wanting to move this method to?

src/etools/applications/partners/models.py
```diff
@@ -2735,6 +2805,13 @@ class InterventionManagementBudget(TimeStampedModel):
...
+        super().save(*args, **kwargs)
+        try:
+            self.intervention.planned_budget.calc_totals()
+        except InterventionBudget.DoesNotExist:
```
roman-karpovich, line 2812: shouldn't we automatically create intervention budget (on intervention create) instead of catching?

reinbach: we could, will need to put a migration in place to create budget records for currently existing interventions

reinbach: Actually this planned_budget relationship is allowed to be null/blank in the model. So not sure if it is best to go ahead and automatically create the relationship. Unless the null attribute is no longer needed.

PR #2766, opened 2020-08-17
[ePD] Transition active PD to new CP

roman-karpovich:
[ch21936]


robertavram:
for all active PDs, when transitioning to a new CP cycle, make sure to add the new CP Structure to the Intervention (do not remove the old one)

src/etools/applications/core/management/commands/init-celery.py
```diff
@@ -73,4 +73,9 @@ def handle(self, *args, **options):
             'enabled': False,
             'crontab': first_day_of_the_month})
 
+        PeriodicTask.objects.get_or_create(name='Transition active PD to new CP', defaults={
```
robertavram, line 76: fixtures

src/etools/applications/reports/tasks.py
```diff
@@ -0,0 +1,43 @@
+from django.db import connection
+from django.utils import timezone
+
+from etools.applications.partners.models import Intervention
+from etools.applications.users.models import Country
+from etools.config.celery import app
+
+
+@app.task
+def transfer_active_pds_to_new_cp():
```
robertavram, line 10: move to partners

robertavram:
needs a few updates

src/etools/applications/reports/tasks.py
roman-karpovich: updated

PR #2769, opened 2020-08-20
[ePD] PD Management endpoints

roman-karpovich:
[ch21928]


roman-karpovich:
@reinbach @robertavram i updated intervention serializer to deal with `Final Partnership Review` section in `Management` tab as we discussed (it's attachment of specific type). at this moment final_partnership_review is separate field which is manageable from permissions matrix. not sure how it looks, i'm in doubt about the way to save them (see todo comment in the intervention serializer update method). please take a look.

PR #2778, opened 2020-09-08
[ePD] Fix backend reported issues

roman-karpovich:
set default 0 for intervention management budget fields
added id to supply item serializer
added missing intervention budget fields to serializer


roman-karpovich:
[ch22257]

PR #2787, opened 2020-09-18
[ePD] Fix duplicate integrity error on output association

roman-karpovich:
[ch22471]


src/etools/applications/reports/models.py
```diff
@@ -375,7 +378,7 @@ class Meta:
         ordering = ('created',)
 
     def save(self, **kwargs):
-        if not self.code:
+        if not self.code or self.tracker.has_changed('result_link_id'):
```
robertavram, line 381: make code nullable and remove the code from save method.. no need for it anymore

roman-karpovich:
merged #2791 into for easier migrations resolving

PR #2793, opened 2020-09-21
Add cfei_number to intervention list results

reinbach:
[ch22493]

Allow searching on `cfei_number` in intervention list api endpoint


reinbach:
This needs to be on the new PMP

PR #2814, opened 2020-09-29
[ePD] Update agreement permission

reinbach:


roman-karpovich (approved):

src/etools/applications/partners/views/interventions_v3.py
```diff
@@ -53,46 +53,11 @@
     InterventionIndicatorsUpdateView,
     InterventionListAPIView,
 )
-from etools.applications.partners.views.v3 import PMPBaseViewMixin
+from etools.applications.partners.views.v3 import APIActionsMixin, PMPBaseViewMixin
 from etools.applications.reports.models import InterventionActivity, LowerResult
 from etools.applications.reports.serializers.v2 import InterventionActivityDetailSerializer
 
 
-class APIActionsMixin:
```
reinbach, line 61: @roman-karpovich I'm thinking of removing this mixin, as I see that it's an abstraction that is really just converting the methods (GET, POST, etc) to another label and then attempting to separate list/retrieve, but not sure we would need to differentiate those 2 in permissions?

Is there something else I'm missing that this mixin is doing?

Would need to update the relevant `view_action_permission`, etc if we removed this mixin.

roman-karpovich: yes, mixin can be removed without big problems. the main reason is readability, actions are just easier to operate with

src/etools/applications/partners/permissions.py
```diff
@@ -461,7 +461,26 @@ def has_permission(self, request, view):
...
+# allow partners to load list ONLY for agreements to keep everything
+# else working right as before; at least for now
+PMPAgreementPermission = (
+    (view_action_permission('create') & (
+        UserIsStaffPermission | user_group_permission('Partnership Manager')
+    )) |
+    (view_action_permission(
+        'list',
+        'retrieve',
+        'update',
```
robertavram, line 479: update and partial update should move on line 473

robertavram (changes requested):

src/etools/applications/partners/permissions.py
reinbach: Sorry forgot to do that

PR #2821, opened 2020-09-30
[ePD] Handle cancelled to status list in serializer

reinbach:
[ch22862]

### DIFF
--- a/src/etools/applications/partners/serializers/interventions_v3.py
+++ b/src/etools/applications/partners/serializers/interventions_v3.py
@@ -278,9 +278,6 @@ class InterventionDetailSerializer(serializers.ModelSerializer):
             ]
         elif obj.status == obj.CANCELLED:
             status_list = [
-                obj.DRAFT,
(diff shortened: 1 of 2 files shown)


src/etools/applications/partners/serializers/interventions_v3.py
```diff
@@ -276,6 +276,13 @@ def get_status_list(self, obj):
                 obj.SIGNED,
                 obj.TERMINATED,
             ]
+        elif obj.status == obj.CANCELLED:
+            status_list = [
+                obj.DRAFT,
```
robertavram, line 281: Cancelled only.

If the status is cancelled no need to show potential previous states as it could have been cancelled in Draft.. we can just simply show: "Cancelled" if we want to see anything else we'd need to go through the history and check all the previous statuses. Don't think it's necessary for the MVP

reinbach: :+1:

PR #2837, opened 2020-10-08
[ePD] Set default currency value for InterventionBudget

reinbach:
[ch23005]

### DIFF
--- a/src/etools/applications/partners/models.py
+++ b/src/etools/applications/partners/models.py
@@ -2591,6 +2591,14 @@ class InterventionBudget(TimeStampedModel):
         Calculate total budget on save
         """
         self.calc_totals(save=False)
+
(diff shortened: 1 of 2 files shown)


src/etools/applications/partners/models.py
```diff
@@ -2591,6 +2591,14 @@ def save(self, **kwargs):
         Calculate total budget on save
         """
         self.calc_totals(save=False)
+
+        # attempt to set default currency
+        if not self.currency:
+            try:
+                self.currency = connection.tenant.local_currency.code
+            except AttributeError:
+                self.currency = ""
```
robertavram, line 2600: use USD if it can;t be found

reinbach: :+1:

PR #2838, opened 2020-10-08
[ePD] Add post_delete signals to ensure totals are recalculated

reinbach:
[ch22907]


src/etools/applications/partners/tests/test_models.py
```diff
@@ -1587,6 +1587,26 @@ def test_totals(self):
...
+        for __ in range(3):
+            item = InterventionSupplyItemFactory(
+                intervention=intervention,
+                unit_number=1,
+                unit_price=2,
+            )
+
+        self.assertEqual(budget.in_kind_amount_local, 6)
+
+        item.delete()
```
robertavram, line 1604: do the signals get called synchronously?

reinbach: my understanding is that they are synchronous, as you can have `pre_*` signals, and those would need to happen before the actual event

PR #2850, opened 2020-10-15
[ePD] Add PD list sent_to_partner filtering option

reinbach:
[ch23137]

Requires update in `unicef_restlib` library, see PR https://github.com/unicef/unicef-restlib/pull/11


src/etools/applications/partners/views/interventions_v3.py
```diff
@@ -118,6 +118,9 @@ class PMPInterventionListCreateView(PMPInterventionMixin, InterventionListAPIVie
         'number__icontains',
         'cfei_number__icontains',
     )
+    filters = InterventionListAPIView.filters + [
```
robertavram, line 121: not sure if this is supposed to check if it was ever sent to partner or if it's in the "partner's court". merging and we'll adjust if needed

PR #2851, opened 2020-10-15
Permissions and actions updates

robertavram:


robertavram:
@roman-karpovich if you could help with tests it would be awesome.
I'm relatively confident for now the permissions until the review state are working ok

src/etools/applications/partners/permissions.py
```diff
@@ -118,6 +121,9 @@ def prp_mode_off():
         def prp_server_on():
             return tenant_switch_is_active("prp_server_on")
 
+        def unlocked(instance):
+            return not(instance.partner_accepted or instance.unicef_accepted)
```
reinbach, line 125: Maybe move this logic to a `locked` method on the `Intervention` model, looks like we use this logic else where

src/etools/applications/partners/views/interventions_v3_actions.py
```diff
@@ -248,19 +273,17 @@ class PMPInterventionUnlockView(PMPInterventionActionView):
     def update(self, request, *args, **kwargs):
         pd = self.get_object()
         request.data.clear()
+        if not(pd.partner_accepted or pd.unicef_accepted):
```
reinbach, line 276: See note above about adding `locked` method on `Intervention` model

reinbach (approved):
Besides minor comment, and not going through the `intervention_permissions.csv` file too closely (makes me cross-eyed) looks good to me

roman-karpovich (approved):

PR #2868, opened 2020-10-27
[ePD] fix duplicated offices for partner

roman-karpovich:
[ch23440]


reinbach (approved):

roman-karpovich:
distinct already exists in ExternalModuleFilterMixin from latest changes. just tests left to avoid functionality degradation

PR #2873, opened 2020-10-29
[ePD] Add v3 endpoint for special reporting requirements

roman-karpovich:
[ch23547]


src/etools/applications/reports/views/v3.py
```diff
@@ -61,6 +69,33 @@ def get_queryset(self, format=None):
...
         SpecialReportingRequirementListCreateView,
 ):
-    """Wrapper for Special Reporting Requirement View"""
+    permission_classes = [
+        IsAuthenticated,
+        IsReadAction | (IsEditAction & intervention_field_is_editable_permission('reporting_requirements')),
+    ]
+
+    def get_root_object(self):
+        return Intervention.objects.filter(pk=self.kwargs.get('intervention_pk')).first()
```
reinbach, line 81: Any reason not to use `get_object_or_404` what is the effect if an invalid `intervention_pk` is provided?

roman-karpovich:
i'm not sure, if i remember correctly, there was issue with swagger in case
of 404 in similar views. i'll re-check

чт, 29 окт. 2020 г., 16:02 Greg Reinbach <notifications@github.com>:

src/etools/applications/reports/views/v3.py
roman-karpovich: i thought swagger will be broken with get_object_or_404, but unexpectedly it's still working. i'm a bit confused :)

PR #2891, opened 2020-11-12
[ePD] Fix epd tests

roman-karpovich:

PR #2894, opened 2020-11-13
Ensure currency is set correctly when creating PD

reinbach:
[ch23755]


roman-karpovich (approved):
#2891 can be used as base. in this case merge migration won't be needed

PR #2918, opened 2020-12-10
[ePD] Residual 'Unassociated to CP Output' after Partner user adds and then deletes a PD output

roman-karpovich:
[ch24144]

### DIFF
--- a/src/etools/applications/partners/views/interventions_v3.py
+++ b/src/etools/applications/partners/views/interventions_v3.py
@@ -228,7 +228,12 @@ class InterventionPDOutputsListCreateView(InterventionPDOutputsViewMixin, ListCr
 
 
 class InterventionPDOutputsDetailUpdateView(InterventionPDOutputsViewMixin, RetrieveUpdateDestroyAPIView):
-    pass
+    def perform_destroy(self, instance):
+        # do cleanup if pd output is still not associated to cp output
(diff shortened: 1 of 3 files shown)


reinbach (approved):

src/etools/applications/partners/views/interventions_v3.py
```diff
@@ -228,7 +228,12 @@ class InterventionPDOutputsListCreateView(InterventionPDOutputsViewMixin, ListCr
 
 
 class InterventionPDOutputsDetailUpdateView(InterventionPDOutputsViewMixin, RetrieveUpdateDestroyAPIView):
-    pass
+    def perform_destroy(self, instance):
+        # do cleanup if pd output is still not associated to cp output
+        result_link = instance.result_link
+        instance.delete()
+        if result_link.cp_output is None:
+            result_link.delete()
```
robertavram, line 236: are we making sure there are no other Lower Results under this before deleting it somewhere? @roman-karpovich

roman-karpovich:
we can check, but in fact there will be always one lower result. we create
temp result link for every new pd output, so there is no case we will
associate multiple pd outputs to incomplete result link.

чт, 10 дек. 2020 г., 20:24 Robert Avram <notifications@github.com>:

PR #2920, opened 2020-12-14
[ePD] Set submission date automatically

roman-karpovich:
[ch24260]


reinbach (approved):

robertavram:
The bug that originally led to the creation of this requirement is no longer reproducible . Users should still be able to override. Closing for now.

PR #2926, opened 2020-12-24
[FM] Geographic Coverage Map: colors are not updated according to colors legend for visited sites.

roman-karpovich:
[ch23940]


src/etools/applications/field_monitoring/analyze/views.py
```diff
@@ -88,15 +88,22 @@ class CoverageGeographicView(ListAPIView):
     def get_queryset(self):
         queryset = super().get_queryset()
 
-        activities_filter = _completed_activities_filter
+        activities_filter = Q(
+            status=MonitoringActivity.STATUSES.completed,
+            location__lft__gte=OuterRef('lft'),
```
robertavram, line 93: it would be great to add a comment here that explains the usage of OuterRef filters such as cumulative addition. Can be done in a separate PR.

PR #2934, opened 2021-01-08
[ePD] Rename PD/SSFA to PD/SPD where applicable

reinbach:
[ch24485]


domdinicola:
@reinbach  does this come with datamart changes?

reinbach:
> @reinbach does this come with datamart changes?

Yes we probably want to update datamart with similar changes

domdinicola (approved):

PR #2946, opened 2021-02-23
sentry staging 1256

domdinicola:
https://excubo.unicef.io/sentry/etools-staging/issues/1256/?query=is%3Aunresolved


src/etools/libraries/azure_graph_api/utils.py
```diff
@@ -29,7 +29,7 @@ def handle_record(record):
 
     record_dict = {
         'Username*': record.get('userPrincipalName', '-'),
-        'Email*': record.get('mail', '-').lower(),
+        'Email*': record.get('mail', '-'),
```
robertavram, line 32: keep lower if record exists

PR #2947, opened 2021-02-23
unique business area code

domdinicola:


robertavram:
let's look at where the get is happening, we should keep business area as not unique.

PR #2951, opened 2021-02-24
[ch24679] PartnerPlannedVisits for 2022

domdinicola:
[ch24679]

### DIFF
--- a/src/etools/applications/field_monitoring/analyze/serializers.py
+++ b/src/etools/applications/field_monitoring/analyze/serializers.py
@@ -9,7 +9,6 @@ from etools.applications.partners.models import Intervention, PartnerOrganizatio
 from etools.applications.partners.serializers.interventions_v2 import MinimalInterventionListSerializer
 from etools.applications.reports.models import Result
 from etools.applications.reports.serializers.v2 import MinimalOutputListSerializer
-from etools.libraries.pythonlib.datetime import get_current_year
 
 
 class OverallSerializer(serializers.Serializer):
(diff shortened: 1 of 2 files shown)


src/etools/applications/hact/management/commands/freeze_hact_data.py
```diff
@@ -106,3 +111,4 @@ def handle(self, *args, **options):
 
                 plan, _ = PlannedEngagement.objects.get_or_create(partner=partner)
                 plan.reset()
+                PartnerPlannedVisits.objects.get_or_create(partner=partner, year=datetime.today().year + 1)
```
robertavram, line 114: PartnerPlannedVisits is only used for Government Partners. Any thoughts whether we still need this here

src/etools/applications/field_monitoring/analyze/serializers.py
```diff
@@ -40,7 +40,7 @@ class Meta:
     def get_planned_visits(self, obj):
         try:
             return obj.planned_visits.get(year=get_current_year()).total
-        except obj.DoesNotExist:
+        except PartnerPlannedVisits.DoesNotExist:
```
robertavram, line 43: This would be an issue. PartnerPlannedVisits is not correctly displayed if taken from here as CSOs don't use this model. I think we can probably use the hact values to be consistent PartnerOrganization (obj) obj.hact_values["programmatic_visits"]["planned"]["total"] i think that';d do it

robertavram (changes requested):

robertavram:
@domdinicola tests failed

PR #2952, opened 2021-02-24
sentry staging 2719 RAMSynchronizer fix

domdinicola:


src/etools/applications/reports/synchronizers.py
```diff
@@ -341,10 +341,10 @@ def _clean_records(self, records):
             a = r['WBS_ELEMENT_CODE']
             code = str(r['INDICATOR_CODE'])
             mapped_records[code] = {
-                'name': r['INDICATOR_DESCRIPTION'][:1024],
-                'baseline': r['INDICATOR_BASELINE'][:255],
+                'name': str(r['INDICATOR_DESCRIPTION'])[:1024],
```
robertavram, line 344: set to empty string or None rather then the string "None" as there's a filter on line331

src/etools/applications/reports/synchronizers.py
```diff
@@ -341,10 +341,10 @@ def _clean_records(self, records):
...
+                'baseline': str(r['INDICATOR_BASELINE'])[:255],
                 'code': code,
-                'target': r['INDICATOR_TARGET'][:255],
+                'target': str(r['INDICATOR_TARGET'])[:255],
```
robertavram, line 347: an indicator without a target is meaningless, I would suggest to use the same logic as for the name and filter these out.

robertavram (changes requested):

src/etools/applications/reports/synchronizers.py
domdinicola: this was @ntrncic input

ntrncic: I meant only for the baseline. @domdinicola

PR #2953, opened 2021-02-24
Sentry 1193

domdinicola:


src/etools/applications/management/views/general.py
```diff
@@ -32,14 +32,9 @@ def get(self, request, format=None):
         except KeyError:
             return Response(status=400, data={'error': 'You must pass "area" as a query param'})
 
-        try:
-            country = Country.objects.get(business_area_code=business_area)
-        except Country.DoesNotExist:
```
robertavram, line 37: use an exists query on the business area code to ensure it's valid and in our database

PR #2970, opened 2021-03-17
[ch15114] psea assessment: added sea_risk_rating filter

domdinicola:
[ch15114]


acory:
@robertavram @domdinicola  
I tested the integration with FE. Filtering doesn't work.
There is a discrepancy between available sea risk rating values in the filter (values extended in PR https://github.com/unicef/etools/pull/2937) and the values taken into consideration here ('high','moderate','low')

PR #2971, opened 2021-03-18
[ch25270] FM Monitoring Activities: migrate single field office to multiple

domdinicola:
[ch25270]


acory:
@robertavram @domdinicola 
I tested the integration with FE (https://github.com/unicef/etools-fm-fe/pull/110)  and it works.

PR #2977, opened 2021-04-08
add key param for bank details from vision

ntrncic:


ntrncic:
not sure if we should put an if statement to check if the key is there. Otherwise we have to time our production push with insight team.

robertavram:
> not sure if we should put an if statement to check if the key is there. Otherwise we have to time our production push with insight team.

I would suggest to only put the key param if the key param is not none.. default in the settings can be none.

PR #3005, opened 2021-05-20
[ePD] New field structure for Input level from PD activities

roman-karpovich:
[ch25801]


src/etools/applications/reports/models.py
```diff
@@ -962,6 +963,19 @@ class InterventionActivityItem(TimeStampedModel):
...
+    unit = models.CharField(
+        verbose_name=_("Unit"),
+        max_length=150,
+    )
+    unit_price = models.DecimalField(
+        verbose_name=_("Unit Price"),
+        decimal_places=2,
+        max_digits=20,
+    )
+    no_units = models.IntegerField(
```
robertavram, line 975: no_units can be sub-unitary . If you have the unit in months, I can say 0.5 months. Validation can be greater than 0

PR #3007, opened 2021-05-21
[ePD] Allow Supply Contribution to be made by Partner also

roman-karpovich:
[ch25799]


src/etools/applications/partners/models.py
```diff
@@ -2676,7 +2685,7 @@ def total_unicef_contribution_local(self):
         return self.unicef_cash_local + self.in_kind_amount_local
 
     def total_cash_local(self):
-        return self.partner_contribution_local + self.unicef_cash_local
+        return self.total_partner_contribution_local + self.unicef_cash_local
```
robertavram, line 2688: `total_partner_contribution_local` seems to include supply and this method should return cash only

roman-karpovich: updated

PR #3049, opened 2021-07-28
[ePD] Backend error on Exporting list of agreements

roman-karpovich:
[ch24823]


src/etools/applications/partners/views/agreements_v3.py
```diff
@@ -41,7 +41,7 @@ def get_serializer_class(self, format=None):
             query_params = self.request.query_params
             if "format" in query_params.keys():
                 if query_params.get("format") in ['csv', 'csv_flat']:
-                    return self.get_serializer(query_params.get("format"))
+                    return AgreementListSerializer
```
domdinicola, line 44: can you set serializer_class on the view?

PR #3051, opened 2021-08-04
[ch26389] fix fr sync

domdinicola:
[ch26389]


robertavram:
from the call: migration to make line item nullable

PR #3075, opened 2021-08-26
[ePD] Update etools-validator package & fix tests

roman-karpovich:


robertavram:
can we update the libraries in develop branch? any library updates should be done in develop to avoid awful to resolve conflicts later on

PR #3076, opened 2021-08-26
[FAM] Add search functionality to users api

roman-karpovich:


src/etools/applications/users/views_v3.py
```diff
@@ -63,10 +63,11 @@ class UsersListAPIView(QueryStringFilterMixin, ListAPIView):
     Country is determined by the currently logged in user.
     """
     model = get_user_model()
-    queryset = get_user_model().objects.all()
+    queryset = get_user_model().objects.all().select_related('profile')
     serializer_class = MinimalUserSerializer
     permission_classes = (IsAdminUser, )
     pagination_class = AppendablePageNumberPagination
+    search_terms = ('email__icontains',)
```
robertavram, line 70: We have to search across the names as well

robertavram (changes requested):

src/etools/applications/users/views_v3.py
roman-karpovich: updated

PR #3093, opened 2021-09-22
[ePD] When partner or unicef focal points are updated on the main intervention, any amendments do not reflect the updates

roman-karpovich:


src/etools/applications/partners/signals.py
```diff
@@ -38,3 +38,33 @@ def create_review_for_prc_officer(sender, instance, action, reverse, pk_set, *ar
...
+            amendment.amended_intervention.unicef_focal_points.remove(*pk_set)
+
+
+@receiver(m2m_changed, sender=Intervention.partner_focal_points.through)
+def sync_partner_focal_points_to_amendment(sender, instance: Intervention, action, reverse, pk_set, *args, **kwargs):
+    active_amendments = instance.amendments.filter(is_active=True)
+
+    for amendment in active_amendments:
+        if action == 'post_add':
+            amendment.amended_intervention.partner_focal_points.add(*pk_set)
```
robertavram, line 60: what happens if this is a "set" rather than and add or remove call?

roman-karpovich: `set` uses combination of `add` and `remove`. checked with custom test - still working

PR #3094, opened 2021-09-22
[PMP] [ch26828 - 1] Disallow similar pairs of intervention + cp output for result links

roman-karpovich:
Story details: https://app.shortcut.com/unicefetools/story/26828


domdinicola:
what if already have not valid pairs?

roman-karpovich:
> what if already have not valid pairs?

there will be validation error on saving invalid objects

PR #3099, opened 2021-09-24
[ch26905] invalid country (no BA) when adding vendor

domdinicola:
[ch26905]

### DIFF
--- a/src/etools/applications/partners/views/partner_organization_v2.py
+++ b/src/etools/applications/partners/views/partner_organization_v2.py
@@ -490,6 +490,9 @@ class PartnerOrganizationAddView(CreateAPIView):
                             status=status.HTTP_400_BAD_REQUEST)
 
         country = request.user.profile.country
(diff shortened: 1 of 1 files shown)


domdinicola:
https://etools.unicef.org/admin/users/country/14/history/

src/etools/applications/partners/views/partner_organization_v2.py
```diff
@@ -490,6 +490,9 @@ def create(self, request, *args, **kwargs):
                             status=status.HTTP_400_BAD_REQUEST)
 
         country = request.user.profile.country
+        if not country.business_area_code:
```
robertavram, line 493: when adding vendor we should not need a business area code? where are we getting into this situation?

domdinicola: I think is this line:
not (business_area_code or self.GLOBAL_CALL)
when business_area_code is "" (empty string)

PR #3100, opened 2021-09-24
[ch26898] fix hact export

domdinicola:
[ch26898]


robertavram:
in staging already: #3086

PR #3129, opened 2021-11-05
[1] [ch26706] clean json exports

domdinicola:
[ch26706]


src/etools/applications/hact/migrations/0004_string2json.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 3.2.6 on 2021-11-05 17:38
+import json
+
+from django.db import migrations
+
+
+def string2json(apps, schema_editor):
```
robertavram, line 7: let's run it as a script

PR #3131, opened 2021-11-08
[ePD] 0.00 totals displayed after adding/editing Result Structure PD Activities

roman-karpovich:


src/etools/applications/reports/management/commands/fix-activity-totals.py
```diff
@@ -0,0 +1,34 @@
+import logging
```
robertavram, line 1: we can get rid of this as it's not supposed to happen typically. Initially i can run something similar manually if needed

src/etools/applications/reports/serializers/v2.py
```diff
@@ -533,6 +533,15 @@ def __init__(self, *args, **kwargs):
         self.intervention = kwargs.pop('intervention')
         super().__init__(*args, **kwargs)
 
+    def validate(self, attrs):
+        attrs = super().validate(attrs)
+        if self.instance and self.partial and 'items' not in attrs and self.instance.items.exists():
+            # when we do partial update for activity having items attached without items provided in request
+            # it's easy to break total values, so we ignore them
+            attrs.pop('unicef_cash', None)
+            attrs.pop('cso_cash', None)
```
robertavram, line 542: 👍

robertavram (changes requested):
we can remove the command

src/etools/applications/reports/management/commands/fix-activity-totals.py
roman-karpovich: okay. i thought it will be good to have this part tested if we'll need to run it

roman-karpovich: removed

PR #3136, opened 2021-11-11
[ePD] Allow indicator title being edited

roman-karpovich:


roman-karpovich:
i've checked pmp -> prp synchronization and it works good in case of editable indicator title

src/etools/applications/reports/serializers/v2.py
```diff
@@ -205,30 +210,45 @@ def validate(self, attrs):
...
+    @transaction.atomic
+    def update(self, instance, validated_data):
+        if 'indicator' in validated_data and self.partial and instance.indicator:
+            indicator_data = validated_data.pop('indicator')
+            lower_result = getattr(self.instance, 'lower_result')
+            if 'title' in indicator_data:
+                if lower_result.applied_indicators.filter(
+                    indicator__title=indicator_data['title']
+                ).exclude(indicator=instance.indicator).exists():
+                    raise ValidationError({
```
robertavram, line 242: instead of raising, create a new blueprint on this exception and re-associate.

roman-karpovich: in this case validation in .create will be broken. IndicatorBlueprintCUSerializer use get_or_create in .save, so we'll receive MultipleObjectsReturned exception.

---

test case:

- create indicator1 with title A - ok
- create indicator2 with title B - ok
- create indicator3 with title A - validation error (indicator is already being monitored), as before
- rename indicator2 to A - ok, new blueprint created according to the logic in message
- create indicator4 with title A - exception in IndicatorBlueprintCUSerializer.create

PR #3148, opened 2021-12-02
[ePD] Amendments: Sites added in Geographical Coverage in amendment mode are not showed as Difference in Amendment details

roman-karpovich:


src/etools/applications/partners/amendment_utils.py
```diff
@@ -558,7 +558,6 @@ def calculate_difference(instance, instance_copy, fields_map, relations_to_copy,
         'created', 'modified',
         'number', 'status', 'in_amendment',
         'title',
-        'sites',
```
robertavram, line 561: it's correct to have sites as IGNORED_FIELD. we should not change sites in the amendment. Sites like programmatic visits are editable at any point in the PD directly.

robertavram (changes requested):

robertavram:
I am updating the ticket

roman-karpovich:
updated. now sites should not be available during amendment mode (read/write permissions are False)

PR #3151, opened 2021-12-07
[ePD] PMP PD should not transition to Signed without signed PCA

roman-karpovich:


robertavram:
I don't think we need this anymore.. It looks like the logic is already there.

PR #3160, opened 2021-12-17
[ePD] 'Delete PD' not present in available_actions

roman-karpovich:
Delete only available if it was never sent to partner. 1. date_sent_to_partner is null. 2. Snapshot model does not include any changes to the date_sent_to_partner field.


src/etools/applications/partners/serializers/interventions_v3.py
```diff
@@ -354,6 +355,13 @@ def _is_partnership_manager(self):
             profile__country=self.context['request'].user.profile.country
         ).exists()
 
+    def _is_representative_office_manager(self):
```
robertavram, line 358: why do we need this?

roman-karpovich: it comes from PartnershipManagerRepPermission. not sure if we really need this group here

PR #3164, opened 2021-12-20
fix freeze hact data command

domdinicola:


src/etools/applications/hact/management/commands/freeze_hact_data.py
```diff
@@ -37,13 +37,13 @@ def freeze_data(self, hact_history):
             ('Partner Type', partner.partner_type),
             ('Shared IP', partner.shared_with),
             ('Assessment Type', partner.type_of_assessment),
-            ('Cash Transfer 1 OCT - 30 SEP', partner.net_ct_cy),
-            ('Liquidations 1 OCT - 30 SEP', partner.reported_cy),
-            ('Cash Transfers Jan - Dec', partner.total_ct_ytd),
+            ('Cash Transfer 1 OCT - 30 SEP', float(partner.net_ct_cy)),
```
robertavram, line 40: can we add the CustomJsonEncoder to the model?

robertavram (changes requested):

src/etools/applications/hact/management/commands/freeze_hact_data.py
domdinicola: absolutely

PR #3167, opened 2022-01-03
Fix for TypeError: Object of type Decimal is not JSON serializable

emaciupe:
https://excubo.unicef.io/sentry/etools-prod/issues/5943/


domdinicola:
addressed by https://github.com/unicef/etools/pull/3164

emaciupe:
already fixed in https://github.com/unicef/etools/pull/3164/

PR #3176, opened 2022-01-07
[ch28019] Error on running migrations on etools-test

emaciupe:
[ch28019]


src/etools/applications/audit/migrations/0022_engagement_reference_number.py
```diff
@@ -10,7 +10,7 @@ def set_reference_number(apps, schema_editor):
         schema_name=connection.tenant.schema_name,
     )
     for engagement in Engagement.objects.all():
-        if engagement.engagement_type == engagement.TYPES.audit:
+        if engagement.engagement_type == 'audit':
```
emaciupe, line 13: @robertavram @domdinicola if you have a better solution here, please jump in.

PR #3177, opened 2022-01-11
[TPM] Can't edit user in TPM, FAM

roman-karpovich:
Story details: https://app.shortcut.com/unicefetools/story/26464

fixed instance usage for EmailValidator. by design, WritableNestedChildSerializerMixin can set instance directly to validator, but only if such attribute exists

### DIFF
--- a/src/etools/applications/users/validators.py
+++ b/src/etools/applications/users/validators.py
@@ -38,7 +38,6 @@ class EmailValidator(UniqueValidator):
             message="This user already exists in the system.",
         )
 
-
     def exclude_current_instance(self, queryset, instance):
         """
         If an instance is already assigned to validator itself, prefer it.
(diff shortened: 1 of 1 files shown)


robertavram:
I don't think we need to fix this for now, I believe it to be part of a larger discussion, the only thing we need is to be able to "Activate" and "deactivate" a user.

roman-karpovich:
> I don't think we need to fix this for now, I believe it to be part of a larger discussion, the only thing we need is to be able to "Activate" and "deactivate" a user.

we're unable to update is_active user field without a fix to email validator, so it's necessary for having activate/deactivate functionality working

robertavram:
> > I don't think we need to fix this for now, I believe it to be part of a larger discussion, the only thing we need is to be able to "Activate" and "deactivate" a user.
> 
> we're unable to update is_active user field without email validator being fixed, so it's necessary for having activate/deactivate functionality working

in theory we could through a separate endpoint. keeping this parked as there is a more comprehensive user access management approach in discussion

robertavram:
will be changed with amp

PR #3185, opened 2022-01-14
[ePD] Terminate action should not break due to validation on quarters and activities.

roman-karpovich:


robertavram (changes requested):
please see the comment in the shortcut card

roman-karpovich:
> please see the comment in the shortcut card

i see no new comments in the card. functionality should be working according to the initial comment

robertavram:
When terminating, the user is asked to provide a termination date that is typically in the future. That forces the end date of the PD to be set at the termination date. This will create a problem when activity quarters can't be automatically set.
When delivering this card, set the activity quarters to end at the termination date (new end date) and cut the quarters from the end, regardless if they have activities to them (activities should not be deleted just the quarters that are no longer there).
If an activity only had quarters after the termination date, those activities should have no quarters. The validation should keep count of the fact that in "Terminated" state the activities don't have to have a quarter.

PR #3186, opened 2022-01-17
cleaning profile

domdinicola:


robertavram:
we should keep these fields for now. eventually we'll want to use this in Trips.

PR #3190, opened 2022-01-31
[ch28232] sentry-6060 TypeError Admin activate partner staff member

emaciupe:
[ch28232]

### DIFF
--- a/src/etools/applications/partners/forms.py
+++ b/src/etools/applications/partners/forms.py
@@ -92,8 +92,10 @@ class PartnerStaffMemberForm(forms.ModelForm):
             if active and not self.instance.active:
                 try:
                     user = User.objects.get(email=email)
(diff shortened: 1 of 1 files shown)


domdinicola (approved):

src/etools/applications/partners/forms.py
```diff
@@ -92,8 +92,10 @@ def clean(self):
             if active and not self.instance.active:
                 try:
                     user = User.objects.get(email=email)
-                except User.DoesNotExist():
-                    pass
+                except User.DoesNotExist:
+                    raise ValidationError(
```
robertavram, line 96: This is fine, we don't need to change anything. the exception is handled, sentry seems to have some odd behavior where it's reporting exceptions that are handled. so weird...

PR #3191, opened 2022-02-01
improve location logging

domdinicola:


src/etools/libraries/locations/tasks.py
```diff
@@ -29,11 +30,18 @@ def __init__(self, pk, schema) -> None:
...
-        self.log.total_processed = new + updated
-        self.log.successful = True
-        self.log.save()
+        try:
+            new, updated, skipped, error = super().sync()
+            self.log.total_records = new + updated + skipped + error
+            self.log.total_processed = new + updated
+            self.log.successful = True
+            self.log.exception_message = 'Congrats: Success'
+            self.log.save()
```
emaciupe, line 39: @domdinicola you can remove this line, it always runs in the `finally` block

emaciupe (changes requested):

src/etools/libraries/locations/tasks.py
domdinicola: yep!

PR #3193, opened 2022-02-02
[ch28271] ePD: workaround if missing https schema for env var DJANGO_ALLOWED_HOST/settings.HOST

emaciupe:
[ch28271]


robertavram:
not needed in our case

PR #3222, opened 2022-03-11
[ch28706] [ePD] fix for Locations CSV Export when no cp_output is associated

emaciupe:
[ch28706]


src/etools/applications/partners/models.py
```diff
@@ -2214,7 +2214,7 @@ def days_from_last_pv(self):
 
     @property
     def cp_output_names(self):
-        return ', '.join(link.cp_output.name for link in self.result_links.all())
+        return ', '.join(link.cp_output.name for link in self.result_links.all() if link.cp_output)
```
domdinicola, line 2217: I guess would be better add the filter instead of all

domdinicola (approved):

PR #3224, opened 2022-03-14
[ch28688] Fix random test failures related to user emails

emaciupe:
[ch28688]

### DIFF
--- a/src/etools/applications/firms/tests/factories.py
+++ b/src/etools/applications/firms/tests/factories.py
@@ -12,18 +12,19 @@ from etools.applications.users.tests.factories import ProfileFactory
 class BaseUserFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = get_user_model()
-        django_get_or_create = ("username", "email", )
+        django_get_or_create = ("username", )
 
-    username = factory.Faker('email')
+    username = factory.Sequence(lambda n: 'person{}.example.com'.format(n))
     first_name = factory.Faker('first_name')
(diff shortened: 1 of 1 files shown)


src/etools/applications/firms/tests/factories.py
```diff
@@ -12,18 +12,19 @@
...
         model = get_user_model()
-        django_get_or_create = ("email", )
+        django_get_or_create = ("username", )
 
-    username = factory.Faker('email')
+    username = factory.Sequence(lambda n: 'person{}@example.com'.format(n))
     first_name = factory.Faker('first_name')
     last_name = factory.Faker('last_name')
-    email = factory.Faker('email')
+    email = factory.Sequence(lambda n: 'person{}@example.com'.format(n))
```
robertavram, line 20: will this work when running tests in parallel?

src/etools/applications/firms/tests/factories.py
```diff
@@ -12,18 +12,19 @@
 class BaseUserFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = get_user_model()
-        django_get_or_create = ("email", )
+        django_get_or_create = ("username", )
```
domdinicola, line 15: why did we change this?

emaciupe: @domdinicola forgot to remove it, thanks.By using factory Sequence, email and username they are always unique so no need for get_or_create

PR #3227, opened 2022-03-17
[ch28785] flat attachments for FAM and PSEA

domdinicola:
[ch28785]


domdinicola:
migrating existing attachments

```
from etools.applications.attachments.models import *
from etools.applications.attachments.utils import *
for country in Country.objects.exclude(schema_name='public'):
    set_country(country)
    for att in AttachmentFlat.objects.filter(source='Financial Assurance (FAM)'):
        att.agreement_reference_number = get_reference_number(att.attachment)
        att.save()
```

src/etools/applications/attachments/utils.py
```diff
@@ -114,13 +116,17 @@ def get_agreement_obj(obj):
     )):
         if obj.content_object.intervention:
             return obj.content_object.intervention.agreement
+    elif isinstance(obj.content_object, Engagement):
```
emaciupe, line 119: @domdinicola you could remove these lines and extend on line 108 e.g.
```
if isinstance(obj.content_object, (Agreement)):
       return obj.content_object
```
to
```
if isinstance(obj.content_object, (Agreement, Engagement, Assessment)):
       return obj.content_object
```

emaciupe (changes requested):

src/etools/applications/attachments/utils.py
domdinicola: yep, I guess make sense [I was thinking to have one entry per module but anyways is not the case]

domdinicola:
@emaciupe did you review also the script?

emaciupe (approved):
Looks good to me!

PR #3228, opened 2022-03-18
[ch28830] Admin integrity err on total_value and exchange_rate blank

emaciupe:
[ch28830]


domdinicola:
@emaciupe to me migrations would be better... @robertavram what do you recommend?

PR #3229, opened 2022-03-22
update_audit_permissions cmd: unique_id instead of reference_number

emaciupe:
@domdinicola @robertavram any particular reason why ```unique_id``` was also introduced  for ```reference_number``` ? should we have this consistent across the project ?


emaciupe:
[ch28876]

domdinicola:
@emaciupe we migrated recently "reference_number" from property to field, most probably this caused the issue

domdinicola:
I think we should always use reference_number but this needs to be synchronized with the FE

emaciupe:
> @emaciupe we agreed to use reference_number everywhere

right,  then merging develop into staging will wait until the unique_id is replaced otherwise you will bump into the same bug this PR is for, breaking FAM engagements.

domdinicola:
@emaciupe correct, this need to be coordinated with a front-end PR

PR #3243, opened 2022-04-05
[ch28908]Trips: exclude Submitted from status_list when travel has status "Approved"

emaciupe:
[ch28908]


emaciupe:
@robertavram Removed "Submitted" from status list when trip is completed, is this correct?

PR #3244, opened 2022-04-05
[ch28879]Trips: Updating activity from Monitoring Activity to other type cleans up the monitoring activity

emaciupe:
[ch28879]


robertavram (changes requested):
as per call

PR #3277, opened 2022-05-03
[ch29098] Sentry-7552 Validate report requirements start/end_date payload for null values

emaciupe:
[ch29098]


src/etools/applications/partners/serializers/interventions_v2.py
```diff
@@ -883,6 +883,8 @@ def run_validation(self, initial_data):
         serializer.fields["end_date"].required = True
         serializer.fields["due_date"].required = True
 
+        serializer.fields["start_date"].allow_null = False
```
domdinicola, line 886: just curious.. why aren't we setting this in the serializer?

robertavram:
not needed for now

PR #3291, opened 2022-05-11
[ch29647] FAM: Check if the attachments are related to the engagement (id in the url)

emaciupe:
[ch29647]


domdinicola (approved):

src/etools/applications/audit/views.py
```diff
@@ -645,10 +646,13 @@ def get_parent_filter(self):
         }
 
     def get_object(self, pk=None):
+        if self.request.method in ['GET', 'DELETE']:
```
robertavram, line 649: get patch delete

PR #3294, opened 2022-05-13
permission tweaks

robertavram:


robertavram:
@emaciupe if you could fix the tests, it would be super!Thanks!

PR #3300, opened 2022-05-20
[ch29820] Pd indicator title to be editable while PD is in Developmen…

merceaemil:
…t status


emaciupe (approved):
Looks good!

robertavram:
This validation will be handled by the permission matrix. The backend for this ticket was done here: #3136 .

PR #3348, opened 2022-09-16
[ePD] Improve pd import from ecn

roman-karpovich:


src/etools/applications/ecn/serializers.py
```diff
@@ -193,6 +194,9 @@ def create(self, validated_data):
             blueprint = indicator_blueprint_serializer.instance
 
         validated_data['indicator'] = blueprint
+        # save section & locations manually, because PrimaryKeyRelatedField casts value to instance and then fails
+        validated_data['section'] = self.root.instance.sections.first()
```
robertavram, line 198: will this not override the user selected section in the case where user selects a section.
Should we do a check like "if not validated_data.get('section', None)" then add them?
same for locations below?

roman-karpovich: in theory it will, but there is no way for section to appear in indicator data, because it comes from ecn which has no such concept as section

robertavram: how about when we add an indicator from the frontend of eTools?

roman-karpovich: this serializer is for ecn import only

robertavram: omg.. I didn't even see the file location I just assumed. cool. merging! THANKS!

PR #3350, opened 2022-09-19
[ch31629] ePD: Allow edit for partners and reports app models only for given emails as env vars

emaciupe:
[ch31629]


robertavram (approved):
🥇

PR #3353, opened 2022-09-20
update local settings

robertavram:
work with jwt RS256 locally
enable testing locally on multiple threads, adjust the threads wih an env var


emaciupe (approved):
tests fail but are fixed in https://github.com/unicef/etools/pull/3354

PR #3363, opened 2022-10-17
Epd amp

robertavram:


src/etools/applications/users/migrations/0021_migrate_to_realms.py
```diff
@@ -0,0 +1,280 @@
...
+                                organization_id=tpm_organization.id,
+                                group_id=tpm_group.id,
+                                is_active=user.is_active
+                            ))
+            # check if user is a partner staff member for each country tenant
+            for country in countries:
+                connection.set_tenant(country)
+                if hasattr(user, 'partner_staff_member') and user.partner_staff_member:
+                    partner_staff = user.partner_staff_member
+                    if not partner_staff.partner.organization:
```
roman-karpovich, line 198: this migration doesn't work. it fails with missing column error on tenant models.
we're trying to use tenant models (PartnerOrganization, PartnerStaffMember, AuditorStaffMember, TPMStaffMember, AuditorFirm, etc) despite users application is in public schema. we should split this particular migration into public part (users, profiles, organizations) and tenant part (organizations and staff members).

roman-karpovich: #3391

PR #3381, opened 2022-11-15
[ePD amp] remove is_active Realm field

roman-karpovich:


roman-karpovich:
motivation: now we have two logical branches (missing realm and inactive realm) which we treat identically like "you have no power here" yet processing in two different ways. it makes amp code more complex

PR #3383, opened 2022-11-21
[ePD] Adjustment to ePD to include adjustments from the procedure

roman-karpovich:


roman-karpovich:
@robertavram i've implemented transition logic as written in the card

> new validation needs to be added on the transition to signed. The document cannot transition to signed if any of the newly added fields are true and there is no corresponding attachment (attachment type). Basically if a field is checked sych as "Data Processing" then there needs to be an attachment of type "Data Processing Agreement" in order for the document to transition to signed.

but i have concerns about it. maybe it would be better to move new attachments validation to accept step instead of signature? 
otherwise we have next problems:
1. attachments are editable only in draft status, so now it's impossible to edit them on signature stage as well as we cannot go back to upload missing attachments (and even if we allow this, intervention auto transitions don't work on attachments update, so we'll need to tweak transitions code)
2. it's not obvious what's wrong with our document, because there are no readable validation messages on auto transition. document just silently wait till ticked documents are provided.

src/etools/applications/partners/serializers/exports/interventions.py
```diff
@@ -481,6 +493,15 @@ def get_planned_visits(self, obj):
             pv.year, pv.programmatic_q1, pv.programmatic_q2, pv.programmatic_q3, pv.programmatic_q4
         ) for pv in obj.planned_visits.all()])
 
+    def get_has_data_processing_agreement(self, obj):
+        return "Yes" if obj.contingency_pd else "No"
```
robertavram, line 497: check this

src/etools/applications/core/data/attachments_file_types.json
```diff
@@ -718,5 +718,38 @@
             "code": "special_audit_final_report",
             "group": ["special_audit_final_report"]
         }
+    },
+    {
+        "model": "unicef_attachments.filetype",
+        "pk": 2104,
```
robertavram, line 724: 2012

roman-karpovich: updated

src/etools/applications/partners/serializers/exports/interventions.py
roman-karpovich: updated

PR #3391, opened 2022-11-29
[ePD-amp] Update migrations

roman-karpovich:


emaciupe (approved):
@roman-karpovich thanks for refactoring, that's the way to do it!

PR #3403, opened 2022-12-14
[ch32541] ePD: implement logout with redirect azure adb2c

emaciupe:
[ch32541]


domdinicola:
in unicef-security we have https://github.com/unicef/unicef-security/blob/develop/src/unicef_security/views.py#L17

robertavram:
> @emaciupe @robertavram FYI in unicef-security we have https://github.com/unicef/unicef-security/blob/develop/src/unicef_security/views.py#L17

thanks, I saw that, i believe we're not using unicef-security at the moment in eTools. Wouldn't want to plug it for one view only. We'll have to sit down and look at our libs together, map out system dependencies, releases and figure out which functions go where. 

After ePD and user management release, I'll set up something with all backend devs.

PR #3407, opened 2022-12-15
[ePD-amp] Switch to Realms instead of TPM staff member

roman-karpovich:


src/etools/applications/tpm/tpmpartners/synchronizers.py
```diff
@@ -86,14 +86,15 @@ def _save_records(self, records):
...
+        # TODO: REALMS - do cleanup
+        # staff_members = partner.staff_members.all()
+        # # deactivate the users
+        # users_deactivate = User.objects.filter(tpmpartners_tpmpartnerstaffmember__in=staff_members)
+        # users_deactivate.update(is_active=False)
         try:
             country = Country.objects.get(name=partner.country)
             Realm.objects\
-                .filter(user__in=users_deactivate, country=country, organization=partner.organization)\
+                .filter(country=country, organization=partner.organization)\
```
emaciupe, line 97: @roman-karpovich i guess here the filtering isn't too broad - just to confirm that when a partner is deleted, then all realms are to be made inactive, regardless of the group.

emaciupe (approved):
Looks good to me!

src/etools/applications/tpm/tpmpartners/synchronizers.py
roman-karpovich: #3414 3414

PR #3411, opened 2022-12-29
[ch32617] FM: custom translations field for db translations

emaciupe:
[ch32617]


robertavram (changes requested):
please update as per our call.

PR #3428, opened 2023-01-31
When exporting comments file, the deleted comments should not be visible

emaciupe:
[ch32806]


emaciupe:
@robertavram excluding deleted state for all content types, let me know it this is only for interventions

PR #3429, opened 2023-01-31
on PDF export of PD, cash transfer modality shows up unkown

emaciupe:
[ch32809]


robertavram (changes requested):
```
File "/root/code/src/etools/applications/partners/models.py", line 2622, in get_cash_transfer_modalities_display
    return ', '.join([choices.get(m, 'Unknown') for m in self.cash_transfer_modalities])
TypeError: sequence item 0: expected str instance, __proxy__ found
```

robertavram (approved):

PR #3438, opened 2023-02-14
[ch32786] Update attachments permissions

roman-karpovich:


roman-karpovich:
linked to https://github.com/unicef/unicef-attachments/pull/40

emaciupe (approved):

roman-karpovich:
tests should be fine after 0.13 unicef attachments release. 
https://github.com/unicef/unicef-attachments/pull/40

src/etools/applications/attachments/permissions.py
```diff
@@ -1,8 +1,31 @@
...
+    def has_permission(self, request, view):
+        return True
+
+    def has_object_permission(self, request, view, obj):
+        content_object = obj.content_object
+        if not content_object:
+            return True
+
+        if hasattr(content_object, 'get_related_third_party_users'):
+            return content_object.get_related_third_party_users().filter(pk=request.user.pk).exists()
```
robertavram, line 26: we need to make sure that the user has at least one active realm in the current tenant workspace

robertavram, line 26: maybe one approach is to send in the user to the instance method it return true or false. In some cases the user might have left one party like a Partner org, but may be working with another one in the same workspace. If the user doesn't have an active realm for the associated partner or vendor, etc... to whatever instance we're probing, then we should impose the limitation.

src/etools/applications/field_monitoring/fm_settings/models.py
```diff
@@ -288,3 +293,6 @@ def related_to_type(self):
             return self.RELATED_TO_TYPE_CHOICES.partner
         elif self.location:
             return self.RELATED_TO_TYPE_CHOICES.location
+
+    def get_related_third_party_users(self):
+        return get_user_model().objects.filter(realms__group=FMUser.as_group())
```
robertavram, line 298: this would be too broad as it filters all realms in the whole world. (may be worth considering a slightly different approach for this method)

robertavram (changes requested):
I love what you did here.. maybe minor tweaks.. I think we're ready to merge this in. Please review comments.

src/etools/applications/field_monitoring/fm_settings/models.py
```diff
@@ -34,6 +36,9 @@ def get_current(cls):
 
         return cls._config
 
+    def get_related_third_party_users(self):
+        return get_user_model().objects.filter(groups=FMUser.as_group())
```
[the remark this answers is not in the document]
roman-karpovich: this is related to Field monitoring, so we want to have FMUser group. but in fact field monitor is always UNICEF, so it can be safely removed

src/etools/applications/attachments/permissions.py
roman-karpovich: added IsActiveInCurrentSchema permission to check this

src/etools/applications/field_monitoring/fm_settings/models.py
roman-karpovich: agreed. removed it as Field monitors are not third party users, so we don't need it

src/etools/applications/attachments/permissions.py
roman-karpovich: added explicit filters inside

robertavram:
@roman-karpovich failing tests.

roman-karpovich:
> linked to [unicef/unicef-attachments#40](https://github.com/unicef/unicef-attachments/pull/40)

depends on https://github.com/unicef/unicef-attachments/pull/40

roman-karpovich:
> @roman-karpovich failing tests.

depends on https://github.com/unicef/unicef-attachments/pull/40

PR #3436, opened 2023-02-13
[ch32923] Strategy tab: Add `Other` field in the Document Details section

roman-karpovich:
Story details: https://app.shortcut.com/unicefetools/story/32923

PR #3439, opened 2023-02-14
[ch32388] Add new fields to eCN

roman-karpovich:
Story details: https://app.shortcut.com/unicefetools/story/32388


roman-karpovich:
depends on #3436

roman-karpovich:
UNPP number ecn import behavior is still subject of question - we already ask it in the import popup

emaciupe (approved):

PR #3441, opened 2023-02-15
[ch32484] Update toasts related to Headquarters Contribution tryout

emaciupe:
[ch32484]
{"planned_budget":{"Capacity Strengthening Costs":["This field may not be null."]}} 
into FR:
{"planned_budget":{"Coûts de renforcement des capacités":["Ce champ ne peut être nul."]}}
@robertavram 
To have "planned_budget" translated would require changes in etools-validator.ValidatorViewMixin library. Please advise.


robertavram:
I realized that this will create some issues. This is used by the FE (as it should be) to map to fields that are throwing the errors. I think there's a mapper on the FE that we can use to translate.. I think there's a helper that currently capitalizes and removes the `_`. Closing this for now.. leave the ticket for now.. it will be after release anyway.

PR #3446, opened 2023-02-21
Translation dupes on PD status

emaciupe:
Changed duplicated translations for Ended/Terminated statuses on fr and es


emaciupe:
included in https://github.com/unicef/etools/pull/3447

PR #3449, opened 2023-02-24
Backend sentry issues 

emaciupe:
[ch33134]
sentry 16128 ETOOLS-DEMO-2F: fix xls PD export


src/etools/applications/reports/synchronizers.py
```diff
@@ -298,7 +298,10 @@ def _clean_records(self, records):
         return {'cps': cps, 'outcomes': outcomes, 'outputs': outputs, 'activities': activities}
 
     def _convert_records(self, records):
-        records = records['ROWSET']['ROW']
+        try:
+            records = records['ROWSET']['ROW']
+        except KeyError as exc:
+            logger.warning(f'Expected key {exc} is missing from API response {records}.')
```
robertavram, line 304: we need to stop the execution here. Otherwise records will not exist on the next line

emaciupe: right, then raising VisionException

PR #3461, opened 2023-03-16
[ch33317] Fix bank details value not showing inside table

merceaemil:
[ch33317] Fix bank details value not showing inside table


emaciupe (approved):
Works now! the issue was that if the account title is None aka ACCT_HOLDER, it will break the table.

PR #3476, opened 2023-04-25
[ch33651] Get language from header request

merceaemil:
[ch33651] Get language from header request


src/etools/applications/core/middleware.py
```diff
@@ -111,8 +111,14 @@ class EToolsLocaleMiddleware(MiddlewareMixin):
     """
 
     def process_request(self, request):
+        header_language_code = request.META.get('HTTP_LANGUAGE', '')
+        if header_language_code and header_language_code in get_languages():
+            translation.activate(header_language_code)
```
robertavram, line 116: I think we shouldn't go through this if the user is not authenticated. Can we move the function below after we handle the `user.is_anonymous`

PR #3520, opened 2023-06-13
[FAM] MicroAssessments: update tested subject areas according to the new questionnaire

roman-karpovich:


emaciupe (approved):
looks good!

PR #3527, opened 2023-06-28
sentry-22592

emaciupe:
'NoneType' object has no attribute 'relationship_types'


src/etools/applications/partners/views/v3.py
```diff
@@ -40,9 +40,11 @@ class PMPBaseViewMixin:
...
 
     def is_partner_staff(self):
-        """Flag indicator whether user is a partner
-        any active group out of IP... , """
-        return self.request.user.is_authenticated and \
+        """
+        Flag indicator shows whether authenticated user is a partner staff
+        based on profile organization relationship_type
+        """
+        return self.request.user.is_authenticated and self.request.user.profile.organization and \
```
robertavram, line 47: @emaciupe I think this gets a little bit more hairy. 
in the world where we could only have either a partner staff or unicef the previous logic makes sense. With this override it might allow unwanted access for someone that is not unicef and does not have an organization set currently. see line 61 https://github.com/unicef/etools/pull/3527/files#diff-9128fd1eafe61a05da4d27ca1e96c37a6299933879f89090feee05efd9052937R63 for example. The PD gets returned..

robertavram (changes requested):
let's do a review of the use of the is_partner_staff method and make sure that there isn't an assumption that `if not is_partner_staff` that means unicef.

PR #3562, opened 2023-08-28
[PMP] Optimise interventions patch endpoint

roman-karpovich:


src/etools/applications/partners/views/interventions_v3.py
```diff
@@ -212,11 +212,6 @@ def update(self, request, *args, **kwargs):
             logging.debug(validator.errors)
             raise ValidationError(validator.errors)
 
-        if getattr(self.instance, '_prefetched_objects_cache', None):
```
robertavram, line 215: this is intentional

roman-karpovich:
we do get_object in the end of my_update so it shouldn't be needed

robertavram:
will close for now, if needed we'll revisit. Thanks!

PR #3569, opened 2023-09-13
Fix freanch and russian pca templates diffs

merceaemil:
Fix freanch and russian pca templates diffs

Following are the diff checks between original documents and the generated ones after fixes:
[diff-checker-french.pdf](https://github.com/unicef/etools/files/12597856/diff-checker-french.pdf)
[diff-checker-russian.pdf](https://github.com/unicef/etools/files/12597857/diff-checker-russian.pdf)


emaciupe (approved):
Looks good to me!

PR #3579, opened 2023-09-25
hotfix performance issues

robertavram:


src/etools/applications/partners/admin.py
```diff
@@ -297,7 +297,7 @@ class InterventionAdmin(
     search_fields = (
         'number',
         'title',
-        'agreement__partner__name'
+        'agreement__partner__organization__name'
```
emaciupe, line 300: currently working, as the PartnerOrganization objects are annotated and the base_manager_name overridden

PR #3621, opened 2024-01-16
[PMP] Amendment: Special Reports set are not displayed in merged amendment details and are not added to the PD/SD

roman-karpovich:
Story details: https://app.shortcut.com/unicefetools/story/33168


robertavram:
I believe the special reports need to be able to be updated on the go without an amendment. on hold until I can confirm

PR #3644, opened 2024-04-02
[FM] Add the ability to notify who will review the final summary analysis.

roman-karpovich:
Story details: https://app.shortcut.com/unicefetools/story/36557


src/etools/applications/field_monitoring/planning/models.py
```diff
@@ -240,6 +243,17 @@ class MonitoringActivity(
                                    verbose_name=_('Person Responsible'), related_name='+',
                                    on_delete=models.SET_NULL)
 
+    # report reviewer is being set in two steps. first we store the preliminary value to be used as initial for submit
+    report_reviewer_preliminary = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True,
```
robertavram, line 247: We don't really need the "preliminary" field. The `report_reviewer` is what gets set in the beginning and gets overridden if needed in the pop-up at the end.

robertavram (changes requested):
please remove the preliminary reviewer

src/etools/applications/field_monitoring/planning/models.py
roman-karpovich: based on how etools vanilla permissions system works we cannot properly separate permissions to provide ability to fill field as part of submit action but disallow it to be editable in submitted state

PR #3671, opened 2024-05-17
[Develop] fix follow up page permissions: show follow up page as early as we show action points

roman-karpovich:


roman-karpovich:
[ch37207]

PR #3678, opened 2024-05-28
PMP: pagination and ordering for partners and pd lists

emaciupe:
[ch31883]

### DIFF
--- a/src/etools/applications/partners/views/interventions_v3.py
+++ b/src/etools/applications/partners/views/interventions_v3.py
@@ -26,6 +26,7 @@ from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.status import is_success
 from rest_framework.views import APIView
+from unicef_restlib.pagination import DynamicPageNumberPagination
 
 from etools.applications.field_monitoring.permissions import IsEditAction, IsReadAction
 from etools.applications.partners.exports_v2 import InterventionXLSRenderer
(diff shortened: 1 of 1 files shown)


src/etools/applications/partners/tests/test_v3_interventions.py
```diff
@@ -287,22 +287,22 @@ def test_filter_editable_by_unicef(self):
...
-        self.assertEqual(len(response.data), 4)
+        self.assertEqual(len(response.data['results']), 4)
 
     def test_updated_country_programmes_field_in_use(self):
         intervention = InterventionFactory()
         country_programme = CountryProgrammeFactory()
         intervention.country_programmes.add(country_programme)
         InterventionFactory()
-        with self.assertNumQueries(10):
+        with self.assertNumQueries(11):
```
robertavram, line 297: why is this increasing the number of queries..

src/etools/applications/partners/tests/test_v3_interventions.py
```diff
@@ -287,22 +287,22 @@ def test_filter_editable_by_unicef(self):
             data={'editable_by': 'unicef'},
         )
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(len(response.data), 4)
+        self.assertEqual(len(response.data['results']), 4)
```
robertavram, line 290: It seems that this introduces breaking changes to the API by default, can we keep the resonse as is by default and only introduce the "results" or pagination if the corresponded query params are set?

src/etools/applications/partners/tests/test_v3_interventions.py
emaciupe: when using drf pagination, it's inherited from django,core.Paginator which will first try calling count(), then fallback to using len()  - so the count is the +1 increase

https://docs.djangoproject.com/en/4.2/topics/pagination/#the-paginator-class

src/etools/applications/partners/tests/test_v3_interventions.py
emaciupe: done, changed to the already implemented AppendablePageNumberPagination as pagination_class

PR #3703, opened 2024-06-28
[FM] Validate report reviewer

roman-karpovich:
Story details: https://app.shortcut.com/unicefetools/story/37260

### DIFF
--- a/src/etools/applications/field_monitoring/planning/serializers.py
+++ b/src/etools/applications/field_monitoring/planning/serializers.py
@@ -162,7 +162,6 @@ class MonitoringActivitySerializer(UserContextSerializerMixin, MonitoringActivit
         ]
 
     def validate_report_reviewer(self, value):
-        user = self.get_user()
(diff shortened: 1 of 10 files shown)


src/etools/applications/field_monitoring/planning/serializers.py
```diff
@@ -159,6 +161,16 @@ def get_transitions(self, obj):
             for transition in get_available_transitions(obj, self.get_user())
         ]
 
+    def validate_report_reviewer(self, value):
+        user = self.get_user()
+        if value == user:
```
robertavram, line 166: we need to exclude this validation for now.

robertavram (changes requested):
change in requirements (See comments on card)

src/etools/applications/field_monitoring/planning/serializers.py
roman-karpovich: updated

PR #3731, opened 2024-07-25
[ch37435] Action Points export patch tablib.XLSXFormat.dset_sheet to export data with non ascii

emaciupe:
[ch37435] patch tablib.XLSXFormat.dset_sheet to export data with non ascii


src/etools/applications/action_points/views.py
```diff
@@ -14,6 +14,7 @@
 from unicef_restlib.views import MultiSerializerViewSetMixin, SafeTenantViewSetMixin
 from unicef_snapshot.views import FSMSnapshotViewMixin
 
+import etools.applications.action_points.export.export_non_ascii_patch  # noqa F401
```
robertavram, line 17: why are we not explicitly importing what we need?

emaciupe: I must have missed that, changed.

PR #3735, opened 2024-07-30
[PMP] Request a change that only the PRC secretary role can cancel the PD

roman-karpovich:
Story details: https://app.shortcut.com/unicefetools/story/37643


src/etools/applications/partners/views/interventions_v3_actions.py
```diff
@@ -362,8 +362,12 @@ def update(self, request, *args, **kwargs):
         if pd.status == Intervention.CANCELLED:
             raise ValidationError(_("PD has already been cancelled."))
 
-        if self.request.user not in pd.unicef_focal_points.all() and self.request.user != pd.budget_owner:
-            raise ValidationError(_("Only focal points or budget owners can cancel"))
+        if not get_user_model().objects.filter(
```
robertavram, line 365: if the user is not PRC Secretary, can we remove the cancel option from the available_actions in the PD Detail Serializer?

robertavram (changes requested):

src/etools/applications/partners/views/interventions_v3_actions.py
roman-karpovich: it's done. `src/etools/applications/partners/serializers/interventions_v3.py` was updated also in this PR

PR #3770, opened 2024-09-20
[ch38143] Request to Disable Currency Change During PD Amendment

emaciupe:


src/etools/applications/partners/permissions.py
```diff
@@ -169,6 +169,15 @@ def not_spd(instance):
         def not_ssfa(instance):
             return not is_spd_non_hum(instance)
 
+        def unlocked_amendment_mode_without_fr(instance):
+            if not unlocked(instance):
+                return False
+
+            if not user_added_amendment(self.instance) or \
+                    (user_added_amendment(self.instance) and not self.instance.amendment.intervention.frs.exists()):
```
robertavram, line 177: no need to check if frs exist... the only check that needs to happen is if intervention is in_amendment_mode. if it's in amendment it does not allow for budget currency

PR #3772, opened 2024-09-24
[PMP] Add a review level for the Rep

roman-karpovich:
Story details: https://app.shortcut.com/unicefetools/story/31859


robertavram:
@emaciupe could we generate translations without fuzzy please

PR #3795, opened 2024-11-18
[ch37587] New field on api response 

smihalache06:


robertavram:
this approach is too specific. we need to use the material number and have  a custom mapping.. that can be done on the FE for now.

PR #3799, opened 2024-12-04
[ch37587]Add new field on the response API to be able to translate

smihalache06:
As discussed with Tiaan this morning, I added the new field on the endpoint to be able to make difference between RUTF materials on the FE based on the backend configuration.


robertavram:
I believe this approach is too specific. This would not be easily or elegantly scalable.. moving it forward... recommending thinking this after this release.

PR #3829, opened 2025-02-14
[ch39322] PRP endpoint for sync locations

emaciupe:
[ch39322]

### DIFF
--- a/src/etools/applications/locations/tests/test_prp_views.py
+++ b/src/etools/applications/locations/tests/test_prp_views.py
@@ -5,6 +5,7 @@ from rest_framework import status
 from unicef_locations.tests.factories import LocationFactory
 
 from etools.applications.core.tests.cases import BaseTenantTestCase
+from etools.applications.locations.models import Location
 from etools.applications.partners.permissions import READ_ONLY_API_GROUP_NAME
 from etools.applications.users.tests.factories import GroupFactory, UserFactory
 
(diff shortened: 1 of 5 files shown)


src/etools/applications/locations/tests/test_prp_views.py
```diff
@@ -0,0 +1,49 @@
...
+        response = self.forced_auth_req('get', self.url, user=self.unicef_staff, data=self.query_param_data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(sorted(response.data[0].keys()), self.expected_keys)
+        # sort the expected locations by name, the same way the API results are sorted
+        self.locations.sort(key=lambda location: location.name)
+
+        for i in range(len(self.locations)):
+            location = self.locations[i]
+            data = response.data[i]
+            self.assertEqual(data["name"], location.name)
```
fpbonet, line 38: Let's test the full json payload here. Will make the test fail if we mistakenly remove some field in the future. 
You could implement a self._assert_location_payload and in it check the fields for the location[i]. This should be easier with copilot or gpt :)

fpbonet (approved):
Approving with a comment, in case you want to merge and submit the fix later

src/etools/applications/locations/tests/test_prp_views.py
emaciupe: @fpbonet solved!

PR #3830, opened 2025-02-19
[LMSM] Admin panel Endpoints

smihalache06:


src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -0,0 +1,394 @@
...
+    def create(self, validated_data):
+        user_profile = validated_data.pop('profile', {})
+        group = Group.objects.get(name="IP LM Editor")
+        password = validated_data.pop('password', None)
+        country_schema = self.context.get('country_schema')
+        country = Country.objects.get(schema_name=country_schema)
+
+        try:
+            user = get_user_model().objects.create(**validated_data)
+            user.set_password(password)
```
fpbonet, line 72: Can we set the password when doing the objects.create ? That way we avoid an additional query

fpbonet: Or another idea:
```
user = get_user_model(**validated_data)
user.set_password(password)
user.save()
```

emaciupe (changes requested):

src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -0,0 +1,670 @@
...
+            self.fields[field].read_only = True
+
+    @transaction.atomic
+    def update(self, instance, validated_data):
+        self.adminValidator.validate_profile(instance)
+        poi_list = validated_data.pop('point_of_interest', None)
+        if poi_list is not None:
+            partner = instance.profile.organization.partner
+            partner.points_of_interest.set(poi_list)
+            instance.save()
```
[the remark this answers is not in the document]
smihalache06: No, only the locations of the user. The point of interests that are assigned to that user.

src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -0,0 +1,670 @@
...
+
+    adminValidator = AdminPanelValidator()
+
+    @transaction.atomic
+    def create(self, validated_data):
+        self.adminValidator.validate_items(validated_data.get('items', []))
+        self.adminValidator.validate_partner_location(validated_data.get('location'), validated_data.get('partner_organization'))
+        items = validated_data.pop('items', [])
+        instance = models.Transfer.objects.create(
+            unicef_release_order=f"{TRANSFER_MANUL_CREATION_NAME} {timezone.now().strftime('%d-%m-%Y %H:%M:%S')}",
```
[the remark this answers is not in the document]
smihalache06: It's a typo error. Will be corrected.

src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -0,0 +1,670 @@
...
+
+    def extract_location_info(self, location_obj):
+        location_data = {"district": None, "region": None, "country": None}
+
+        current = location_obj
+        while current:
+            admin_level = getattr(current, "admin_level_name", None)
+            name = getattr(current, "name", None)
+
+            if admin_level == "District":
```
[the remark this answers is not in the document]
smihalache06: @fpbonet , @emaciupe at the moment, for this admin panel we are interested only in this 3 cases, to show the Region, District and Country. Maybe in the future, we will add more.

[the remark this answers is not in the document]
smihalache06: I just saw that on Somalia we have only Country,Region and District. But on other countries we have multiple types of regions, so we need somehow to map them to some general naming. Because we are interested only on this 3 fields.

src/etools/applications/last_mile/admin_panel/validators.py
```diff
@@ -0,0 +1,60 @@
...
+from rest_framework.exceptions import ValidationError
+
+from etools.applications.last_mile.models import Material, PointOfInterest, PointOfInterestType
+from etools.applications.partners.models import PartnerOrganization
+from etools.applications.users.models import Group, Realm
+
+
+class AdminPanelValidator:
+
+    def validate_poi_type(self, value: str):
```
emaciupe, line 13: @smihalache06 the encapsulation in a class here is a bit pointless, self is never used. If keeping it in class feels cleaner for you, all methods should be @staticmethods  and no objects instantiation is then required in the serializer classes. IMHO they should all be functions, and where it is possible, used in the  field declaration in the serializers by adding the attr validators=[ <func_name>, <.. >]

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -0,0 +1,468 @@
...
+        'last_name',
+        'profile__organization__name',
+        'profile__organization__vendor_number',
+        'profile__country__name',
+        'profile__country__id',
+    ]
+
+    ordering = ('id',)
+
+    def get_serializer_context(self):
```
emaciupe, line 106: @smihalache06 Minor comment, based on less is more concept:  `get_serializer_context `could be removed, you can always import anywhere the db connection, in serializers module as well and use it directly

emaciupe (changes requested):

src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -0,0 +1,670 @@
...
+
+
+class PointOfInterestAdminSerializer(serializers.ModelSerializer):
+    partner_organizations = SimplePartnerOrganizationSerializer(many=True, read_only=True)
+    poi_type = PointOfInterestTypeSerializer(read_only=True)
+    country = serializers.SerializerMethodField(read_only=True)
+    region = serializers.SerializerMethodField(read_only=True)
+    district = serializers.SerializerMethodField(read_only=True)
+
+    def extract_location_info(self, location_obj):
```
fpbonet, line 227: This method has a mix of responsibilities: getting the upper parent location and then serializing it. We have a good opportunity for refactor here.

src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -0,0 +1,670 @@
...
+        return PointOfInterestAdminSerializer().extract_location_info(obj).get('region')
+
+    def get_district(self, obj):
+        return PointOfInterestAdminSerializer().extract_location_info(obj).get('district')
+
+    def get_primary_type(self, obj):
+        return obj.poi_type.name if obj.poi_type else None
+
+    def get_implementing_partner(self, obj):
+        partners = obj.partner_organizations.all()
```
fpbonet, line 304: prefetch partner_organizations.organization or otherwise we have an N+1 problem here

src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -0,0 +1,670 @@
...
+    point_of_interest = serializers.PrimaryKeyRelatedField(
+        many=True,
+        queryset=models.PointOfInterest.objects.all(),
+        write_only=True
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        for field in ['last_name', 'first_name', 'email']:
+            self.fields[field].read_only = True
```
fpbonet, line 358: the read_only property might be better to define it in the fields themselves rather than in the __init__ method. I would declare last_name, first_name and email explicitely as WhateverSerializerFields(read_only=True)

src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -0,0 +1,670 @@
...
+    items = TransferItemDetailSerializer(many=True, write_only=True, required=True)
+
+    adminValidator = AdminPanelValidator()
+
+    @transaction.atomic
+    def create(self, validated_data):
+        self.adminValidator.validate_items(validated_data.get('items', []))
+        self.adminValidator.validate_partner_location(validated_data.get('location'), validated_data.get('partner_organization'))
+        items = validated_data.pop('items', [])
+        instance = models.Transfer.objects.create(
```
fpbonet, line 520: Seeing that some of these parameters are fixed, we could think of having a Transfer.objects.create_manual(...) and only pass the parameters which are variable.

src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -0,0 +1,670 @@
...
+            unicef_release_order=f"{TRANSFER_MANUL_CREATION_NAME} {timezone.now().strftime('%d-%m-%Y %H:%M:%S')}",
+            partner_organization=validated_data.get('partner_organization'),
+            origin_point=models.PointOfInterest.objects.get(pk=1),  # Unicef warehouse
```
fpbonet, line 523: PointOfInterest.objects.get_unicef_warehouse()
semantics! hehe

src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -0,0 +1,670 @@
...
+
+    class Meta:
+        model = PartnerOrganization
+        fields = ('id', 'name', 'vendor_number')
+
+
+class AuthUserPermsDetailSerializer(serializers.ModelSerializer):
+    admin_perms = serializers.SerializerMethodField()
+
+    def get_admin_perms(self, obj):
```
fpbonet, line 658: First, let's avoid abbreviations as Perms as they require extra effort to understand when reading the code. Second, I'm thinking this logic would be better suited in a special class, like UserPermissionRetriever or something like this. Let's discuss.

src/etools/applications/last_mile/admin_panel/validators.py
```diff
@@ -0,0 +1,60 @@
...
+from django.utils.translation import gettext_lazy as _
+
+from rest_framework.exceptions import ValidationError
+
+from etools.applications.last_mile.models import Material, PointOfInterest, PointOfInterestType
+from etools.applications.partners.models import PartnerOrganization
+from etools.applications.users.models import Group, Realm
+
+
+class AdminPanelValidator:
```
fpbonet, line 11: Nice!
A couple of minor notes. I would avoid naming the variable "value" as it's very generic. I would choose a naming like this:

```
def validate_poi_type(self, poi_type: str) -> None
```

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -0,0 +1,468 @@
...
+
+    def get_queryset(self):
+        schema_name = connection.tenant.schema_name
+        User = get_user_model()
+
+        queryset = User.objects.filter(realms__country__schema_name=schema_name)
+
+        has_active_location = self.request.query_params.get('hasActiveLocation')
+        if has_active_location == "1":
+            queryset = queryset.filter(profile__organization__partner__points_of_interest__isnull=False)
```
fpbonet, line 82: Same thing here, we might do something like this:
queryset.with_points_of_interest()
and 
queryset.without_points_of_interest()

A little extra work but clearer and easier to read

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -0,0 +1,468 @@
...
+    permission_classes = [IsLMSMAdmin]
+    pagination_class = CustomDynamicPageNumberPagination
+
+    def get_queryset(self):
+        schema_name = connection.tenant.schema_name
+        User = get_user_model()
+
+        queryset = User.objects.filter(realms__country__schema_name=schema_name)
+
+        has_active_location = self.request.query_params.get('hasActiveLocation')
```
fpbonet, line 80: this parameter is only for one of the endpoints right? the List one?

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -0,0 +1,468 @@
...
+    ]
+
+    ordering = ('id',)
+
+    def get_serializer_context(self):
+        context = super().get_serializer_context()
+        context['country_schema'] = connection.tenant.schema_name
+        return context
+
+    def get_serializer_class(self):
```
fpbonet, line 111: The pains of viewsets hahah

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -0,0 +1,468 @@
...
+        return UserAdminSerializer
+
+    @action(
+        detail=False,
+        methods=['get'],
+        url_path='export/csv',
+        renderer_classes=(ExportCSVRenderer,),
+    )
+    def list_export_csv(self, request, *args, **kwargs):
+        users = self.filter_queryset(self.get_queryset())
```
fpbonet, line 128: This definitely needs to be on a class of its own. What about an UserAdminExporter? We should avoid business logic in the views as much as possible.

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -0,0 +1,468 @@
...
+        return PointOfInterestAdminSerializer
+
+    @action(
+        detail=False,
+        methods=['get'],
+        url_path='export/csv',
+        renderer_classes=(ExportCSVRenderer,),
+    )
+    def list_export_csv(self, request, *args, **kwargs):
+        users = self.filter_queryset(self.get_queryset())
```
fpbonet, line 196: Same thing as above. Write an Exporter class and remove logic from view.

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -0,0 +1,468 @@
...
+        })
+
+
+class LocationsViewSet(mixins.ListModelMixin,
+                       mixins.CreateModelMixin,
+                       mixins.RetrieveModelMixin,
+                       mixins.UpdateModelMixin,
+                       viewsets.GenericViewSet):
+    """
+    GeoJSON representation of a geographical point.
```
fpbonet, line 150: Is this comment accurate?? If not, remove. Not useful.

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -0,0 +1,468 @@
...
+class OrganizationListView(mixins.ListModelMixin, GenericViewSet):
+    serializer_class = OrganizationAdminSerializer
+    permission_classes = [IsAuthenticated]
+
+    def get_queryset(self):
+        queryset = Organization.objects.all() \
+            .select_related('partner', 'auditorfirm', 'tpmpartner') \
+            .prefetch_related('auditorfirm__purchase_orders', 'tpmpartner__countries')
+
+        q_partner = Q(partner__isnull=False, partner__hidden=False)
```
fpbonet, line 341: this query is quite complex. Better naming would help a lot.
instaed of q_partner you might write `with_partner = Q(...)`

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -0,0 +1,468 @@
...
+
+    filter_backends = (DjangoFilterBackend, SearchFilter, OrderingFilter)
+
+    filterset_class = TransferHistoryFilter
+
+    ordering_fields = ['unicef_release_order', 'transfer_name', 'transfer_type', 'status', 'partner_organization', 'destination_point', 'origin_point']
+    search_fields = ['unicef_release_order', 'transfer_name', 'transfer_type', 'status', 'partner_organization', 'destination_point', 'origin_point']
+
+    def get_queryset(self):
+        transfer_subquery = models.Transfer.objects.filter(
```
fpbonet, line 413: Ouch this query can be really expensive. Why can origin_transfer_id not be a regular FK?

src/etools/applications/last_mile/models.py
```diff
@@ -469,3 +477,17 @@ class ItemTransferHistory(TimeStampedModel, models.Model):
 
     class Meta:
         unique_together = ('transfer', 'item')
+
+
+class AdminPanelPermission(models.Model):
```
fpbonet, line 482: Wait, if we don't need a table, do we really need a django Model for this? Can it not be a normal Class?

src/etools/applications/last_mile/tests/test_admin_panel/test_general_views.py
```diff
@@ -0,0 +1,141 @@
...
+
+from etools.applications.core.tests.cases import BaseTenantTestCase
+from etools.applications.last_mile.admin_panel.constants import *  # NOQA
+from etools.applications.last_mile.tests.factories import MaterialFactory
+from etools.applications.organizations.tests.factories import OrganizationFactory
+from etools.applications.partners.tests.factories import PartnerFactory
+from etools.applications.users.tests.factories import UserPermissionFactory
+
+
+class TestManageGeneralView(BaseTenantTestCase):
```
fpbonet, line 15: what's ManageGeneralView?

src/etools/applications/last_mile/tests/test_admin_panel/test_manage_alert_notifications.py
```diff
@@ -0,0 +1,414 @@
...
+
+    def test_create_alert_notification_success(self):
+        payload = {
+            "email": self.partner_staff.email,
+            "group": self.wastage_notification.pk,
+        }
+        response = self.forced_auth_req('post', self.url, user=self.partner_staff, data=payload)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        # Verify that a Realm instance was created with the correct attributes.
```
fpbonet, line 64: what about asserting the created notification exists?

src/etools/applications/last_mile/tests/test_admin_panel/test_manage_alert_notifications.py
```diff
@@ -0,0 +1,414 @@
...
+        cls.url = reverse(f'{ADMIN_PANEL_APP_NAME}:{ALERT_NOTIFICATIONS_ADMIN_PANEL}-list')
+        valid_group_name = list(ALERT_TYPES.keys())
+        cls.wastage_notification = GroupFactory(name=valid_group_name[0])
+        cls.acknowledgement = GroupFactory(name=valid_group_name[1])
+        cls.waybill = GroupFactory(name=valid_group_name[2])
+        cls.invalid_group = GroupFactory(name="Invalid Group")
+
+    def test_get_alert_notifications(self):
+        response = self.forced_auth_req('get', self.url, user=self.partner_staff)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
fpbonet, line 46: If we only check the status code then the test is very light. Let's add checks on the payload too. At least the most important fields.

src/etools/applications/last_mile/tests/test_admin_panel/test_manage_alert_notifications.py
```diff
@@ -0,0 +1,414 @@
...
+        realm = RealmFactory(
+            user=self.partner_staff,
+            country=self.country,
+            organization=self.partner_staff.profile.organization,
+            group=self.wastage_notification
+        )
+        detail_url = reverse(f'{ADMIN_PANEL_APP_NAME}:{ALERT_NOTIFICATIONS_ADMIN_PANEL}-detail', kwargs={'pk': realm.pk})
+        # Update the group to acknowledgement.
+        payload = {
+            "group": self.acknowledgement.pk,
```
fpbonet, line 159: Does this endpoint support updating other fields? 
I would enrich this test adding more updated fields and asserting them

src/etools/applications/last_mile/tests/test_admin_panel/test_manage_locations.py
```diff
@@ -0,0 +1,315 @@
...
+
+from etools.applications.core.tests.cases import BaseTenantTestCase
+from etools.applications.last_mile.admin_panel.constants import *  # NOQA
+from etools.applications.last_mile.tests.factories import PointOfInterestFactory, PointOfInterestTypeFactory
+from etools.applications.organizations.tests.factories import OrganizationFactory
+from etools.applications.partners.tests.factories import PartnerFactory
+from etools.applications.users.tests.factories import SimpleUserFactory, UserPermissionFactory
+
+
+class TestManageLocationsView(BaseTenantTestCase):
```
fpbonet, line 15: ```class TestLocationsViewSet(...)```
And same for the others

src/etools/applications/last_mile/admin_panel/validators.py
fpbonet: I don't necessarily dislike the approach here, a validator object with methods makes sense to me haha. I find it weird thogh that it's an attribute in the serializers. Let's discuss.

src/etools/applications/last_mile/tests/test_admin_panel/test_manage_alert_notifications.py
smihalache06: it was checked above, with the Realm.exists()

src/etools/applications/last_mile/tests/test_admin_panel/test_manage_alert_notifications.py
smihalache06: Support only 1 filed to be updated.

src/etools/applications/last_mile/admin_panel/filters.py
```diff
@@ -38,7 +38,7 @@ def filter_district(self, queryset, name, value):
         matching_ids = []
```
fpbonet, line 38: Given how Locations are modeled and what we learned from parent locations, this filtering can be very expensive (slow). Let's marke a shortcut card to evaluate query performance on Staging/Prod once we release this. We can use New Relic for that.

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -410,18 +394,33 @@ class TransferHistoryListView(mixins.ListModelMixin, GenericViewSet):
...
+            'origin_transfer__destination_point',
+            'origin_transfer__origin_point'
+        ).order_by('-created').annotate(
+            unicef_release_order=F('origin_transfer__unicef_release_order'),
+            transfer_name=F('origin_transfer__name'),
+            transfer_type=F('origin_transfer__transfer_type'),
+            status=F('origin_transfer__status'),
+            partner_organization=F('origin_transfer__partner_organization__organization__name'),
+            destination_point=F('origin_transfer__destination_point__name'),
+            origin_point=F('origin_transfer__origin_point__name')
```
fpbonet, line 409: why do we need to annotate all these? it's for the serializer? can't you use 
SerializerSomethingField(source="origin_transfer__unicef_release_order")
or something like that?

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -435,7 +434,27 @@ def get_queryset(self):
...
+            'name',
+            'transfer_type',
+            'transfer_subtype',
+            'status',
+            'origin_point__name',
+            'origin_point',
+            'destination_point__name',
+            'destination_point',
+            'from_partner_organization__organization__name',
+            'recipient_partner_organization__organization__name',
```
fpbonet, line 456: how many fields are we NOT loading with this .only() ? is it worth it, does it have a significant performance improvement to compensate for the verbosity in the code?

src/etools/applications/last_mile/admin_panel/views.py
fpbonet: Another idea: use @property methods to retrieve for example destination_point_name

smihalache06: @fpbonet if we want to change this annotate, we need to replace the entire FilterClass of this. Also the serializer. So, for now is better to keep as it is, without making the things more hard to read on the Filter and Serializer class.

fpbonet (approved):

PR #3851, opened 2025-03-10
[ch39479] FAM - Min Assurance Activities Required Email being sent to users with inactive Au....

ZakiTech-UNICEF:
[ch39479]

### DIFF
--- a/src/etools/applications/hact/tasks.py
+++ b/src/etools/applications/hact/tasks.py
@@ -103,6 +103,7 @@ def notify_hact_update(partner_list, country_id):
     recipients = get_user_model().objects.filter(
         realms__group=UNICEFAuditFocalPoint.as_group(),
         realms__country__id=country_id,
+        realms__is_active=True,
         is_superuser=False,
     ).values_list('email', flat=True).distinct()
     send_notification_with_template(
--- a/src/etools/applications/hact/tests/test_tasks.py
+++ b/src/etools/applications/hact/tests/test_tasks.py
@@ -116,3 +116,88 @@ class TestUpdateHactValues(BaseTenantTestCase):
         with patch("etools.applications.hact.tasks.update_hact_for_country.delay", mock_send):
             update_hact_values()
         self.assertEqual(mock_send.call_count, 1)
+
+class TestNotifyHactUpdate(BaseTenantTestCase):
+
+    def test_notify_hact_update(self):
+        logs = VisionSyncLog.objects.all()
+        self.assertEqual(logs.count(), 0)
+        partner = PartnerFactory(organization=OrganizationFactory(name="Partner XYZ"), reported_cy=20000)
+
+        active_unicef_focal_user = UserFactory(
+            is_staff=True, realms__data=[UNICEF_USER, UNICEFAuditFocalPoint.name]
+        )
+        active_unicef_focal_user.realms.update(is_active=True)
+
+        active_unicef_focal_user2 = UserFactory(
+            is_staff=True, realms__data=[UNICEF_USER, UNICEFAuditFocalPoint.name]
+        )
+        active_unicef_focal_user2.realms.update(is_active=True)
+
+        inactive_unicef_focal_user = UserFactory(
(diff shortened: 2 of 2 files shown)


src/etools/applications/hact/tests/test_tasks.py
```diff
@@ -116,3 +122,89 @@ def test_update_hact_values(self):
...
+        mock_send = Mock()
+        with patch("etools.applications.hact.tasks.send_notification_with_template", mock_send):
+            notify_hact_update(["Partner XYZ"], Country.objects.get(business_area_code=self.tenant.business_area_code).id)
+        self.assertEqual(mock_send.call_count, 1)
+
+        expected_recipients = list([active_unicef_focal_user2.email, active_unicef_focal_user.email])
+
+        actual_call_args = mock_send.call_args[1]
+        assert set(actual_call_args["recipients"]) == set(expected_recipients), \
+            f"Expected {expected_recipients}, got {actual_call_args['recipients']}"
```
fpbonet, line 205: well done on asserting the call, but on a quick note, there is this python native function `mock.assert_called_once_with` that does exactly what you need, without reinventing the wheel :)  https://docs.python.org/3.12/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with

```
mock_send.assert_called_once_with(set(active_unicef_focal_user2.email, active_unicef_focal_user.email))
```
Something like this should make it

src/etools/applications/hact/tests/test_tasks.py
```diff
@@ -116,3 +122,89 @@ def test_update_hact_values(self):
         with patch("etools.applications.hact.tasks.update_hact_for_country.delay", mock_send):
             update_hact_values()
         self.assertEqual(mock_send.call_count, 1)
+
+
+class TestNotifyHactUpdate(BaseTenantTestCase):
+
+    def test_notify_hact_update(self):
+        logs = VisionSyncLog.objects.all()
+        self.assertEqual(logs.count(), 0)
```
fpbonet, line 131: Do we need this? DB is flushed in between test cases

src/etools/applications/hact/tests/test_tasks.py
ZakiTech-UNICEF: In my latest test for the HACT notify function, I encountered inconsistencies in the ordering of recipients. Since the recipients are retrieved from the database, their order is not guaranteed to be the same across different test runs.
To ensure that the test reliably compares the expected and actual recipients, I had to apply sorting on both sides:

In the function: Sorting ensures that send_notification_with_template always receives recipients in a consistent order.
In the test: Sorting ensures that assert_called_once_with compares two lists that are ordered identically, preventing test failures due to random order variations.
Without sorting, the test fails intermittently when the order of recipients differs between the expected and actual output, even though they contain the same elements. 
image

src/etools/applications/hact/tasks.py
```diff
@@ -103,10 +103,11 @@ def notify_hact_update(partner_list, country_id):
     recipients = get_user_model().objects.filter(
         realms__group=UNICEFAuditFocalPoint.as_group(),
         realms__country__id=country_id,
+        realms__is_active=True,
         is_superuser=False,
     ).values_list('email', flat=True).distinct()
     send_notification_with_template(
-        recipients=list(recipients),
+        recipients=sorted(list(recipients)),
```
emaciupe, line 110: @ZakiTech-UNICEF @fpbonet Let's stick to best practices here and remove sorted from the code for 2 reasons: 1. it is not necessary (also the list might get big - all UNICEF focal points of a country ) 2. one should not change the code for the sake of the tests, but change the tests instead.  
So here IMHO it's best to go with the original version of the recipients assertion as it is correct.

emaciupe (changes requested):

src/etools/applications/hact/tasks.py
ZakiTech-UNICEF: done

emaciupe (approved):

PR #3855, opened 2025-03-12
[ch38340] Added indicators render to pdf export, based on the conditions from t…

ZakiTech-UNICEF:
[ch38340]

### DIFF
--- a/src/etools/applications/partners/templates/pd/detail.html
+++ b/src/etools/applications/partners/templates/pd/detail.html
@@ -295,27 +295,27 @@
                         {% if indicator.indicator.unit == "number" %}
                             {% with variable="#" %}
                                 <td class="w-p30">
-                                    {% if not indicator.is_active %} ({ translate 'Inactive' }){% endif %}
+                                    {% if not indicator.is_active %} ({% translate 'Inactive' %}){% endif %}
                                     {{ variable }} {{ indicator.indicator.title }}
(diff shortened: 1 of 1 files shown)


src/etools/applications/partners/templates/pd/detail.html
```diff
@@ -291,7 +291,35 @@
...
+                            {% with variable="%" %}
+                                <td class="w-p30">
+                                    {% if not indicator.is_active %} ({{ _('Inactive') }}){% endif %}
+                                    {{ variable }} {{ indicator.indicator.title }}
+                                </td>
+                            {% endwith %}
+                        {% elif indicator.indicator.unit == "percentage" and indicator.indicator.display_type == "ratio" %}
+                            {% with variable="÷" %}
+                                <td class="w-p30">
+                                    {% if not indicator.is_active %} ({{ _('Inactive') }}){% endif %}
```
emaciupe, line 312: @ZakiTech-UNICEF is the syntax `_()` in _('Inactive') translating in html rendering? asking because on template rendering, so far only `translate` tag is being used

ZakiTech-UNICEF: yes it is doing the same how I saw, but will change it back to use the same syntax how we are using everywhere. Thank you

emaciupe (approved):

PR #3860, opened 2025-03-25
[ch37278] FR creation from external also triggers pd auto transition if all reqs are met

emaciupe:
[ch37278]

### DIFF
--- a/src/etools/applications/funds/tests/test_views.py
+++ b/src/etools/applications/funds/tests/test_views.py
@@ -13,9 +13,7 @@ from django.utils import timezone
 from rest_framework import status
 from rest_framework.authtoken.models import Token
 from rest_framework.test import APIClient
-from unicef_locations.tests.factories import LocationFactory
 
-from etools.applications.attachments.tests.factories import AttachmentFactory
 from etools.applications.core.tests.cases import BaseTenantTestCase
 from etools.applications.environment.tests.factories import TenantSwitchFactory
 from etools.applications.funds.tests.factories import (
@@ -29,16 +27,8 @@ from etools.applications.partners.models import Agreement, Intervention
 from etools.applications.partners.tests.factories import (
     AgreementFactory,
     InterventionFactory,
-    InterventionResultLinkFactory,
     PartnerFactory,
-)
-from etools.applications.reports.models import ResultType
-from etools.applications.reports.tests.factories import (
-    InterventionActivityFactory,
-    LowerResultFactory,
-    OfficeFactory,
-    ReportingRequirementFactory,
-    SectionFactory,
+    SignedInterventionFactory,
 )
 from etools.applications.users.tests.factories import UserFactory
 from etools.libraries.tests.vcrpy import VCR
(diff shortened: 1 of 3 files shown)


src/etools/applications/funds/tests/test_views.py
```diff
@@ -398,30 +421,108 @@ def test_post_201(self):
...
+            self.assertEqual(self.get_instance_str(actual_value), data[field])
 
         self.assertEqual(funds_reservation.fr_items.count(), len(data['fr_items']))
         for actual_item, expected_item in zip(funds_reservation.fr_items.all(), data['fr_items']):
             for field in data['fr_items'][0].keys():
                 actual_value = getattr(actual_item, field)
-                self.assertEqual(get_instance_str(actual_value), expected_item[field])
+                self.assertEqual(self.get_instance_str(actual_value), expected_item[field])
+
+    @override_settings(ETOOLS_EZHACT_EMAIL='test@example.com', ETOOLS_EZHACT_TOKEN='testkey')
```
fpbonet, line 447: I'm curious to know why we need this. A signal?

src/etools/applications/funds/tests/test_views.py
```diff
@@ -398,30 +421,108 @@ def test_post_201(self):
...
+            signed_by_partner_date=date.today() - timedelta(days=1),
+            unicef_signatory=unicef_staff,
+            partner_authorized_officer_signatory=self.partner.active_staff_members.all().first(),
+            cash_transfer_modalities=[Intervention.CASH_TRANSFER_DIRECT],
+        )
+        signed_intervention.flat_locations.add(LocationFactory())
+        signed_intervention.partner_focal_points.add(partner_user)
+        signed_intervention.unicef_focal_points.add(unicef_staff)
+        signed_intervention.offices.add(OfficeFactory())
+        signed_intervention.sections.add(SectionFactory())
```
fpbonet, line 498: Only a suggestion: If you find this setup is too long, we might consider adding a method to the factory like InterventionFactory.signed() and "hide" the details under this method. Leaving the test a little bit cleaner.

src/etools/applications/funds/views_ext.py
```diff
@@ -25,4 +28,13 @@ def create(self, request, *args, **kwargs):
         serializer.is_valid(raise_exception=True)
         intervention = get_object_or_404(Intervention, number=serializer.validated_data.get('pd_reference_number'))
         serializer.save(intervention=intervention)
+
+        old_status = intervention.status
+        admin_user = get_object_or_404(get_user_model(), username=settings.TASK_ADMIN_USER)
```
fpbonet, line 33: minor detail, but should we return 404 for this?

fpbonet (approved):
Overall looks good. 
I think this endpoint is a great opportunity to do a refactor and encapsulate the logic of creating an ExternalFundsReservation + its items + the validation/transition into an action, removing business logic from the serializer and the view. 
That way we have everything in one place. This refactor can shouldn't be too hard as the endpoint has good testing coverage.

src/etools/applications/funds/tests/test_views.py
```diff
@@ -339,16 +357,21 @@ def setUpTestData(cls):
...
-    def test_post_201(self):
-        self.client.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
-        tenant_switch = TenantSwitchFactory(name="ezhact_external_fr_disabled")
-        tenant_switch.countries.add(connection.tenant)
-        self.assertTrue(tenant_switch.is_active())
-        data = {
+        try:
+            cls.admin_user = get_user_model().objects.get(username=settings.TASK_ADMIN_USER)
+        except get_user_model().DoesNotExist:
+            cls.admin_user = UserFactory(username=settings.TASK_ADMIN_USER)
```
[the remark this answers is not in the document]
emaciupe: @fpbonet in the context when running all the tests, that user already exists in the test db - but in the context of running only the class or one test alone, it will not.

src/etools/applications/funds/tests/test_views.py
emaciupe: @fpbonet this is the usually approach in tests to override settings - these are actually environment variables in production env

src/etools/applications/funds/views_ext.py
emaciupe: @fpbonet I don't follow here, what do you suggest?  it's standard rest framework function for validation that the obj exists in db, with many usages all over this project- the 404 is the accepted status code when accessing something that does not exist - since the request was valid, however no intervention / user was found. 400 is when the client request is malformed / invalid

PR #3864, opened 2025-04-01
Updating circle_ci config: avoid test reruns on PR merges

fpbonet:
Objective:
- Only run tests on PRs, not on pushes to branches (even PR merges)
- Pushes (PR merges) should only build and push for develop, staging and master branches.

Changes:
- Updated config to circle_ci v2.1
- Split setup, test, execution into 2 workflows: one for PRs, one for pushes and merges.
- Remove launch of build_and_push job from the test job. Not needed anymore as they're split workflows now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2.1
+version: 2
 
 parameters:
   base-tag:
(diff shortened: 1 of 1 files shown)


emaciupe (approved):
great work!

PR #3875, opened 2025-04-14
[ch39585] PMP : add pending unsupported amount column in partner details

emaciupe:
[ch39585]

### DIFF
--- a/src/etools/applications/audit/serializers/engagement.py
+++ b/src/etools/applications/audit/serializers/engagement.py
@@ -372,22 +372,28 @@ class EngagementHactSerializer(EngagementLightSerializer):
     object_url = serializers.ReadOnlyField(source='get_object_url')
 
     def get_amount_tested(self, obj):
-        if obj.engagement_type == 'audit':
+        if obj.engagement_type == Engagement.TYPE_AUDIT:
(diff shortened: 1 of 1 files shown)


src/etools/applications/audit/serializers/engagement.py
```diff
@@ -369,25 +369,30 @@ def validate(self, data):
...
 
     def get_outstanding_findings(self, obj):
-        if obj.engagement_type in ['audit', 'sc']:
+        if obj.engagement_type == Engagement.TYPE_AUDIT:
+            return obj.financial_findings or 0
+        return 0
+
+    def get_pending_unsupported_amount(self, obj):
+        if obj.engagement_type in [Engagement.TYPE_AUDIT, Engagement.TYPE_SPOT_CHECK]:
             return obj.pending_unsupported_amount or 0
```
fpbonet, line 390: this kind of logic would be better suited in the model itself rather than the serializer

fpbonet (approved):
Checking the existing code, it's unfortunate this endpoint does not test the returned payload in `test_hact_view`. I think this is something we should do moving forward. What do you think? Approving to not block

PR #3879, opened 2025-04-25
[ch39773] Updated filter to exclude users with is_active=False roles when roles are not em…

ZakiTech-UNICEF:
[ch39773]


emaciupe (approved):
Looks good!

PR #3883, opened 2025-04-29
[ch39792] FAM - correction to date validation

emaciupe:


fpbonet (approved):
Approved to not block the hotfix but please add a test case on a next PR

PR #3912, opened 2025-05-22
[ch40019] remove rutf and add partner specific material

smihalache06:


src/etools/applications/last_mile/models.py
```diff
@@ -499,6 +502,22 @@ def description(self):
     def should_be_hidden(self):
```
fpbonet, line 502: do we still need this other method? is it used elsewhere? If not, please remove. If possible, let's have only one "def should_be_hidden"

src/etools/applications/last_mile/tests/test_views.py
```diff
@@ -389,21 +400,21 @@ def test_partial_checkin_RUFT_material(self):
         self.assertIn(response.data['proof_file'], self.attachment.file.path)
         self.assertEqual(self.incoming.name, checkin_data['name'])
         self.assertEqual(self.incoming.items.count(), len(response.data['items']))
-        self.assertEqual(self.incoming.items.count(), 1)  # only 1 checked-in item is visible, non RUFT
+        self.assertEqual(self.incoming.items.count(), 2)  # only 1 checked-in item is visible, non RUFT
```
fpbonet, line 403: the comment got deprecated haha

PR #3915, opened 2025-05-26
[ch36919] (reason) fm interface - add modified_date to partners, interventions,…

ZakiTech-UNICEF:
[ch36919] … cp_outputs, ALSO added updated_entities to include those above, but filter them by fitting in current year + test

### DIFF
--- a/src/etools/applications/field_monitoring/planning/serializers.py
+++ b/src/etools/applications/field_monitoring/planning/serializers.py
@@ -1,5 +1,6 @@
 from copy import copy
 
+from django.utils import timezone
 from django.utils.translation import gettext as _
 
 from rest_framework import serializers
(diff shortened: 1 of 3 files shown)


src/etools/applications/field_monitoring/planning/serializers.py
```diff
@@ -134,8 +140,48 @@ class Meta:
...
+        year_start_boundary = date(cur_start_year, 1, 1)
+        year_end_boundary = date(cur_end_year, 12, 31)
+
+        covering_acts = (
+            MonitoringActivity.objects
+            .filter(
+                start_date__lte=year_start_boundary,
+                end_date__gte=year_end_boundary
+            )
+            .prefetch_related("partners", "interventions", "cp_outputs")
```
fpbonet, line 159: can you use select_related here? It's more efficient as it does the join in the SQL query. Prefetch only joins them in python.

ZakiTech-UNICEF: hello, select_related helps only with single-row relations, as we have many-to-many we will still need prefetch post-filter (select, filter, prefetch), so i did it this way, let me know if i understood you correctly. Thanks!

fpbonet (approved):

PR #3926, opened 2025-06-06
[LMSM]Fixes for the inventories

smihalache06:


src/etools/applications/last_mile/serializers.py
```diff
@@ -86,6 +86,10 @@ def get_material_type_translate(self, obj):
 class TransferListSerializer(serializers.ModelSerializer):
     origin_point = PointOfInterestLightSerializer(read_only=True)
     destination_point = PointOfInterestLightSerializer(read_only=True)
+    shipment_date = serializers.SerializerMethodField()
```
fpbonet, line 89: No need for a custom method, you can use smthing like this:
```
serializers.DateTimeField(source='created', required=False)
```

src/etools/applications/last_mile/views.py
```diff
@@ -286,7 +286,7 @@ def outgoing(self, request, *args, **kwargs):
         location = self.get_parent_poi()
         qs = self.get_queryset()\
             .filter(status=models.Transfer.PENDING, origin_point=location)\
-            .exclude(destination_point=location)
+            .exclude(destination_point=location).order_by("-created")
```
fpbonet, line 289: is there a test that checks the order of these records?

PR #3940, opened 2025-06-24
[LMSM]Improves and tests on external views part 1

smihalache06:


src/etools/applications/last_mile/tests/test_views_ext.py
```diff
@@ -378,3 +389,267 @@ def test_robustness_with_numeric_fields_as_strings(self):
         item = models.Item.objects.first()
         self.assertEqual(item.quantity, 150)
         self.assertEqual(item.amount_usd, Decimal("350.55"))
+
+
+class VisionIngestMaterialsApiViewTest(BaseTenantTestCase):
```
fpbonet, line 394: syntax should be TestBlablablaView for all test classes

PR #3981, opened 2025-08-04
[257182] Add action points and action taken to visit pdf export

emaciupe:


src/etools/applications/field_monitoring/planning/views.py
```diff
@@ -274,7 +274,12 @@ def visit_pdf(self, request, *args, **kwargs):
             "interventions": ', '.join([str(intervention) for intervention in ma.interventions.all()]),
             "overall_findings": list(ma.activity_overall_findings().values('entity_name', 'narrative_finding', 'on_track')),
             "summary_findings": ma.get_export_activity_questions_overall_findings(),
-            "data_collected": ma.get_export_checklist_findings()
+            "data_collected": ma.get_export_checklist_findings(),
+            "action_points": ma.get_export_action_points(request),
+            "related_attachments": ma.get_export_related_attachments(request),
+            "reported_attachments": ma.get_export_reported_attachments(request),
+            "checklist_attachments": ma.get_export_checklist_attachments(request),
```
fpbonet, line 281: These export methods might be better placed in a dedicated class, as the MonitoringActivity model is already quite loaded with methods.
In the LM side of the app we've developed dedicated classes like DataExportService to handle these querying and data setup.

fpbonet (approved):

PR #3983, opened 2025-08-04
[269043] LMSM Fix identified bugs

smihalache06:


src/etools/applications/last_mile/admin_panel/services/reverse_transfer.py
```diff
@@ -7,43 +7,56 @@
 
 class TransferReverse:
```
fpbonet, line 8: Please make file and class name consistent

smihalache06: Fixed

PR #3999, opened 2025-08-14
[259030] LMSM Stock Management Admin

smihalache06:


src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -609,28 +681,7 @@ class TransferItemCreateSerializer(serializers.ModelSerializer):
...
-                    transfer=instance,
-                    material=item.get('material'),
-                    quantity=item.get('quantity'),
-                    uom=item.get('uom'),
-                    batch_id=item.get('item_name'),
-                )
-            )
-        models.Item.objects.bulk_create(items_to_create)
-        return instance
+        return StockManagementCreateService().create_stock_management(validated_data)
```
fpbonet, line 684: nice

src/etools/applications/last_mile/admin_panel/serializers.py
```diff
@@ -397,10 +400,42 @@ def get_lat(self, obj):
...
+        data.update(parent_locations)
+        return data
+
+    def generate_rows(self, instance):
+        base = self.base_representation(instance)
+        transfers = (
+            instance.destination_transfers
+            .all()
+            .prefetch_related('items')
+        )
```
fpbonet, line 415: Mmm I don't understand what this code is doing here in this serializer

smihalache06: Is creating the rows for each item. For example we have in location X 20 items. Will duplicate all the info and add 20 rows for each different item.

src/etools/applications/last_mile/admin_panel/views.py
```diff
@@ -386,29 +403,73 @@ def get_queryset(self):
...
+        if self.action == "_import_file":
+            return ImportFileSerializer
+        return ItemTransferAdminSerializer
+
+    @action(detail=False, methods=['post'], url_path='import/xlsx')
+    def _import_file(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        excel_file = serializer.validated_data['file']
+        valid, out = CsvImporter().import_stock(excel_file)
```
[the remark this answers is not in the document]
smihalache06: It's hard to split in 3 different classes, because from the CsvImporter we are using the same logic for all the imports (locaitons, stock, users). At the moment we are using the general CsvImporter but with specific methods like import_stock, import_users, import_locations.

